PR #1, opened 2015-04-01
Rename mentoring to problem builder to allow simultaneous installation

bradenmacdonald:
This PR makes the necessary changes to "mentoring v2" (now called Problem Builder) so that it can be simultaneously installed with mentoring v1.

There are no real code changes other than a bunch of `s/mentoring/problem_builder/`

I did also delete some old stuff like the DB migrations (which would conflict with mentoring v1 if left in) and the LightChildren DB table.


bradenmacdonald:
@antoviaque FYI. What sort of review do we need for this?

bradenmacdonald:
Also, @antoviaque, I don't know if we have time to do this, but: I would love to convert the `Answer` block in this problem_builder/mentoring-v2 to use edx-submissions rather than the custom Django model. I think it would support all the same functionality, plus more. If we could do that before any students start using the new blocks, it will make things much nicer. The only consequence would be that courses could only be upgraded between runs, and not while running. Which is a good thing anyways.

antoviaque:
@bradenmacdonald It's true that it would be nice to switch to edx-submissions, but we are running a bit out of time (we need to deliver to the course team so they can start authoring content), and I don't want to add another last minute review to the OSPR team, we have pushed a lot of PRs to them already lately. I know it will be more complicated to do this upgrade after the name change, but we will also have more time to do things calmly - which for migrations can be a good thing : )

sarina:
This seems like a reasonable change. I don't feel compelled to look at it closely, as long as you guys have verified that all functionality remains and you're all comfortable with the work. :+1:  from my perspective - go through whatever review your team deems necessary and then we can install it on edx.org.

bradenmacdonald:
Closed and opened to trigger Travis build

bradenmacdonald:
Merging now because this is blocking a bunch of time critical PRs and I need to rebase other work on this.

Build is green, this is a brand new repo, and the code here has been reviewed extensively except for this rename commit which doesn't change any of the behaviour, just the name. I have been testing it and found no issues so far caused by this rename.

antoviaque:
We discussed this - we'll still do a careful code review of this PR and testing on a sandbox before updating the requirements on edx-platform, to be on the safe side. But agreed that this should be pretty straightforward.

PR #2, opened 2015-04-03
Upgrade Script

bradenmacdonald:
When testing the upgrade path from mentoring v1 to problem builder (mentoring v2), I found a few bugs in the upgrade script:
* if `<tip>` elements had a `width` or `height`, they wouldn't work
* Warnings from the XML upgrade script were not displayed to the user
* Did not convert the actual block type from `mentoring` to `problem-builder`

I also added a check for multiple blocks that have the same `url_name`, since that should not happen but it is the case on some courses I worked with.

Note:
* This is based on PR #1 so ignore the many changes coming from that
* This is irrelevant to new courses - it only affects upgrading an old course.

### DIFF
--- a/problem_builder/questionnaire.py
+++ b/problem_builder/questionnaire.py
@@ -113,6 +113,15 @@ class QuestionnaireAbstractBlock(StudioEditableXBlockMixin, StudioContainerXBloc
 
         return block
 
+    @property
+    def html_id(self):
+        """
+        A short, simple ID string used to uniquely identify this question.
+
+        This is only used by templates for matching <input> and <label> elements.
+        """
+        return unicode(id(self))  # Unique as long as all choices are loaded at once
+
     def student_view(self, context=None):
         name = getattr(self, "unmixed_class", self.__class__).__name__
 
--- a/problem_builder/questionnaire.py
+++ b/problem_builder/questionnaire.py
@@ -81,7 +81,7 @@ class QuestionnaireAbstractBlock(StudioEditableXBlockMixin, StudioContainerXBloc
         scope=Scope.content,
         enforce_type=True
     )
-    editable_fields = ('question', 'message', 'weight')
+    editable_fields = ('question', 'message', 'weight', 'display_name', 'show_title')
     has_children = True
 
     def _(self, text):
@@ -113,12 +113,6 @@ class QuestionnaireAbstractBlock(StudioEditableXBlockMixin, StudioContainerXBloc
 
         return block
 
-    @property
-    def display_name_with_default(self):
-        if not self.lonely_step:
-            return self._(u"Question {number}").format(number=self.step_number)
-        return self._(u"Question")
-
     def student_view(self, context=None):
         name = getattr(self, "unmixed_class", self.__class__).__name__
 
@@ -127,6 +121,7 @@ class QuestionnaireAbstractBlock(StudioEditableXBlockMixin, StudioContainerXBloc
         context = context or {}
         context['self'] = self
         context['custom_choices'] = self.custom_choices
+        context['hide_header'] = context.get('hide_header', False) or not self.show_title
 
         fragment = Fragment(loader.render_template(template_path, context))
         # If we use local_resource_url(self, ...) the runtime may insert many identical copies
--- a/problem_builder/questionnaire.py
+++ b/problem_builder/questionnaire.py
@@ -20,6 +20,7 @@
 
 # Imports ###########################################################
 
+from django.utils.safestring import mark_safe
 from lxml import etree
 from xblock.core import XBlock
 from xblock.fields import Scope, String, Float, List, UNIQUE_ID
@@ -160,7 +161,7 @@ class QuestionnaireAbstractBlock(StudioEditableXBlockMixin, StudioContainerXBloc
 
     @property
     def human_readable_choices(self):
-        return [{"display_name": c.content, "value": c.value} for c in self.custom_choices]
+        return [{"display_name": mark_safe(c.content), "value": c.value} for c in self.custom_choices]
 
     @staticmethod
     def choice_values_provider(question):
--- a/problem_builder/questionnaire.py
+++ b/problem_builder/questionnaire.py
@@ -21,6 +21,7 @@
 # Imports ###########################################################
 
 from django.utils.safestring import mark_safe
+import logging
 from lxml import etree
 from xblock.core import XBlock
 from xblock.fields import Scope, String, Float, List, UNIQUE_ID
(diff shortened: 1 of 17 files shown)


problem_builder/questionnaire.py
```diff
@@ -98,6 +99,9 @@ def parse_xml(cls, node, runtime, keys, id_generator):
 
         # Load XBlock properties from the XML attributes:
         for name, value in node.items():
+            if name not in block.fields:
```
e-kolpakov: @bradenmacdonald heads up - `parse_xml` is going to be removed: https://github.com/open-craft/xblock-mentoring/pull/16

bradenmacdonald: @e-kolpakov Yep, that's great. Thanks :)

problem_builder/v1/upgrade.py
```diff
@@ -42,14 +43,18 @@ def upgrade_block(block):
     Given a MentoringBlock "block" with old-style (v1) data in its "xml_content" field, parse
     the XML and re-create the block with new-style (v2) children and settings.
     """
-    assert isinstance(block, MentoringBlock)
+    assert isinstance(block, (MentoringBlock, NewMentoringBlock))
     assert bool(block.xml_content)  # If it's a v1 block it will have xml_content
     store = block.runtime.modulestore
     xml_content_str = block.xml_content
     parser = etree.XMLParser(remove_blank_text=True)
```
e-kolpakov: @bradenmacdonald does it work fine with mentorings with comments? It should at least convert them right removing comments, but preserving comments would be ideal (but it might be non-trivial)

problem_builder/v1/upgrade.py
```diff
@@ -148,15 +144,46 @@ def parse_xml_for_HtmlDescriptor(cls, node, runtime, keys, id_generator):
...
+        block = course.runtime.get_block(block_id)
+        if block.url_name in url_names:
+            print(u" ➔ Mentoring block {} conflicts with another block using the same url_name".format(block.url_name))
+            print(u'   (display_name: "{}", parent {}: "{}")'.format(
+                block.display_name, block.parent, block.get_parent().display_name
+            ))
+            stop = True
+            continue
+        if block.url_name != unicode(block_id.block_id):
+            print(u" ➔ Mentoring block {} has two different url_names assigned. Real url_name is {}".format(
```
e-kolpakov: @bradenmacdonald causes `IndexError` - template contains two placeholders, while only one argument passed to format - probably missed second argument or should use `{0}` instead of `{}`

problem_builder/v1/upgrade.py
```diff
@@ -148,15 +144,46 @@ def parse_xml_for_HtmlDescriptor(cls, node, runtime, keys, id_generator):
...
+    for block_id in blocks_found:
+        block = course.runtime.get_block(block_id)
+        if block.url_name in url_names:
+            print(u" ➔ Mentoring block {} conflicts with another block using the same url_name".format(block.url_name))
+            print(u'   (display_name: "{}", parent {}: "{}")'.format(
+                block.display_name, block.parent, block.get_parent().display_name
+            ))
+            stop = True
+            continue
+        if block.url_name != unicode(block_id.block_id):
```
e-kolpakov: @bradenmacdonald if `url_name` is not set in XML it still reports "two" urlnames set. Probably should add check for `url_name` being empty

problem_builder/v1/upgrade.py
```diff
@@ -148,15 +144,46 @@ def parse_xml_for_HtmlDescriptor(cls, node, runtime, keys, id_generator):
...
+            print(u" ➔ Mentoring block {} has two different url_names assigned. Real url_name is {}".format(
+                unicode(block_id.block_id)
+            ))
+            print(u'   (display_name: "{}", parent: "{}")'.format(block.display_name, block.get_parent().display_name))
+            print(u"   (to fix, edit the XML and set the url_name to the real value given above).")
+            stop = True
+        url_names.add(block.url_name)
+
+    if stop:
+        print(" ➔ Exiting due to errors preventing the upgrade.")
```
e-kolpakov: @bradenmacdonald nit: other strings are unicode, but this one is not

problem_builder/v1/upgrade.py
```diff
@@ -148,15 +144,46 @@ def parse_xml_for_HtmlDescriptor(cls, node, runtime, keys, id_generator):
...
+        sys.exit(1)
```
e-kolpakov: @bradenmacdonald nit: might make sense to merge the two lines into one:

    sys.exit(" ➔ Exiting due to errors preventing the upgrade.")

According to [docs](https://docs.python.org/2/library/sys.html#sys.exit) it would yeild the same exit code of 1.

e-kolpakov:
@bradenmacdonald also, there's a dangling statement in [xm_changes.py:187](https://github.com/open-craft/problem-builder/blob/mcka-upgrade-bug-fix/problem_builder/v1/xml_changes.py#L187) - it was introduced earlier, but now is a good enough moment to remove it.

problem_builder/v1/upgrade.py
e-kolpakov: @bradenmacdonald nevermind, I've checked it out and it indeed works just fine with comments

e-kolpakov:
@bradenmacdonald :+1: conditional exception is fixed.

problem_builder/v1/upgrade.py
```diff
@@ -148,15 +144,46 @@ def parse_xml_for_HtmlDescriptor(cls, node, runtime, keys, id_generator):
...
             blocks_found.append(block.scope_ids.usage_id)
         elif block.has_children:
             for child_id in block.children:
                 find_mentoring_blocks(block.runtime.get_block(child_id))
     find_mentoring_blocks(course)
+
     total = len(blocks_found)
     print(" ➔ Found {} mentoring blocks".format(total))
 
+    print(" ➔ Doing a quick sanity check of the url_names")
```
Kelketek: Same nit @e-kolpakov expressed earlier about other strings being unicode-- this one isn't.

Kelketek:
Looks like @e-kolpakov found all the good stuff already, save for one other unicode nit. :+1: conditional on you fixing the issue he mentioned with string formatting.

problem_builder/v1/upgrade.py
bradenmacdonald: Yeah, there are actually a lot of non-unicode strings in this file that contain unicode characters. My bad. Because they're just being treated as bytes and getting piped directly to stdout, it hasn't caused a problem, but you're right that they should be marked as Unicode.

bradenmacdonald:
Rebased and addressed all your comments. I added in code to automatically fix the url_names if a `--force` argument is given. Thanks @e-kolpakov and @Kelketek. I'll merge tomorrow based on your tentative thumbs up, but if you can review again, that would be ideal.

Note: extra commits are showing up because I should be merging to `develop`, not `master`. I will fix that when merging.

e-kolpakov:
@bradenmacdonald I've added a PR to include import-export fix. It targets this branch though, as I tried to simplify your work of merging this all, so please, take a look at #5 and merge.

bradenmacdonald:
Merged this to develop via ba65ce0aa6917ca8ea5d5d5681f64ade7be8ee3a, merging Eugeny's import fix as #7.

PR #3, opened 2015-04-03
Dashboard Block (OC-560)

bradenmacdonald:
Rebased version of https://github.com/open-craft/xblock-mentoring/pull/10 .

**Description**: This is a new "Dashboard" XBlock needed for the upcoming Data Wise course. It provides a nice way to summarize a user's self-assessment results.

You configure it like this, by pasting in the `url_name`s of some Problem Builder ("Mentoring") blocks:
dashboard-configuration

And it will then look like this (after a student has submitted their answers):
dashboard-example

Also, if you configure some fancy rules like this:
dashboard-visual-config

You can get a visual summary like this, that reflects the average value of each step. It can use the same color rules to color in the images:
dashboard-visual

**Sandbox**: http://sandbox3.opencraft.com:18010/container/i4x://BradenX/DWD/vertical/67f9498331e94553aeab1f5a8cfd5c8e


bradenmacdonald:
Rebased in preparation for merge. The old commit history is available at 2670c4deef9f40d31c9b8de4ac7680cbbb366a14 in case it is needed.

PR #5, opened 2015-04-07
Import export fix

e-kolpakov:


bradenmacdonald:
@e-kolpakov FYI I found a small bug with this that was causing one of the tests to fail: 9f101fc90657d74f6749a82894069924cab9315c (choice `value` is not a list).

I have fixed that and will be merging this in PR #7. Thanks :-)

PR #9, opened 2015-04-10
Hotfix: Four minor issues to address before release

bradenmacdonald:
**Description**: The `edx-release` branch contains only the code that has been reviewed by both OpenCraft and edX for release on edX.org. In testing that release branch, we found a few minor issues which this PR addresses.

Fixes:
* Bug: MCQ/MRQ question titles were not visible in the LMS, even when "Show Title" was true, if another MCQ/MRQ that came earlier in the same subsection had show_title set false.  
  Fixed with 45072bbc661d9306fa751295e276827ab5fef3b7
* Usability improvement: In Studio, an MCQ's question text could not be set until at least one choice was created. In addition, a validation warning was shown immediately when adding an MCQ.  
  Fixed with 3d4bf295bd760526a97735102fb448ccb28889a9
* Bug: If a problem builder in assessment mode had max_attempts=2, then a student completes one attempt, then max_attempts was changed to 0 (unlimited) by the course author, that student is unable to try again.  
  Fixed with 8bf362dd1edd7359393a812014213bbb8f079e95
* Usability improvement: the default value of "max_attempts" would change when the "mode" setting was changed, but it would only change after the user closed the edit dialog. This was confusing, as the value would not match what the user just saw. For now I have just removed this behaviour so that the default value is always '0' (no limit) and it never changes unless manually set by the user.  
  Fixed with 0d85a3fe174ab063aedeaa68b0d9a5b0e00e1ec4

Once these fixes are reviewed and approved here, I am also planning to cherry-pick them onto our master (development) branch.

Sandbox: http://sandbox2.opencraft.com:18010/

### DIFF
--- a/problem_builder/public/css/mentoring.css
+++ b/problem_builder/public/css/mentoring.css
@@ -40,12 +40,13 @@
     font-style: italic;
 }
 
-.mentoring h4 {
-    margin-bottom: 20px;
+.mentoring fieldset {
+    margin-top: 10px;
 }
 
-.mentoring h4 {
-    margin-top: 25px;
+.mentoring h3 {
+    margin-top: 0px;
+    margin-bottom: 7px;
 }
 
 .mentoring .submit {
@@ -60,6 +61,10 @@
 .mentoring legend {
     white-space: normal;
     display: table; /* Enable line-wrapping in IE8 */
+}
+
+.mentoring .choices legend.question p:last-child { /* Selector must be more specific than 'div.course-wrapper section.course-content p' */
+    margin-bottom: 0;
 }
 
 .mentoring .attempts {
--- a/problem_builder/mcq.py
+++ b/problem_builder/mcq.py
@@ -133,9 +133,7 @@ class MCQBlock(SubmittingXBlockMixin, QuestionnaireAbstractBlock):
         all_values = set(self.all_choice_values)
         correct = set(data.correct_choices)
 
-        if not all_values:
-            add_error(self._(u"No choices set yet."))
-        elif not correct:
+        if all_values and not correct:
             add_error(
                 self._(u"You must indicate the correct answer[s], or the student will always get this question wrong.")
             )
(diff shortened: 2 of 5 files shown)


bradenmacdonald:
I just pushed one more fix - in Firefox, the spacing around question titles was weird.

**Before**:


**After**:

e-kolpakov:
@bradenmacdonald verified on sandbox :+1:; however, description for fix 4 (default value of "max_attempts" ) is confusing - I thought it would automatically change max_attempts in studio editor, while it just does not update max_attempts at all anymore.

bradenmacdonald:
@sarina These are our last changes to problem builder as required by Harvard. We would like to get them onto edx.org as soon as possible, though it looks like we've missed this week's release unfortunately. Can you please take a quick look at this and if it looks good, I'll merge it and then update the hash in https://github.com/edx/edx-platform/pull/7638 and get your approval to merge that as well.

Let me know if you want an OSPR.

sarina:
Braden - I'm still getting caught up from Pycon. Can I take a look
tomorrow? An OSPR would be great, then I will be tracking it explicitly.

On Mon, Apr 13, 2015 at 12:21 PM, Braden MacDonald <notifications@github.com

bradenmacdonald:
Done: https://openedx.atlassian.net/browse/OSPR-530

Thanks @sarina

problem_builder/public/css/mentoring.css
```diff
@@ -62,6 +63,10 @@
     display: table; /* Enable line-wrapping in IE8 */
 }
 
+.mentoring legend p:last-child {
+    margin-bottom: 0 !important;
```
sarina: We're very allergic to !important - is there a way you can write without this? !important overrides everything on the page and is strongly discouraged

sarina:
@bradenmacdonald thanks for the description and laying out the fixes. 

I share @e-kolpakov 's feedback about how the fourth fix works, but I don't think it's terrible, just maybe a bit unexpected from your description.

problem_builder/public/css/mentoring.css
bradenmacdonald: Sure, I can rewrite this to use a longer (more specific) selector, so that it can still override the existing style in the LMS that's causing the issue here.

bradenmacdonald:
@sarina OK, I have removed the use of `!important` and clarified the description of that fix.

sarina:
@bradenmacdonald LGTM. :+1:

bradenmacdonald:
Thanks @sarina !

PR #10, opened 2015-04-10
Temporary workaround for missing data issue when importing

bradenmacdonald:
**Summary**:
There is currently a bug that causes child blocks of MCQ/MRQ/Rating blocks to be missing when importing a course. This is a fix, but I would rather not include this code permanently, so I'm calling it a workaround.

**Details**:
The bug seems to only affect MCQ/MRQ/Rating blocks that have at least one `Scope.content` field (question, message, weight) set. The child blocks (`pb-choice` and/or `pb-tip` blocks) get successfully imported into mongo, but the MCQ/MRQ/Rating block's `children` field will be empty, so the MCQ/MRQ/Rating will appear to have no choices available.

I have been unable to determine what specifically is causing this problem, or why it only affects these blocks under those circumstances.

This workaround switches our MCQ/MRQ/Rating blocks from using the standard XBlock serialization to using edx-platform's newly revised `XmlParserMixin`. As that code is very complex, I have not yet been able to determine what exactly it does differently that fixes this issue.

Sandbox: http://sandbox2.opencraft.com:18010/

### DIFF
--- a/problem_builder/questionnaire.py
+++ b/problem_builder/questionnaire.py
@@ -20,14 +20,22 @@
 
 # Imports ###########################################################
 
+import logging
 from lxml import etree
 from xblock.core import XBlock
 from xblock.fields import Scope, String, Float, List, UNIQUE_ID
 from xblock.fragment import Fragment
+from xblock.mixins import XmlSerializationMixin
 from xblock.validation import ValidationMessage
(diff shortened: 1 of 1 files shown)


bradenmacdonald:
This issue is being investigated in https://github.com/edx/XBlock/pull/295 so I'm closing this.

PR #14, opened 2015-04-16
Fix: Checkmark not shown if answer name is numeric

bradenmacdonald:
I found an issue while testing birch rebase: If a Problem Builder contains a Long Answer field with a "name" that is numeric (e.g. "1234"), then the checkmark does not appear after you submit your answer. (Everything works fine though, just the checkmark is missing).


e-kolpakov:
:+1:

Kelketek:
:+1:

PR #15, opened 2015-04-18
Fix upgrade bugs

bradenmacdonald:
This fixes some bugs in the upgrade script that upgrades a course from mentoring v1 to problem builder.

There is one fix per commit. The issues fixed are:

* Minor unicode error
* If `<title></title>` was present, it would cause an error because etree empty tags have `.text` value set to `None` rather than `""`
* The third is more interesting: when a unit (vertical) contained two mentoring blocks, the first one would disappear after the upgrade. The reason was aggressive caching of the "children" property by the platform. I bypassed some caching and it fixed the bug.
* Support for upgrading assessments that consisted of alternating `<html>` and `<mcq>` (or similar) blocks that were intended to be displayed at the same time.
* The upgrade script will now discard empty tips like `<tip reject="choice3"></tip>`. There is no reason to keep them, and it might result in the user seeing an empty blue tip popup.
* If the old mentoring v1 XML did not have a `<title>` element, then set `show_title="false"` when upgrading. This makes the new block's appearance consistent with the old.
* If a mentoring block with `display_submit="false"` came above other mentoring blocks, the subsequent blocks were broken due to a JS error.
* A course was using `<tip require="a">` in MCQ blocks, where `<tip display="a">` was expected. That seems incorrect, but we know what was meant, so the upgrade script can now handle that.
* Some of the component blocks like Answer were adding themselves to the shared `context`, causing an error when other blocks that shared the context called `context.copy` - e.g. the fix from 90d1317. I applied that same fix to all the child blocks and it solved the error.

### DIFF
--- a/problem_builder/answer.py
+++ b/problem_builder/answer.py
@@ -131,7 +131,8 @@ class AnswerBlock(AnswerMixin, StepMixin, StudioEditableXBlockMixin, XBlock):
         display_name=_("Question"),
         help=_("Question to ask the student"),
         scope=Scope.content,
-        default=""
+        default="",
+        multiline_editor=True,
     )
     weight = Float(
         display_name=_("Weight"),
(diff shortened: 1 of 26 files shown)


e-kolpakov:
@bradenmacdonald :+1: with few minor nits.

PR #16, opened 2015-04-22
Presentation updates

e-kolpakov:
**JIRA Ticket:** https://openedx.atlassian.net/browse/OSPR-543
**Details:** See JIRA ticket
**Sandbox:** [LMS](http://sandbox4.opencraft.com/courses/course-v1:edX+PB101+now/courseware/ee978680a3bc409a93e1e5d2efac7787/1654a7e6c6244bb18d6fd729b9f26557/), [CMS](http://sandbox4.opencraft.com:18010/container/block-v1:edX+PB101+now+type@vertical+block@977dea6433d044ddaac7389138175751?action=new)
**Merge deadline:** Friday 24th (in time for upcoming release)

**Testing instructions:**

1. Set up a couple of mentoring XBlocks
2, Set up Dashboard block to draw data from those blocks
3. Set custom value for "Label for average value"
4. Set "Display numeric values" to False
5. Set some "Questions to be hidden" (see help string). Note that only questions are counted toward question index, e.g. if Mentoring contains freeform answer and MCQ, MCQ index would be 1, not 2.
6. Configure visual representation according to help string.
7. Publish
8. Go to LMS, provide answers for mentoring blocks
9. Navigate to dashboard block

Observe the following:
* Visual progress image is horizontally centered.
* Average values have custom label specified at step 3
* Numeric values are not displayed, only color codes are available
* Questions configured to be excluded are excluded and not participate in average calculations.

### DIFF
--- a/problem_builder/dashboard.py
+++ b/problem_builder/dashboard.py
@@ -36,7 +36,7 @@ from .mcq import MCQBlock
 from .sub_api import sub_api
 from lazy import lazy
 from xblock.core import XBlock
-from xblock.fields import Scope, List, String
+from xblock.fields import Scope, List, String, Boolean, Dict
 from xblock.fragment import Fragment
 from xblock.validation import ValidationMessage
 from xblockutils.helpers import child_isinstance
@@ -172,6 +172,20 @@ class DashboardBlock(StudioEditableXBlockMixin, XBlock):
         ).format(example_here='["2754b8afc03a439693b9887b6f1d9e36", "215028f7df3d4c68b14fb5fea4da7053"]'),
         scope=Scope.settings,
     )
+    exclude_questions = Dict(
+        display_name=_("Questions to be hidden"),
+        help=_(
+            "Optional rules to exclude specific questions both from displaying in dashboard and from the calculated "
+            "average. Rules must start with the url_name of a mentoring block, followed by list of question numbers "
+            "to exclude. Rule set must be in JSON format. Question numbers are one-based (the first question being "
+            "number 1). Must be in JSON format. Examples: {examples_here}"
+        ).format(
+            examples_here='{"2754b8afc03a439693b9887b6f1d9e36":[1,2], "215028f7df3d4c68b14fb5fea4da7053":[1,5]}'
+        ),
+        scope=Scope.content,
+        multiline_editor=True,
+        resettable_editor=False,
+    )
     color_rules = String(
         display_name=_("Color Coding Rules"),
         help=_(
@@ -207,8 +221,23 @@ class DashboardBlock(StudioEditableXBlockMixin, XBlock):
         ),
         scope=Scope.content,
     )
+    average_label = String(
+        display_name=_("Label for average value"),
+        default=_("Average"),
+        help=_("Label to be shown for calculated average"),
+        scope=Scope.content,
+    )
+    show_numbers = Boolean(
+        display_name=_("Display values"),
+        default=True,
+        help=_("Toggles if numeric values are displayed"),
+        scope=Scope.content
+    )
 
-    editable_fields = ('display_name', 'mentoring_ids', 'color_rules', 'visual_rules', 'visual_title', 'visual_desc')
+    editable_fields = (
+        'display_name', 'mentoring_ids', 'exclude_questions', 'average_label', 'show_numbers',
+        'color_rules', 'visual_rules', 'visual_title', 'visual_desc'
+    )
     css_path = 'public/css/dashboard.css'
     js_path = 'public/js/dashboard.js'
 
@@ -321,6 +350,12 @@ class DashboardBlock(StudioEditableXBlockMixin, XBlock):
             except Exception:
                 return ""
 
+    def _get_problem_questions(self, mentoring_block):
+        """ Generator returning only children of specified block that are MCQs """
+        for child_id in mentoring_block.children:
+            if child_isinstance(mentoring_block, child_id, MCQBlock):
+                yield child_id
+
     def student_view(self, context=None):  # pylint: disable=unused-argument
         """
         Standard view of this XBlock.
@@ -336,20 +371,35 @@ class DashboardBlock(StudioEditableXBlockMixin, XBlock):
                 'display_name': mentoring_block.display_name,
                 'mcqs': []
             }
-            for child_id in mentoring_block.children:
-                if child_isinstance(mentoring_block, child_id, MCQBlock):
-                    # Get the student's submitted answer to this MCQ from the submissions API:
-                    mcq_block = self.runtime.get_block(child_id)
-                    mcq_submission_key = self._get_submission_key(child_id)
-                    try:
-                        value = sub_api.get_submissions(mcq_submission_key, limit=1)[0]["answer"]
-                    except IndexError:
-                        value = None
-                    block['mcqs'].append({
-                        "display_name": mcq_block.display_name_with_default,
-                        "value": value,
-                        "color": self.color_for_value(value) if value is not None else None,
-                    })
+            try:
+                hide_questions = self.exclude_questions.get(mentoring_block.url_name, [])
+            except Exception:  # pylint: disable=broad-except-clause
+                log.exception("Cannot parse exclude_questions setting - probably malformed: %s", self.exclude_questions)
+                hide_questions = []
+
+            for question_number, child_id in enumerate(self._get_problem_questions(mentoring_block), 1):
+                try:
+                    if question_number in hide_questions:
+                        continue
+                except TypeError:
+                    log.exception(
+                        "Cannot check question number - expected list of ints got: %s",
+                        hide_questions
+                    )
+
+                # Get the student's submitted answer to this MCQ from the submissions API:
+                mcq_block = self.runtime.get_block(child_id)
+                mcq_submission_key = self._get_submission_key(child_id)
+                try:
+                    value = sub_api.get_submissions(mcq_submission_key, limit=1)[0]["answer"]
+                except IndexError:
+                    value = None
+
+                block['mcqs'].append({
+                    "display_name": mcq_block.display_name_with_default,
+                    "value": value,
+                    "color": self.color_for_value(value) if value is not None else None,
+                })
             # If the values are numeric, display an average:
             numeric_values = [
                 float(mcq['value']) for mcq in block['mcqs']
@@ -384,6 +434,8 @@ class DashboardBlock(StudioEditableXBlockMixin, XBlock):
             'blocks': blocks,
             'display_name': self.display_name,
             'visual_repr': visual_repr,
+            'average_label': self.average_label,
+            'show_numbers': self.show_numbers,
         })
 
         fragment = Fragment(html)
@@ -405,6 +457,14 @@ class DashboardBlock(StudioEditableXBlockMixin, XBlock):
             list(self.get_mentoring_blocks(data.mentoring_ids, ignore_errors=False))
         except InvalidUrlName as e:
             add_error(_(u'Invalid block url_name given: "{bad_url_name}"').format(bad_url_name=unicode(e)))
+
+        if data.exclude_questions:
+            for key, value in data.exclude_questions.iteritems():
+                if not isinstance(value, list):
+                    add_error(
+                        _(u"Exclude questions is malformed: value for key {key} is {value}, expected list of integers")
+                        .format(key=key, value=value)
+                    )
 
         if data.color_rules:
             try:
--- a/problem_builder/templates/html/dashboard.html
+++ b/problem_builder/templates/html/dashboard.html
@@ -42,16 +42,30 @@
           {% for mcq in block.mcqs %}
             <tr>
               <th class="desc">{{ mcq.display_name }}</th>
-              <td class="value" {% if mcq.color %}style="border-right-color: {{mcq.color}};"{% endif %}>
-                {% if mcq.value %}{{ mcq.value }}{% endif %}
+              <td class="value"
+                  {% if mcq.color %} style="border-right-color: {{mcq.color}};"{% endif %}
+                  {% if not show_numbers %}
+                    {% if mcq.value %} aria-label="Score: {{mcq.value}}" {% else %} aria-label="{% trans 'No value yet' %}" {%endif%}
+                  {% endif %}
+              >
+                {% if mcq.value and show_numbers %}
+                  {{ mcq.value }}
+                {% endif %}
               </td>
             </tr>
           {% endfor %}
           {% if block.has_average %}
            <tr class="avg-row">
-              <th class="desc">{% trans "Average" %}</th>
-              <td class="value" {% if block.average_color %}style="border-right-color: {{block.average_color}};"{% endif %}>
-                {{ block.average|floatformat }}
+              <th class="desc">{{ average_label }}</th>
+              <td class="value"
+                  {% if block.average_color %} style="border-right-color: {{block.average_color}};"{% endif %}
+                  {% if not show_numbers %}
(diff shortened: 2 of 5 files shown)


problem_builder/dashboard.py
```diff
@@ -336,20 +372,23 @@ def student_view(self, context=None):  # pylint: disable=unused-argument
...
-                    try:
-                        value = sub_api.get_submissions(mcq_submission_key, limit=1)[0]["answer"]
-                    except IndexError:
-                        value = None
-                    block['mcqs'].append({
-                        "display_name": mcq_block.display_name_with_default,
-                        "value": value,
-                        "color": self.color_for_value(value) if value is not None else None,
-                    })
+            hide_questions = self.exclude_questions.get(mentoring_block.url_name, [])
```
sarina: You ought to add in some validation here, in case someone puts in data in an unreadable format (for example swaps [] for the { } or something).

problem_builder/dashboard.py
```diff
@@ -336,20 +372,23 @@ def student_view(self, context=None):  # pylint: disable=unused-argument
...
+            for question_number, child_id in enumerate(self._get_problem_questions(mentoring_block)):
+                if question_number+1 in hide_questions:
```
sarina: why are you doing `question_number+1`?

sarina: oh indexing, right :) anyway I'd add a comment to clarify this.

sarina: _actually_ ah yes, love Python. Change the enumerate statement to just be one-indexed!!

```python
for question_number, child_id in enumerate(self._get_problem_questions(mentoring_block), 1):
```

problem_builder/dashboard.py
sarina: eg maybe,
```python
try:
    hide_questions = self.exclude_questions.get(mentoring_block.url_name, [])
except Exception:
    # log exception & what self.exclude_questions is
```

problem_builder/dashboard.py
```diff
@@ -384,6 +423,8 @@ def student_view(self, context=None):  # pylint: disable=unused-argument
             'blocks': blocks,
             'display_name': self.display_name,
             'visual_repr': visual_repr,
+            'average_label': self.average_label,
+            'show_numbers': self.show_numbers
```
sarina: Nit: trailing comma here will make future diffs nicer

sarina:
I see you have two travis builds running on this repo (you have both the push job and the pr job enabled). You can pretty easily fix this:

We noticed within the last couple of days that some Travis-run repos are starting two jobs for a pull request. (One job labeled "continuous-integration/travis-ci/pr" and another labeled "continuous-integration/travis-ci/push").

If this is happening to you, you will need the following stanza in your .travis.yml:
```
branches:
    only:
      - master
```
[Example fix here](https://github.com/edx/edx-analytics-dashboard/commit/dabcc910d25b0831de78faaceb530fe4afc81300)

sarina:
@e-kolpakov this seems reasonable. Do you have a sandbox where I can look at this?

bradenmacdonald:
@sarina We are actually deliberately running two Travis builds on this repo. One build is just for this branch (the push build), and one tests the result of the merge (the pr build). In addition, we found that the "fix" you mention above seems to sometimes prevent builds when PRs are opened against non-master branches.

sarina:
@bradenmacdonald ok, as long as you're fine with it. It was annoying on my repos, but I'm not doing lots of fancy stuff and didn't want to annoy Travis.

problem_builder/dashboard.py
```diff
@@ -172,6 +172,21 @@ class DashboardBlock(StudioEditableXBlockMixin, XBlock):
         ).format(example_here='["2754b8afc03a439693b9887b6f1d9e36", "215028f7df3d4c68b14fb5fea4da7053"]'),
         scope=Scope.settings,
     )
+    exclude_questions = Dict(
+        display_name=_("Questions to be hidden"),
+        help=_(
+            "Optional rules to exclude specific questions both from displaying in dashboard and from average "
+            "calculation. Rules must start with url_name of mentoring block, followed by list of question numbers "
+            "to exclude. Rule set must be in JSON format. Question numbers are one-based (i.e. first question "
+            "has number 1)"
```
bradenmacdonald: Put a period and space after the `)` here. It is currently appearing as "has number 1)Examples: "

problem_builder/dashboard.py
```diff
@@ -207,8 +222,23 @@ class DashboardBlock(StudioEditableXBlockMixin, XBlock):
...
         scope=Scope.content,
     )
+    average_label = String(
+        display_name=_("Label for average value"),
+        default=_("Average"),
+        help=_("Label to be shown for calculated average"),
+        scope=Scope.content,
+    )
+    show_numbers = Boolean(
+        display_name=_("Display numeric values"),
```
bradenmacdonald: I think this should just be "Display values", since the values *can* be non-numeric.

problem_builder/dashboard.py
bradenmacdonald: Also, I'd suggest you rephrase: "Optional rules to exclude specific questions both from displaying in dashboard and from **the** calculated average. Rules must start with **the** url_name of **a** mentoring block, followed by list of question numbers to exclude (**the first question being number 1**). **Must be in JSON format. Examples:** ..."

problem_builder/dashboard.py
bradenmacdonald: Cool!

problem_builder/templates/html/dashboard.html
```diff
@@ -42,16 +42,26 @@
           {% for mcq in block.mcqs %}
             <tr>
               <th class="desc">{{ mcq.display_name }}</th>
-              <td class="value" {% if mcq.color %}style="border-right-color: {{mcq.color}};"{% endif %}>
-                {% if mcq.value %}{{ mcq.value }}{% endif %}
+              <td class="value"
+                  {% if mcq.color %} style="border-right-color: {{mcq.color}};"{% endif %}
+                  {% if mcq.value and not show_numbers %} aria-label="{{mcq.value}}"{% endif %}
```
bradenmacdonald: For accessibility, when "show numbers" is False but there is no value yet, it might be nice to put aria-label="{% trans "No value yet" %}" or something?

problem_builder/templates/html/dashboard.html
```diff
@@ -42,16 +42,26 @@
...
           {% endfor %}
           {% if block.has_average %}
            <tr class="avg-row">
-              <th class="desc">{% trans "Average" %}</th>
-              <td class="value" {% if block.average_color %}style="border-right-color: {{block.average_color}};"{% endif %}>
-                {{ block.average|floatformat }}
+              <th class="desc">{{ average_label }}</th>
+              <td class="value"
+                  {% if block.average_color %} style="border-right-color: {{block.average_color}};"{% endif %}
+                  {% if block.average and not show_numbers %} aria-label="{{block.average}}"{% endif %}
```
bradenmacdonald: Should be `{{ block.average|floatformat }}` as it is below.

problem_builder/tests/integration/xml/dashboard_3.xml
```diff
@@ -0,0 +1,61 @@
...
+            <pb-choice value="B">Option B</pb-choice>
+            <pb-choice value="C">Option C</pb-choice>
+        </pb-mcq>
+    </problem-builder>
+    <!-- Test patches url_name to return display_name instead -->
+    <pb-dashboard mentoring_ids='["dummy-value"]'
+      exclude_questions='{"Step 1": [2, 3], "Step 2":[3], "Step 3":[2]}'
+    >
+    </pb-dashboard>
+</vertical_demo>
```
bradenmacdonald: Nit: missing blank line at EOF

problem_builder/tests/integration/xml/dashboard_1.xml
```diff
@@ -0,0 +1,58 @@
...
+        </pb-mcq>
+        <pb-mcq display_name="3.2 MCQ with non-numeric values"
+          question="Which option?" correct_choices='["1","2","3","4"]'>
+            <pb-choice value="A">Option A</pb-choice>
+            <pb-choice value="B">Option B</pb-choice>
+            <pb-choice value="C">Option C</pb-choice>
+        </pb-mcq>
+    </problem-builder>
+    <pb-dashboard mentoring_ids='["dummy-value"]'>
+    </pb-dashboard>
```
bradenmacdonald: These three XML files seem to be identical except for the Problem Builder block at the end.

If you want, you can just use one XML file, and make this part a template and insert the `<pb-dashboard/>` part via code before setting it as the scenario. There's also a `self.load_root_xblock` method in `SeleniumXBlockTest` that allows you to modify a block's fields via code and then refresh the scenario.

problem_builder/dashboard.py
bradenmacdonald: +1

bradenmacdonald:
@e-kolpakov I forgot to mention this earlier, but the current way that one sets "Questions to be hidden" using JSON is not very user-friendly.

*If* you have the time and are feeling ambitious: you could modify this XBlock so that its `author_view` is almost the same as the normal `student_view` but `author_view` includes a `[x] Hide` checkbox next to each question row. You could also make the "Average" row labels into text inputs in the author_view. The only tricky thing is you need to use new JavaScript to call a handler method whenever it wants to save changes. I'm not sure if that approach goes against Studio UX design goals, but it sure would be easier to use! (I've been wanting to do something similar for choices in Problem Builder, but haven't had the chance yet)

e-kolpakov:
@bradenmacdonald Making "Questions to be hidden" is a good suggestion, but this task is due tomorrow and I'm not feeling ambitious :) So please add this to backlog as an improvement.

e-kolpakov:
@sarina sandbox: not yet, I'll spin it up when I have time (probably in a matter of one hour from now)

e-kolpakov:
@sarina just a reminder, it would be great if this could be merged today, in time for upcoming release.

sarina:
Sure thing. Is there a sandbox available to take a look at it?

antoviaque:
@sarina Yup, you have the links in the PR description above - it took me some time to find them too : )

problem_builder/dashboard.py
```diff
@@ -207,8 +221,23 @@ class DashboardBlock(StudioEditableXBlockMixin, XBlock):
...
+    average_label = String(
+        display_name=_("Label for average value"),
+        default=_("Average"),
+        help=_("Label to be shown for calculated average"),
+        scope=Scope.content,
+    )
+    show_numbers = Boolean(
+        display_name=_("Display values"),
+        default=True,
+        help=_("Toggles if numeric values are dissplayed"),
```
sarina: Typo: "dissplayed"

problem_builder/templates/html/dashboard.html
```diff
@@ -42,16 +42,30 @@
           {% for mcq in block.mcqs %}
             <tr>
               <th class="desc">{{ mcq.display_name }}</th>
-              <td class="value" {% if mcq.color %}style="border-right-color: {{mcq.color}};"{% endif %}>
-                {% if mcq.value %}{{ mcq.value }}{% endif %}
+              <td class="value"
+                  {% if mcq.color %} style="border-right-color: {{mcq.color}};"{% endif %}
+                  {% if not show_numbers %}
+                    {% if mcq.value %} aria-label="{{mcq.value}}" {% else %} aria-label="{% trans 'No value yet' %}" {%endif%}
```
sarina: I wonder if the aria-label should give a little bit more information here, eg, "Score: mcq.value"

sarina: I guess we can wait for Mark's advice on this. My worry is that it's really unclear what this block represents until you stare at it a bit, and if you're not sighted and you can't see the graph, it may be even more unclear, so adding more descriptive aria-labels may help.

sarina:
@e-kolpakov :+1: to merge once the typo is addressed.

I do want @cptvitamin to take a look at this block, even if it's after it's merged. He will be back from vacation next week. Could you please keep the sandboxes up until he's gotten a chance to review?

e-kolpakov:
@sarina there's one feature left not implemented: the ability to set up average labels per mentoring (now it's one setting for dashboard). Since this already get thumbs up and release is quite close, I'll merge this PR when the build turns green, and implement that in a follow up PR (probably today, Monday for sure).

sarina:
@e-kolpakov you're going to keep the sandbox running yes?

e-kolpakov:
@sarina yes, once it's not needed it is just freed to be used later, but not shutdown/disabled/etc.

problem_builder/templates/html/dashboard.html
```diff
@@ -42,16 +42,30 @@
           {% for mcq in block.mcqs %}
             <tr>
               <th class="desc">{{ mcq.display_name }}</th>
-              <td class="value" {% if mcq.color %}style="border-right-color: {{mcq.color}};"{% endif %}>
-                {% if mcq.value %}{{ mcq.value }}{% endif %}
+              <td class="value"
+                  {% if mcq.color %} style="border-right-color: {{mcq.color}};"{% endif %}
+                  {% if not show_numbers %}
+                    {% if mcq.value %} aria-label="Score: {{mcq.value}}" {% else %} aria-label="{% trans 'No value yet' %}" {%endif%}
```
cptvitamin: @e-kolpakov I've never seen anyone try to use ```aria-label``` on a ```td``` element, and I don't blame you for trying since the [spec says it is allowed](http://www.w3.org/TR/html-aria/#document-conformance-requirements-for-use-of-aria-attributes-in-html).  However, I wanted to perform a quick implementation test for [using aria-label on a td element](http://sadecki.com/sandbox/test/aria-label-on-td.html) and I found that only Voiceover will read the ```aria-label```, and only if there is an explicit value in the ```td``` element, and it reads *both*.  If the cell is empty, it ignores the ```aria-label```.  Testing in NVDA and JAWS revealed that *only* explicit ```td``` values are read. ```aria-labels``` are ignored.  

I recommend using off-screen text (class=sr) inside the table cells in this use case.

problem_builder/templates/html/dashboard.html
```diff
@@ -42,16 +42,30 @@
...
           {% if block.has_average %}
            <tr class="avg-row">
-              <th class="desc">{% trans "Average" %}</th>
-              <td class="value" {% if block.average_color %}style="border-right-color: {{block.average_color}};"{% endif %}>
-                {{ block.average|floatformat }}
+              <th class="desc">{{ average_label }}</th>
+              <td class="value"
+                  {% if block.average_color %} style="border-right-color: {{block.average_color}};"{% endif %}
+                  {% if not show_numbers %}
+                    {% if block.average %} aria-label="Score: {{block.average|floatformat}}" {% else %} aria-label="{% trans 'No value yet' %}" {%endif%}
```
cptvitamin: @e-kolpakov see above comment

sarina: @e-kolpakov @antoviaque this is the only a11y issue found in Mark's review. Can you guys submit a PR to fix it asap? Thanks!

antoviaque: @sarina Sure! Do you want it this week, or can it be done in next week's sprint? I would prefer the later if possible, as there will also likely be some additional follow-up fixes/improvements from Harvard, this way we could do everything at once. But let me know.

Thanks for the review @cptvitamin

sarina: Next week is fine - "ASAP" as in, "as soon as prioritization possible" :)

antoviaque: @sarina Ok, perfect : ) Thank you!

PR #18, opened 2015-04-23
Hotfix for OC-639: Database conflict between mentoring and problem_builder

bradenmacdonald:
Hotfix for OC-639:

**Bug details**:
The old mentoring XBlock uses South migrations to create a table called `mentoring_answers`. The new Problem Builder XBlock uses syncdb to create a table also called `mentoring_answers`.

Unfortunately, if a new edX instance is spun up and mentoring v1 and Problem Builder are both being installed at the same time (as they are if `requirements/edx/edx-private.txt` is used), there will be a database error since the South migration will abort when it finds that the table already exists: https://gist.github.com/maxrothman/3868d5e7ad87533a287c

Because syncdb does nothing if the table already exists, an edX instance could only have both blocks installed as long as mentoring was installed and setup in the DB first.

**Fix details**:
My proposed fix is just to rename the table in Problem Builder. I've also added a data migration that copies *all* student answers from the `mentoring_answers` table to the new table, which should prevent any students who may be currently using problem builder from losing data.

**Consequences**:
* Running the upgrade script that converts from the old block to the new one will no longer automagically migrate any student answers made since problem builder was installed. However, I don't really want anyone attempting to do that upgrade on a course that students are using anyways.

**Recovery**:  
If someone encountered the error seen in the gist link above, it should be possible to recover with:
```bash
./manage.py lms migrate mentoring 0003 --fake --settings=devstack
```
(Change `devstack` to whatever configuration is being used)

### DIFF
--- a/problem_builder/models.py
+++ b/problem_builder/models.py
@@ -29,10 +29,11 @@ class Answer(models.Model):
     """
     Django model used to store AnswerBlock data that need to be shared
     and queried accross XBlock instances (workaround).
+
+    TODO: Deprecate this and move to edx-submissions
     """
 
     class Meta:
-        db_table = 'mentoring_answer'
         unique_together = (('student_id', 'course_id', 'name'),)
 
     name = models.CharField(max_length=50, db_index=True)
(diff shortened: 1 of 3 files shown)


antoviaque:
:+1: 

 @bradenmacdonald Thanks for taking on this fix!

sarina:
:+1:

PR #21, opened 2015-04-24
Fix two more bugs

bradenmacdonald:
This fixes:

* SOL-787 - the HTML editor was changing URLs in `<a` elements, breaking links. Fixed in xblock-utils; this is only a version-bump fix in this repo
* SOL-782 - when `max_attempts` was set to `0` (unlimited), the "Review Grade" button was always shown and enabled, when it shouldn't be.

I also cleaned up the assessment integration tests while I was at it, converting them from `SeleniumBaseTest` to `SeleniumXBlockTest ` so they can be parametrized and will load faster.


e-kolpakov:
@bradenmacdonald :+1:

problem_builder/tests/integration/base_test.py
```diff
@@ -57,7 +60,18 @@ def popup_check(self, mentoring, item_feedbacks, prefix='', do_submit=True):
             self.assertFalse(item_feedback_popup.is_displayed())
 
 
-class MentoringAssessmentBaseTest(MentoringBaseTest):
+class MentoringBaseTest(SeleniumBaseTest, PopupCheckMixin):
+    module_name = __name__
+    default_css_selector = 'div.mentoring'
```
e-kolpakov: @bradenmacdonald niy: moving this before `PopuCheckMixin will reduce the diff size.

e-kolpakov:
@bradenmacdonald :+1:

problem_builder/tests/integration/base_test.py
bradenmacdonald: @e-kolpakov True, but I don't like referencing a base class before it's defined.

Also, `MentoringAssessmentBaseTest` should not be a child of `MentoringBaseTest`, because `MentoringBaseTest` inherits the `setUp` method from `SeleniumBaseTest` which we don't want (It parses and loads *all* of the integration test .xml files for every test, and starts each test by loading the workbench home page - we don't really want either of those to be happening since they pointlessly slow down the tests).

PR #22, opened 2015-04-25
Data Export block prototype

bradenmacdonald:
This new **Data Export Block** provides an instructor interface for generating CSV exports of student answers to problems in the course. The actual export of answers to CSV is done by a new celery task, which integrates with the existing `ReportStore` API functionality to save the CSV file to S3.

Screenshot:

screenshot of the LMS UI

Testing:

To test with celery on devstack, configure rabbitmq then run `paver celery --settings=devstack`. Edit `lms/envs/devstack.py` to set `CELERY_ALWAYS_EAGER = False`.

### DIFF
--- a/problem_builder/data_export.py
+++ b/problem_builder/data_export.py
@@ -60,11 +60,11 @@ class DataExportBlock(XBlock):
         """ Studio View """
         # Warn the user that this block will only work from the LMS. (Since the CMS uses
         # different celery queues; our task listener is waiting for tasks on the LMS queue)
-        return Fragment(u"<p>Data Export Block</p><p>This block only works from the LMS.</p>")
+        return Fragment(u'<p>Data Export Block</p><p>This block only works from the LMS.</p>')
 
     def studio_view(self, context=None):
         """ 'Edit' form view in Studio """
-        return Fragment(u"<p>This block has no configuration options.</p>")
+        return Fragment(u'<p>This block has no configuration options.</p>')
 
     def check_pending_export(self):
         """
@@ -78,27 +78,22 @@ class DataExportBlock(XBlock):
 
     def _save_result(self, task_result):
         """ Given an AsyncResult or EagerResult, save it. """
-        self.active_export_task_id = ""
+        self.active_export_task_id = ''
         if task_result.successful():
             if isinstance(task_result.result, dict) and not task_result.result.get('error'):
                 self.last_export_result = task_result.result
             else:
-                self.last_export_result = {"error": u"Unexpected result: {}".format(repr(task_result.result))}
+                self.last_export_result = {'error': u'Unexpected result: {}'.format(repr(task_result.result))}
         else:
-            self.last_export_result = {"error": unicode(task_result.result)}
+            self.last_export_result = {'error': unicode(task_result.result)}
 
     def student_view(self, context=None):
         """ Normal View """
-        # TODO: Verify instructor permissions
-        # Check if any pending export has finished:
-        self.check_pending_export()
-        # Render our HTML:
-        html = loader.render_template('templates/html/data_export.html', {
-            'export_pending': bool(self.active_export_task_id),
-            'last_export_result': self.last_export_result,
-            'download_url': self.download_url_for_last_report,
-        })
+        if not self.user_is_staff():
+            return Fragment(u'<p>This interface can only be used by course staff.</p>')
+        html = loader.render_template('templates/html/data_export.html')
         fragment = Fragment(html)
+        fragment.add_css_url(self.runtime.local_resource_url(self, 'public/css/data_export.css'))
         fragment.add_javascript_url(self.runtime.local_resource_url(self, 'public/js/data_export.js'))
         fragment.initialize_js('DataExportBlock')
         return fragment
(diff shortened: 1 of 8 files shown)


ormsbee:
Why is this functionality baked into a specific XBlock, rather than being a more general platform feature that the XBlock makes use of?

bradenmacdonald:
@ormsbee Good question. By isolating the code now, we can move faster and test the concept before it's more generally available... We are on a tight timeframe for this work, and I don't want to make any commitments that significantly broaden the scope for this implementation. But certainly the new code components could be made a *bit* more general and placed elsewhere in platform from the get go. Do you have a suggestion for where they should live?

cpennington:
It seems odd to build this as an XBlock (which is meant to plug into course content), rather than as a separate django app (perhaps that just exposes a django-admin view). Is is just because we don't have any plugin points in the UX except for XBlock that you're building it as an XBlock?

cpennington:
(Also: it should go in a separate repo, even if it does end up as a pluggable django app)

bradenmacdonald:
@cpennington Yes, that's why. if there were another integration point that could make the export UI easily accessible to instructors, I'd be happy to use it rather than an XBlock.

bradenmacdonald:
@cpennington Do you think there's another way we could integrate this feature into the platform as a plugin, or if not, is there a relatively approachable way that we could modify the platform to support this sort of thing as a plugin? If we can find an approach that doesn't add too much work, it would be nicer to integrate this into the platform rather than as an XBlock in the course content.

CC @antoviaque

cpennington:
@andy-armstrong has been working on a similar feature here: https://github.com/edx/edx-platform/pull/7937#issuecomment-99251526. I'm not sure if it's exactly the pluggability point that you wanted, but it would be great if you could coordinate with him.

smarnach:
@bradenmacdonald This is ready for a first review now.  The integration test needs a bit of expanding, but I spent too much time on this already, so I'd like some review comments first.

bradenmacdonald:
Looks very nice, but it's not working: "Data export failed. Reason: 'student_id'". I investigated and pointed out the problem in the edx-submissions PR.

Also please be sure to fix the minor PEP8 issue, I think it's all that's preventing the build from passing.

problem_builder/data_export.py
```diff
@@ -0,0 +1,177 @@
...
+        self.check_pending_export()
+        return {
+            'export_pending': bool(self.active_export_task_id),
+            'last_export_result': self.last_export_result,
+            'download_url': self.download_url_for_last_report,
+        }
+
+    @XBlock.json_handler
+    def get_status(self, request, suffix=''):
+        return self._get_status()
```
bradenmacdonald: This could be more RESTful, with just an 'export' method that accepts GET, POST (start), and DELETE verbs. Not a big deal though.

problem_builder/public/js/data_export.js
```diff
@@ -0,0 +1,90 @@
...
+                $statusArea.append($('<p>').text(
+                    'Data export failed. Reason: ' + status.last_export_result.error
+                ));
+            } else {
+                startTime = new Date(status.last_export_result.start_timestamp * 1000);
+                $statusArea.append($('<p>').text(
+                    'A report is available for download.'
+                ));
+                $statusArea.append($('<p>').text(
+                    'It was created at ' + startTime.toUTCString() +
```
bradenmacdonald: Could this be printed in the browser's local time instead? That would be nicer.

problem_builder/tasks.py
```diff
@@ -0,0 +1,87 @@
...
+from .mcq import MCQBlock, RatingBlock
+from .sub_api import sub_api
+
+logger = get_task_logger(__name__)
+
+
+@task()
+def export_data(source_block_id_str, user_id):
+    """
+    Reruns a course in a new celery task.
```
bradenmacdonald: Whoops, this docstring needs to be updated.

problem_builder/tasks.py
```diff
@@ -0,0 +1,87 @@
+"""
+This file contains celery tasks for contentstore views
```
bradenmacdonald: out of date docstring.

bradenmacdonald:
Thanks for pushing the fix to that error I had.

+1 if you fix the docstrings, get the build passing, and use local time in the JS status view.

problem_builder/data_export.py
smarnach: You are right, that would be a nicer option.  I think it's good enough the way it is though, so I'll just leave it that way.

problem_builder/public/js/data_export.js
smarnach: Done.

problem_builder/tasks.py
smarnach: Done.

problem_builder/tasks.py
smarnach: Done.

problem_builder/tasks.py
```diff
@@ -0,0 +1,87 @@
+"""
+Celery task for CSV student anser export.
```
Kelketek: answer

smarnach: Fixed.

PR #23, opened 2015-04-28
Legacy Django migrations

e-kolpakov:
See [SOL-814](https://openedx.atlassian.net/browse/SOL-814)

This is one of two alternative solutions to the problem. Another solution is in platform [PR#436](https://github.com/edx-solutions/edx-platform/pull/436)

Advantages of this solution:
* edx-platform does not contain problem-builder related configuration.

Disadvantages of this solution:
* Problem builder migrations might become incompatible with modern Django (>=1.7)


Kelketek:
:+1: if the tests pass. This is the favored solution. Meant to comment that explanation here rather than on the hash update.

bradenmacdonald:
:+1: if tests are fixed, though I'd prefer bumping South in edx-solutions platform.

PR #24, opened 2015-04-29
Fix max attempts message

bradenmacdonald:
This fixes OC-667 (upstream bug is a comment in SOL-423): The "Message (max # attempts)" component was not working.

**Explanation**:  
`get_message()` was called before `num_attempts` was incremented, so the `max_attempts_reached` message would never be used, since `num_attempts` was always less than `max_attempts` at that point.

**Notes**:  
I wanted to just do a simple fix, which involves moving the `get_message()` method to after where `num_attempts` is updated, but the implications of that move were not at all clear due to the complexity of the "submit" handler, and uncertainty about when the various messages are even supposed to be display. So I added a bunch of new tests and documentation to clarify the expected behavior of each type of message (I hope I got it right), and I also simplified the "submit" handler (in Python at least) while I was at it, in order to make the code easier to understand.

Details:
* Fixed by moving `get_message()`
* I added seven new tests for the messages, two of which fail without this fix
* Added explanations to each type of message so that everyone is clear on the expected behaviour
* Added per-message-type default text to the messages, which improves the authoring experience significantly
* Cleaned up the `submit()` handler somewhat
* Removed `attempted` which was not used
* Change to behavior: Previously if a user submitted an answer to a PB block that "required" the user to complete another block first, the user's individual answers would be saved but the submission would not count. To me, that is a confusing in-between state. The code is much simpler if we just make the handler throw an error on any request to submit a block whose dependencies are not satisfied, so that's what I've done. We do still show the submit button though in case the prerequisite block is on the same page - otherwise the user would have to refresh.

Confusion:
* In assessment mode, "completed" means the student has gone through all the questions at least once. But in normal mode, it means that the student has gotten a perfect score. Those are totally different things, and should not be stored in the same field. For example, I'm wondering: if the student runs out of attempts in normal mode before achieving a perfect score, do we consider that "completed" ?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,3 +17,5 @@ script:
     - python run_tests.py --with-coverage --cover-package=problem_builder
 notifications:
   email: false
+addons:
+  firefox: "36.0"
--- a/problem_builder/mentoring.py
+++ b/problem_builder/mentoring.py
@@ -148,6 +148,7 @@ class MentoringBlock(XBlock, StepParentMixin, StudioEditableXBlockMixin, StudioC
         # Has the student attempted this mentoring step?
         default=False,
         scope=Scope.user_state
+        # TODO: Does anything use this 'attempted' field? May want to delete it.
     )
     completed = Boolean(
         # Has the student completed this mentoring step?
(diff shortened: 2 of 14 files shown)


bradenmacdonald:
Kelketek: I figured out how to fix the Travis build!! I do not understand why it was suddenly broken, but bumping the Firefox version has fixed it :)

Also, this is ready for your review.

Kelketek:
@bradenmacdonald Thank you for the changes on this. The extra documentation makes things a great deal more clear. I remember trying to puzzle through what all the messages were for and getting frustrated, so this will certainly help.

This works for me, and the code seems clean enough, so I'm giving it a :+1:

PR #26, opened 2015-04-30
Assessment style tweaks

antoviaque:
Style tweaks to match more closely the comp:

* Adds horizontal margin on the side of the grades results
* Moves the bottom line separator to the same block (instead of only showing it when the assessment feedback block is shown)
* Reduce the boldness of the text on assessments results

screenshot-courses mckinseyacademy local 2015-04-30 11-59-37


e-kolpakov:
@antoviaque :+1:

PR #27, opened 2015-04-30
Hide "Review final grade" link by default.

smarnach:
The link is only used during extended feedback in assessment mode.  It's hidden
on page load by JS in all cases, and only shown if the student goes back to a
question after using up all attempts, so we can just as well hide it in CSS.

This fixes MCKIN-3099.

@bradenmacdonald I can't reproduce the issue on my local Apros installation, neither with Firefox nor with Chrome.  I can reproduce it using the link and credentials in MCKIN-3099, though, so it's probably dependent on the timing of AJAX responses, or something along these lines.  Could you please try whether you are able to reproduce the issue, and whether this patch fixes it?  I'm rather confident it does.

The issue occurs when moving forward (or back) to a unit with a mentoring question in standard mode using the blue Apros navigation arrows -- see the video in MCKIN-3099.

### DIFF
--- a/problem_builder/public/css/mentoring.css
+++ b/problem_builder/public/css/mentoring.css
@@ -165,4 +165,5 @@
 
 .mentoring .review-link {
     float: right;
+    display: none;
 }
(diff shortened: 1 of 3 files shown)


bradenmacdonald:
I wasn't able to reproduce the issue but I'm confident this will fix it, and I tested to check for any loss of functionality.

:+1: if you rebase and the tests pass.

PR #25, opened 2015-04-30
Test + fix for crash when block is deleted after students have used it

bradenmacdonald:
Forward port of https://github.com/edx-solutions/xblock-mentoring/pull/112

PR #28, opened 2015-05-01
Test + fix an additional failure case

bradenmacdonald:
This is a follow up to #25 - it turns out I missed another failure mode (if you delete the last step, while the student is on that step).


Kelketek:
:+1: Works for me.

smarnach:
Same remarks as on https://github.com/edx-solutions/xblock-mentoring/pull/115, otherwise :+1:.

PR #29, opened 2015-05-05
Presentation updates, part 2

e-kolpakov:
**Fixes and improvements:**
* Give all the tables of the dashboard the same width (should be ~700px to give enough place to display long titles), and center the tables
* Increase the width of the color codes in the table, to make them squarred instead of rectangular
* Make the label for the "Average" row in the tables semi-bold (font-weight: 600)
* Change aria-labels to off-screen text, as per Mark Sadeki's accessibility review: https://github.com/open-craft/problem-builder/pull/16#discussion_r29346587
* In the export HTML:
** Use left alignment for text within the tables, which is currently shown centered<
** Apply the changes done to the table in the LMS: center the tables, use the same width for all tables (~700px - do not make it stretch to all available space, as widows can be big and it will look ugly), make the color codes squared, make the label for the &quot;Average&quot; row squarred
* In Problem Builder, allow to customize the title for the Messages, which by default reads "Feedback"

**Sandbox:** [LMS](http://sandbox3.opencraft.com/), [Studio](http://sandbox3.opencraft.com:18080)
**JIRA Ticket:** [OSPR-575](https://openedx.atlassian.net/browse/OSPR-575)
**Merge deadline:** preferably Friday, May 8 - this fixes are needed for course starting very soon.

**Test instructions:**
1. Set up a couple of mentoring XBlocks.
2. Set up Dashboard block to draw data from those blocks
3. Set custom value for "Feedback Header" for one of mentoring XBlocks, add completion message to this XBlock
4. Publish everything
5. Go to LMS, provide answers for mentoring blocks. Observe that overriding feedback messages header works - when mentoring with custom "Feedback Header" and "Completion message" is submitted, both configured header and message is shown (If "Feedback" is shown instead of overridden value - it is a bug).
6. Navigate to dashboard block

Observe the following:
* All tables are centered and have 700px width
* "Average" row caption is semibold (font-weight: 600)
* Color-codes are much wider (30px) - should be square if question text fit to one line
* Clicking "Download Report" (surprisingly) downloads a report. The report contains a copy of dashboard block visual representation and tables. All observations above applies to report content (table width, color codes and semi-bold "Average"). In addition, text in tables is left-aligned (it used to be centered).

### DIFF
--- a/problem_builder/dashboard.py
+++ b/problem_builder/dashboard.py
@@ -403,7 +403,7 @@ class DashboardBlock(StudioEditableXBlockMixin, XBlock):
                 block['mcqs'].append({
                     "display_name": mcq_block.display_name_with_default,
                     "value": value,
-                    "accessible_value": _("Score: {value}").format(value=value) if value else _("No value yet"),
+                    "accessible_value": _("Score: {score}").format(score=value) if value else _("No value yet"),
                     "color": self.color_for_value(value) if value is not None else None,
                 })
             # If the values are numeric, display an average:
@@ -415,8 +415,8 @@ class DashboardBlock(StudioEditableXBlockMixin, XBlock):
                 average_value = sum(numeric_values) / len(numeric_values)
                 block['average'] = average_value
                 # average block is shown only if average value exists, so accessible text for no data is not required
-                block['accessible_average'] = _("Score: {average_value}").format(
-                    average_value=floatformat(average_value)
+                block['accessible_average'] = _("Score: {score}").format(
+                    score=floatformat(average_value)
                 )
                 block['average_label'] = self.average_labels.get(mentoring_block.url_name, _("Average"))
                 block['has_average'] = True
--- a/problem_builder/dashboard.py
+++ b/problem_builder/dashboard.py
@@ -30,6 +30,7 @@ import ast
 import json
 import logging
 import operator as op
+from django.template.defaultfilters import floatformat
 
 from .dashboard_visual import DashboardVisualData
 from .mcq import MCQBlock
@@ -414,7 +415,9 @@ class DashboardBlock(StudioEditableXBlockMixin, XBlock):
                 average_value = sum(numeric_values) / len(numeric_values)
                 block['average'] = average_value
                 # average block is shown only if average value exists, so accessible text for no data is not required
-                block['accessible_average'] = _("Score: {average_value:.2f}").format(average_value=average_value)
+                block['accessible_average'] = _("Score: {average_value}").format(
+                    average_value=floatformat(average_value)
+                )
                 block['average_label'] = self.average_labels.get(mentoring_block.url_name, _("Average"))
                 block['has_average'] = True
                 block['average_color'] = self.color_for_value(average_value)
--- a/problem_builder/dashboard.py
+++ b/problem_builder/dashboard.py
@@ -402,6 +402,7 @@ class DashboardBlock(StudioEditableXBlockMixin, XBlock):
                 block['mcqs'].append({
                     "display_name": mcq_block.display_name_with_default,
                     "value": value,
+                    "accessible_value": _("Score: {value}").format(value=value) if value else _("No value yet"),
                     "color": self.color_for_value(value) if value is not None else None,
                 })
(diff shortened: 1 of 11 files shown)


e-kolpakov:
@sar

problem_builder/dashboard.py
```diff
@@ -402,6 +403,7 @@ def student_view(self, context=None):  # pylint: disable=unused-argument
                 block['mcqs'].append({
                     "display_name": mcq_block.display_name_with_default,
                     "value": value,
+                    "accessible_value": _("Score: {value}").format(value=value) if value else _("No value yet"),
```
sarina: perhaps "No score yet"? "value" isn't as specific ("value for what?") whereas "score" or "problem score" is much more specific.

problem_builder/dashboard.py
```diff
@@ -412,6 +414,10 @@ def student_view(self, context=None):  # pylint: disable=unused-argument
             if numeric_values:
                 average_value = sum(numeric_values) / len(numeric_values)
                 block['average'] = average_value
+                # average block is shown only if average value exists, so accessible text for no data is not required
+                block['accessible_average'] = _("Score: {average_value}").format(
```
sarina: Just a note about translations: this will be in your po file as a _separate translation string_ from the `"Score: {value}"` string above. However... why? They're the same exact string, semantically. If you match them exactly, translators will only have to translate the strings once.

problem_builder/mentoring.py
```diff
@@ -131,6 +131,12 @@ class MentoringBlock(XBlock, StepParentMixin, StudioEditableXBlockMixin, StudioC
         default=_("Mentoring Questions"),
         scope=Scope.settings
     )
+    feedback_label = String(
+        display_name=_("Feedback Messages Title"),
```
sarina: This is super awkward. I'm not sure how it's used yet until you have a sandbox but I'm going to suggest something different.

sarina: eh I think I get it now. Maybe "Feedback Title" or "Feedback Header" or "Feedback Label".

"Feedback Messages Title" is too long and awkward, and is just confusing.

sarina:
This is looking good to me. Please ping when there's a sandbox and I'll take a look there, and have Mark review your sr additions.

problem_builder/public/css/dashboard.css
```diff
@@ -24,7 +32,7 @@
     min-width: 4em;
     text-align: right;
     padding-right: 5px;
-    border-right: 0.6em solid transparent;
+    border-right: 1.875em solid transparent;
```
Kelketek: Why 1.875? If the goal is to make them square, shouldn't it be 2? 2 certainly looks more square when I try it.

Kelketek:
@e-kolpakov Please consider the note I put about the border width, but :+1: . @sarina Links to the sandbox have been added in the PR description.

sarina:
@cptvitamin can you take a look at the a11y changes here? (you reviewed https://github.com/open-craft/problem-builder/pull/16 a few days ago, this is the response)

problem_builder/public/css/dashboard.css
e-kolpakov: @Kelketek cell height is 21.6 + 2*3.2 (padding) + 2*1 (border) = 30. Font-size or line-height, or whatever is equal to 1em for the cell is 16px; 30/16 = 1.875;

I've started with 2em and doesn't look more square to me :)

problem_builder/public/css/dashboard.css
```diff
@@ -24,7 +32,7 @@
     min-width: 4em;
     text-align: right;
     padding-right: 5px;
-    border-right: 0.6em solid transparent;
+    border-right: 2em solid transparent;
```
sarina: You might want to just consider defining some baseline variables then having sizes be relative to that (eg baseline/4, baseline/10 or whatever). That way you don't have to do fussy math all the time and worry that numbers like 1.875 are correct.

Perhaps a future improvement to think about.

cptvitamin:
:+1: from me.

sarina:
:+1:

PR #30, opened 2015-05-07
Feedback persistence

e-kolpakov:


Kelketek:
@e-kolpakov This is working for me now! However, I noticed something which... probably is counter-productive to what we want, and I'm not sure how you would fix it.

Same scenario as before with the MRQ. The difference is, when I submit originally, only the answer I submitted is checked or flagged depending on whether it was right or wrong. When I reload, it shows me the entire answer key-- whether all of the options were right or wrong. That's essentially giving me the full answer.

e-kolpakov:
@Kelketek this was actually relatively easy to fix; anyway, thank you for noticing that.

Kelketek:
@e-kolpakov Looks good to me! :) :+1:

PR #32, opened 2015-05-12
Merging edx-release into master

e-kolpakov:


bradenmacdonald:
:+1: This looks good to me. I see it's now mostly the dashboard presentation changes that are in the diff and not much else. Thanks @e-kolpakov.

PR #33, opened 2015-05-15
Jump to children prototype - XBlock side

e-kolpakov:


problem_builder/mentoring.py
```diff
@@ -284,21 +285,26 @@ def student_view(self, context):
         fragment = Fragment()
         child_content = u""
 
+        activate_child_id = context.get('activate_child_id', None)
```
bradenmacdonald: If you rename this parameter to just `activate_block_id`, then you could remove this code from `mentoring.py`. As long as the parent block passes the context on to the children, you don't need to add special handling code in the parent. The child can just look for `if context.get('activate_block_id') == self.scope_ids.usage_id`. I think that would be a more flexible approach.

antoviaque: +1, nice approach. I like the fact that this would allow other blocks to alter their presentation based on the selected block, even if it's not the current block.

PR #34, opened 2015-05-18
Fixed submit button being enabled after page reload

e-kolpakov:


Kelketek:
@e-kolpakov This works! However I'm seeing a similar issue to what I was seeing before, but in reverse on the MRQ-- I hit submit, I get feedback for all answers, I reload, and get feedback for only the checkboxes I've done. This also has a weird side effect. Try this:

Create an MRQ with four answers, two correct and two incorrect, and a complete and incomplete message. Check one of the correct answers and submit. You will see the incomplete message. Then reload. You will see a checkbox next to the one correct answer you did, no X marks, and a feedback message that says 'Good job!' (or whatever you put in complete). The results should be exactly the same when submitting as when reloading.

e-kolpakov:
@Kelketek please take a look again

Kelketek:
:+1:

PR #35, opened 2015-05-19
Allow to include an HTML header and footer in reports.

smarnach:


problem_builder/dashboard.py
```diff
@@ -238,10 +238,24 @@ class DashboardBlock(StudioEditableXBlockMixin, XBlock):
         help=_("Toggles if numeric values are displayed"),
         scope=Scope.content
     )
+    header_html = String(
+        display_name=_("Header HTML"),
+        default="",
+        help=_("Additional HTML to include after the heading."),
+        multiline_editor=True,
```
bradenmacdonald: You can set this to `multiline_editor="html"` if you want it to be a WSYWIG editor. I think that would be preferable for most course authors. You should probably also add `resettable_editor=False,` which takes away the reset button (
screen shot 2015-05-19 at 6 16 05 pm ) since it's not really useful for big fields that have no default text, and the UI is nicer without it.

bradenmacdonald:
:+1: I made two minor suggestions, feel free to make them and merge, or leave it as is if you have good reason.

smarnach:
@bradenmacdonald Thanks for the suggestions, implemented.

smarnach:
Closing because I need to target edx-release instead of master.

PR #37, opened 2015-05-22
Fix upgrade error when <message> has no 'type' specified explicitly

bradenmacdonald:
This PR addresses OC-694. The issue was caused by the XML parser expecting the "type" attribute to always be present, when in fact it is optional and has a default value of "completed".

I fixed the custom xml parser so it doesn't require that attribute, and I updated the relevant test to check for this.


e-kolpakov:
@bradenmacdonald :+1:

PR #39, opened 2015-06-09
XBlock-side support for jumping to children

Kelketek:


problem_builder/beeper.py
```diff
@@ -0,0 +1,45 @@
...
+        help=_("Text to display"),
+        scope=Scope.content,
+        default="",
+    )
+    editable_fields = ('content',)
+
+    def student_view(self, context):
+        normalized_id = self.scope_ids.usage_id.replace(run=None, branch=None, version_guid=None)
+        fragment = Fragment()
+        if context.get('activate_block_id', None) == normalized_id:
```
bradenmacdonald: I know this isn't being merged, but this was not working for me anymore - until I changed it to:

```python
if context.get('activate_block_id', None) == unicode(normalized_id):
```

PR #40, opened 2015-06-12
Minor tweaks to better match the Apros theme

bradenmacdonald:
When the Apros theme is active, this makes the main header (an `h2`) match the color of Apros `<h2 class="main">` headers. It also makes the subheaders use all-caps, as Apros does.

Screenshot:
screen shot 2015-06-12 at 2 34 19 pm

### DIFF
--- a/problem_builder/public/themes/apros.css
+++ b/problem_builder/public/themes/apros.css
@@ -7,3 +7,12 @@
     padding: 4px 3px 0 3px;
     font-size: 16px;
 }
+
+.mentoring .title h2 {
+	/* Same as h2.main in Apros */
+	color: #66a5b5;
+}
+
+.mentoring h3 {
+    text-transform: uppercase;
+}
--- a/problem_builder/tests/integration/base_test.py
+++ b/problem_builder/tests/integration/base_test.py
@@ -133,9 +133,10 @@ class MentoringAssessmentBaseTest(ProblemBuilderBaseTest):
 
         return mentoring, controls
 
-    def expect_question_visible(self, number, mentoring):
-        question_text = self.question_text(number)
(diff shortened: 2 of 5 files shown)


Kelketek:
@bradenmacdonald These changes look good, but the tests are failing. They don't look like they should be related, but do we know why they are?

bradenmacdonald:
@Kelketek Yeah, I think it's unrelated. I assumed it was the issue[s] that @smarnach was working on, so I didn't look into it too much. @smarnach can you please advise if these failures here are related to something that you're working to fix, or should I investigate them?

smarnach:
@bradenmacdonald These are not issues I've been working on, though I've seen the three test failures with the message `'checkmark-correct' not found` before.  However, I haven't seen the timeout yet.

There were many more test failures which have been fixed by the xblock-sdk PR.  The PR that is still pending is necessary to get the dependencies consistent again, but it won't fix any further tests.

I planned to look into the `'checkmark-correct' not found` issues, but didn't find the time to do so yet, so I wouldn't mind if you could take this on.

bradenmacdonald:
@Kelketek This should be ready now. I got the build passing again, after a lot of frustrating debugging (since the failures only happened on Travis and never locally). Long story short is most of the failures were caused by the test reloading the page and then running tests before the page's JavaScript had finished initializing everything on the page. No idea why those errors were suddenly showing up now and not before.

I also bumped the xblock-utils and xblock versions to match Sven's recent version bump commits in other places.

Kelketek:
@bradenmacdonald :+1: @smarnach Will you be able to rebase upon these changes as soon as @bradenmacdonald finishes them, so that all the builds are green?

PR #41, opened 2015-06-19
Extended export

Kelketek:
Adds in the following functionality:

1. Long answer blocks are now exported.
2. User can now filter export results by username, block type, or block ID.

### DIFF
--- a/problem_builder/data_export.py
+++ b/problem_builder/data_export.py
@@ -28,7 +28,6 @@ from xblock.exceptions import JsonHandlerError
 from xblock.fields import Scope, String, Dict
 from xblock.fragment import Fragment
 from xblockutils.resources import ResourceLoader
-from problem_builder.sub_api import SubmittingXBlockMixin
 
 loader = ResourceLoader(__name__)
 
@@ -40,7 +39,7 @@ def _(text):
 
 @XBlock.needs("i18n")
 @XBlock.wants('user')
-class DataExportBlock(SubmittingXBlockMixin, XBlock):
+class DataExportBlock(XBlock):
     """
     DataExportBlock: An XBlock for instructors to export student answers from a course.
 
@@ -162,12 +161,11 @@ class DataExportBlock(SubmittingXBlockMixin, XBlock):
         username = data.get('username', None)
         root_block_id = data.get('root_block_id', None)
         if not root_block_id:
-            root_block_id = unicode(self.scope_ids.usage_id)
+            root_block_id = self.scope_ids.usage_id
+            # Block ID not in workbench runtime.
+            root_block_id = unicode(getattr(root_block_id, 'block_id', root_block_id))
             get_root = True
         else:
-            from xmodule.modulestore.django import modulestore
-            block = modulestore().get_items(self.runtime.course_id, qualifiers={'name': root_block_id}, depth=0)[0]
-            root_block_id = unicode(block.location)
             get_root = False
         user_service = self.runtime.service(self, 'user')
         if not self.user_is_staff():
@@ -182,7 +180,12 @@ class DataExportBlock(SubmittingXBlockMixin, XBlock):
                 self.raise_error(404, _("Could not find the specified username."))
 
         async_result = export_data_task.delay(
-            root_block_id, block_types, user_id, get_root=get_root,
+            # course_id not available in workbench.
+            unicode(getattr(self.runtime, 'course_id', 'course_id')),
+            root_block_id,
+            block_types,
+            user_id,
+            get_root=get_root,
         )
         if async_result.ready():
             # In development mode, the task may have executed synchronously.
--- a/problem_builder/data_export.py
+++ b/problem_builder/data_export.py
@@ -24,6 +24,7 @@ All processing is done offline.
 """
 import json
 from xblock.core import XBlock
+from xblock.exceptions import JsonHandlerError
 from xblock.fields import Scope, String, Dict
 from xblock.fragment import Fragment
 from xblockutils.resources import ResourceLoader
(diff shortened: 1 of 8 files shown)


Kelketek:
@bradenmacdonald This is ready for your review.

Kelketek:
@bradenmacdonald Wanted to let you know I've updated this and now it can take a normal block ID without the whole full detail of a usage key. Good hint on the `jump_to_id` front.

problem_builder/public/js/data_export.js
```diff
@@ -42,49 +61,76 @@ function DataExportBlock(runtime, element) {
...
+                            'seconds': status.last_export_result.generation_time_s.toFixed(1)
+                        }
+                    )
                 ));
             }
         } else {
             if (status.export_pending) {
                 $statusArea.append($('<p>').text(
-                    'The report is currently being generated…'
+                    gettext('The report is currently being generated…')
```
bradenmacdonald: Nit: edX generally prefers not to include ellipses in messages, I think because they can be hard to translate.

problem_builder/public/js/data_export.js
```diff
@@ -42,49 +61,76 @@ function DataExportBlock(runtime, element) {
...
+
+    function addHandler($button, handlerName, form_submit) {
         $button.on('click', function() {
+            var data;
+            if (form_submit) {
+                data = {};
+                data['block_types'] = $blockTypes.val();
+                data['block_types'] = $blockTypes.val();
+                data['root_block_id'] = $rootBlockIds.val();
+                data['username'] = $username.val();
```
bradenmacdonald: Tiny tiny nit: You could write this more simply using dict literal syntax (`data = { block_types: $blockTypes.val(), ... }`)

problem_builder/data_export.py
```diff
@@ -131,13 +156,34 @@ def _delete_export(self):
...
         """ Start a new asynchronous export """
+        block_types = data.get('block_types', None)
+        username = data.get('username', None)
+        root_block_id = data.get('root_block_id', None)
+        if not root_block_id:
+            root_block_id = unicode(self.scope_ids.usage_id)
+            get_root = True
+        else:
+            from xmodule.modulestore.django import modulestore
+            block = modulestore().get_items(self.runtime.course_id, qualifiers={'name': root_block_id}, depth=0)[0]
```
bradenmacdonald: I think it would be best to push this logic into `export_data_task`. XBlocks are not supposed to know about the modulestore, and all of the other modulestore logic is currently in that task.

problem_builder/tasks.py
```diff
@@ -6,40 +6,52 @@
...
+        # Get the root block for the course.
+        while root.parent:
+            root = root.get_parent()
+
+    type_map = {cls.__name__: cls for cls in [MCQBlock, RatingBlock, AnswerBlock]}
+
+    if not block_types:
+        block_types = tuple(type_map.values())
+    else:
+        block_types = tuple([type_map[class_name] for class_name in block_types])
```
bradenmacdonald: Nit: You can remove `[` and `]` so this is a generator expression (which is more efficient) rather than a list comprehension. Not significantly different here but it's a good habit.

problem_builder/data_export.py
```diff
@@ -24,20 +24,34 @@
...
 
+# Make '_' a no-op so we can scrape strings
+def _(text):
+    return text
+
+
+@XBlock.needs("i18n")
 @XBlock.wants('user')
-class DataExportBlock(XBlock):
+class DataExportBlock(SubmittingXBlockMixin, XBlock):
```
bradenmacdonald: I don't think this class needs `SubmittingXBlockMixin` ?

problem_builder/templates/html/data_export.html
```diff
@@ -1,12 +1,40 @@
...
-  <button class="data-export-delete">Delete result</button>
+  <button class="data-export-download">{% trans "Download result" %}</button>
+  <button class="data-export-cancel">{% trans "Cancel current export" %}</button>
+  <button class="data-export-delete">{% trans "Delete result" %}</button>
 </div>
+
+<div class="export-options">
+  <div>
+    <label>
+      {% trans "Select which blocks to filter by (selecting none will grab all types):" %}
```
bradenmacdonald: Nit: "Filter" is a bit technical; I would suggest "Select which type of problems to include"

You could also consider starting with all types initially selected, which may be less confusing than saying that selecting none is the same as selecting all.

problem_builder/templates/html/data_export.html
```diff
@@ -1,12 +1,40 @@
...
+  <div>
+    <label>
+      {% trans "Input the username of a student if you wish to query for a specific one. Otherwise, it will grab all results for all students." %}
+      <input type="text" name="username" />
+    </label>
+  </div>
+</div>
+<div class="data-export-actions">
+    <button class="data-export-start">{% trans "Start a new export" %}</button>
+</div>
```
bradenmacdonald: Missing EOL

problem_builder/tests/integration/test_data_export.py
```diff
@@ -82,3 +81,11 @@ def test_data_export(self):
         self.wait_until_visible(download_button)
         self.wait_until_visible(delete_button)
         self.assertIn('A report is available for download.', status_area.text)
+
+    def test_non_staff_disabled(self):
+        data_export = self.go_to_view()
+        self.assertRaises(NoSuchElementException, data_export.find_element_by_class_name, 'data-export-start')
```
bradenmacdonald: Could also do `self.assertEqual(len(data_export.find_elements_by_class_name('data-export-start')), 0)`

bradenmacdonald:
The UI needs a bit of love, I think. It's very crowded and uneven:
screen shot 2015-06-22 at 11 39 06 pm

I think if you just put a line break before each of the input elements, it would go a long way. Even better would be to add a label for each input, and move the current verbose descriptions to become a smaller secondary help text.

bradenmacdonald:
@Kelketek I left you some notes, but they're pretty minor. This is a nice improvement and it worked flawlessly in my tests :)

problem_builder/data_export.py
Kelketek: @bradenmacdonald I considered that, but I ran into a little snag with the environment differences between the SDK and the LMS (which prompted some concerns about guarantees in the runtime). In this section, I want to get the ID of the intended root block, but not the full usage key, so that the code down the line could be the same. But I'd have to extract that ID from the usage ID there. In the workbench, the ID is unicode. In the LMS, the ID is a Usage Key. So in the LMS, I can get the block ID from that usage key, but in the workbench, that concept doesn't work.

Kelketek: I think I know a way around this, so nevermind.

problem_builder/tests/integration/test_data_export.py
Kelketek: I like this version since it feels more explanatory. You didn't label this nit, so I want to know if it's important to you.

bradenmacdonald: Nope, feel free to leave it. I don't think that's the usual way to code it, but I don't have any objections :)

Kelketek:
@bradenmacdonald Waiting on the tests, but this should be ready for you now.

bradenmacdonald:
:+1:

PR #42, opened 2015-07-07
Export bugfixes

Kelketek:
@mtyaka @itsjeyd 

This PR contains two commits-- one of which is for the issue of missing children causing a crash, and one may resolve the issue where the download button was hidden.

A few notes:

1. I don't have access to the Harvard instance, and I can't reproduce this locally, so this fix is my best guess. I might have an easier time of it if I can see it in person. @mtyaka , would you be able to show me where this is and get me set up with it?
2. The primary fix for the download button is fixing a conditional that would always be true (objects only ever equal themselves in Javascript. Deep equality is not checked). I believe that this assumption might cause a race condition whereby the updateStatus information might be updated twice rapidly with two different sets of information.
3. A (probably unneccessary-- @bradenmacdonald can you comment?) additional change was forcing a save of the fields on the XBlock before firing off the task, lest the task end before the fields were saved and the result be overwritten. I don't know if this is a realistic fear, but because I couldn't reproduce the issue and it seemed plausible that this sort of thing might cause the issue, I added that as well.

### DIFF
--- a/problem_builder/data_export.py
+++ b/problem_builder/data_export.py
@@ -172,6 +172,8 @@ class DataExportBlock(XBlock):
             return {'error': 'permission denied'}
         from .tasks import export_data as export_data_task  # Import here since this is edX LMS specific
         self._delete_export()
+        # Make sure we nail down our state before sending off an asynchronous task.
+        self.save()
         if not username:
             user_id = None
         else:
--- a/problem_builder/public/js/data_export.js
+++ b/problem_builder/public/js/data_export.js
@@ -26,7 +26,7 @@ function DataExportBlock(runtime, element) {
     }
 
     function updateStatus(newStatus) {
-        var statusChanged = newStatus !== status;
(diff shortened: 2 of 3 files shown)


itsjeyd:
The fix for missing blocks looks good. I applied the patch from 4dcb2aa and was able to successfully export answer data for the demo course.

Good catch with the equality check, it's definitely better to wait for the status to *actually* change before updating the UI. However, at this point I can't definitively confirm that 8d6fcbf fixes the button issue because I haven't been able to reproduce that issue locally.

mtyaka:
@Kelketek @itsjeyd The instance is available at https://openedx.gse.harvard.edu. If you create an account, I can grant you admin/superuser permissions.

itsjeyd:
@mtyaka Thanks, I registered as `tikr` (using my OpenCraft email address).

itsjeyd:
As discussed on IRCCloud and JIRA, this seems to be a configuration issue.

:+1:

PR #43, opened 2015-07-08
Allow students to download a copy of their answer table.

Kelketek:


Kelketek:
@e-kolpakov I'm going to break down these changes into two PRs, because I intend to pause this task to work on the rebase task and come back to it, and this change is whole enough on its own to warrant review (Doing both at once would get pretty large)

Kelketek:
@e-kolpakov Since you didn't review this earlier, I went ahead and appended the sharing feature to it. Ready for your review.

problem_builder/table.py
```diff
@@ -66,23 +74,122 @@ class MentoringTableBlock(StudioEditableXBlockMixin, StudioContainerXBlockMixin,
...
+        }
+
+    @XBlock.json_handler
+    def share_results(self, data, suffix=''):
+        target_usernames = data.get('usernames')
+        target_usernames = [username.strip().lower() for username in target_usernames if username.strip()]
+        current_username = self.current_user_key
+        if not target_usernames:
+            raise JsonHandlerError(400, _('Usernames not provided.'))
+        shared_with = self.shared_with.get(current_username, [])
```
e-kolpakov: @Kelketek shared_with and view_options could be sets - in such case it wouldn't be necessary to check if it's already there - just add to a set.

problem_builder/table.py
```diff
@@ -66,23 +74,122 @@ class MentoringTableBlock(StudioEditableXBlockMixin, StudioContainerXBlockMixin,
...
+
+    @XBlock.json_handler
+    def remove_share(self, data, suffix=''):
+        target_username = data.get('username')
+        if not target_username:
+            raise JsonHandlerError(400, _('Username not provided.'))
+        current_username = self.current_user_key
+        view_options = self.view_options.get(target_username, [])
+        shared_with = self.shared_with.get(current_username, [])
+        view_options = [option for option in view_options if option != current_username]
```
e-kolpakov: @Kelketek if view_options and shared_with were sets it would be enough to do `view_options -= {current_username}`

e-kolpakov:
@Kelketek aside from two improvement suggestions, there are a couple of possible bugs:

* Downloaded report still shows "Share" button. It's not working though. I suppose it shouldn't even be there.
* Sharing is kind of flaky. I haven't got neither the reason nor steps to reproduce, but quite often switching "Display From" does not change the answers. To be precise, it sends ajax request with correct username and successfully parses it. The problem is that response contains just the same data as before (most likely the user's own answers). Not sure how to replicate it.

Also, it was a bit unexpected that "Share" dialog didn't close when clicked outside of it. Maybe I'm got a bit too used to closing popups by clicking anywhere :)

bradenmacdonald:
@Kelketek I'm still looking at this but here's some early feedback:

1. Tests are passing but you've got some pylint and PEP8 warnings.
2. Can the "Share" button be made configurable like the download ?
3. Shouldn't the Share and Download links be in the same place?

bradenmacdonald:
@Kelketek This may be a minor issue, but:

In the LMS, there's some CSS rule in Problem Builder that makes the columns of the table even in width (I can't remember what/where at the moment unfortunately):
screen shot 2015-07-16 at 2 36 36 pm

But in the downloaded report, the CSS is different and the widths are uneven:
screen shot 2015-07-16 at 2 37 59 pm

Also nit: There should be a bit of space between "Date: blah" and the table in the downloaded report.

bradenmacdonald:
@Kelketek The Share UI has zero instructions and is super confusing:
screen shot 2015-07-16 at 2 39 59 pm

Is this going to post to Twitter? What is this for? What am I supposed to type into the box? Do I click "+" or "Share" once I've entered something? I think at the least it needs to explain what the sharing does and have a placeholder attribute on the text input saying "Enter a username" or something.

bradenmacdonald:
Ooh, this is nice. It has words and everything!
screen shot 2015-07-16 at 2 52 06 pm

problem_builder/table.py
```diff
@@ -66,23 +75,148 @@ class MentoringTableBlock(StudioEditableXBlockMixin, StudioContainerXBlockMixin,
...
+    allow_sharing = Boolean(
+        display_name=_("Allow Sharing"),
+        help=_("Allow students to share their results with other students."),
+        default=True,
+        scope=Scope.content
+    )
+    # Stores who a user has allowed to see their table.
+    shared_with = Dict(scope=Scope.user_state_summary)
+    # Stores whose tables a user is allowed to view. This needs to be kept in sync with the previous field.
+    view_options = Dict(scope=Scope.user_state_summary)
```
bradenmacdonald: This might need to be changed :-/

`Scope.user_state_summary` fields have no locking so if two students make changes at the same time, the second one will overwrite the first one. It also could easily lead to these two fields being out of sync, if a different student "wins" for each write.

Quote from Dave when I proposed something similar: "I would strongly discourage you from using Scope.user_state_summary for this purpose. As you point out, that scope was not meant to be used for items with high concurrency needs, and write conflicts will happen."

bradenmacdonald: Since we're in a time crunch I would suggest storing this data in a Django table instead. And get rid of the username->anonymous ID map table; you can include that data in the new shares table if necessary.

bradenmacdonald:
:sparkles: :+1: :sparkles:

antoviaque:
@Kelketek Thanks for getting this merged in time! Hope you have good holidays : )

@bradenmacdonald I read quickly on IRC this morning that you had concerns about the UX on this story. Is it still the case or were they addressed as part of the review? If there are still concerns, can you let me know what you think needs to be improved?

bradenmacdonald:
@antoviaque I did but @Kelketek fixed my main concern, which was that there was no explanation or instructions in the UI. That's now been addressed during the review.

I do have one remaining minor concern which is that the "+" button behaviour for sharing with multiple people is a bit strange (when you click the "+" button, you get another one, so there are now two "+" buttons, and you can get many "+" buttons appearing even though only one is ever needed). But that's very minor.

antoviaque:
@bradenmacdonald OK, great, thanks. And agreed about the '+' button, it's true that it would have been better to only keep one, next to the last input field. But yes, it's minor.

PR #44, opened 2015-07-12
Implement tooltips/clarifications.

mtyaka:
The idea is based on capa `<clarification>` nodes: https://github.com/edx/edx-platform/pull/6679

I decided to go with `<span class="pb-clarification>` instead of a custom `<clarification>` node because supporting clarifications inside the `html` child block was a requirement, but the `html` block is defined inside edx-platform and we aren't able to preprocess the contents of the `html` block from within problem-builder.

Even if we were able to get a patch for html block to support `<clarification>` elements upstream, I wouldn't feel very comfortable with that idea since the html block is supposed to contain plain html and making it recognize and preprocess custom elements wouldn't feel right.

The best way to implement this would probably be custom DOM elements (`document.registerElement`),
but browser support is not there yet and there are no reliable polyfills.

This code uses JS to process `<span class="pb-clarification">` elements and transform them into the same HTML structure that's used by capa `<clarification>` tooltips. It relies on the LMS/Studio JS code to handle mouseovers/clicks on the tooltips, which means the tooltips don't work quite correctly in the workbench - if workbench support is required, we could copy the tooltip JS code and styles from edx-platform.

To test this, use `<span class="pb-clarification">My clarification</span>` inside a question of any problem type. Clarifications are also supported in MCQ/MRQ choices.


problem_builder/public/css/problem-builder.css
```diff
@@ -0,0 +1,7 @@
+.pb-clarification span.clarification i {
+    font-style: normal;
+}
+
+.pb-clarification span.clarification i:hover {
+    color: rgb(0, 159, 230);
+}
```
bradenmacdonald: Can this be combined with `mentoring.css` ? Mentoring.css should really be renamed to `problem_builder.css` since it applies to the block as a whole. This would be a good opportunity to do so, and would reduce the number of separate CSS/JS files we are sending to the browser.

Also nit: missing newline at EOF

problem_builder/public/js/problem_builder.js
```diff
@@ -0,0 +1,40 @@
...
+                    Array.prototype.forEach.call(mutation.addedNodes, function(node) {
+                        transformExisting(node);
+                    });
+                })
+            });
+            observer.observe(element[0], {childList: true, subtree: true});
+        }
+    }
+
+};
```
bradenmacdonald: Same question: can this be combined with `mentoring.js` ? (Ideally the final would be named `problem_builder.js`)

problem_builder/mcq.py
```diff
@@ -123,7 +123,11 @@ def author_edit_view(self, context):
         fragment = Fragment(u"<p>{}</p>".format(self.question))
         self.render_children(context, fragment, can_reorder=True, can_add=False)
         fragment.add_content(loader.render_template('templates/html/questionnaire_add_buttons.html', {}))
+        fragment.add_css_url(self.runtime.local_resource_url(self, 'public/css/problem-builder.css'))
         fragment.add_css_url(self.runtime.local_resource_url(self, 'public/css/questionnaire-edit.css'))
+        fragment.add_javascript_url(self.runtime.local_resource_url(self, 'public/js/problem_builder.js'))
+        fragment.add_javascript_url(self.runtime.local_resource_url(self, 'public/js/questionnaire_edit.js'))
+        fragment.initialize_js('QuestionnaireEdit')
```
bradenmacdonald: Do you think there's a way to modify the `QuestionnaireAbstractBlock.author_edit_view` so that these subclasses can just call the superclass method with minor modifications? Right now, there is a lot of repeated code in the various `author_edit_view` implementations.

problem_builder/tests/integration/test_clarifications.py
```diff
@@ -0,0 +1,100 @@
...
+
+    mcq_template = """
+        <problem-builder>
+            <pb-mcq question="Who was your favorite character? {clarify}">
+                <pb-choice value="gaius">Gaius Baltar</pb-choice>
+                <pb-choice value="adama">Admiral William Adama {clarify}</pb-choice>
+                <pb-choice value="starbuck">Starbuck</pb-choice>
+            </pb-mcq>
+        </problem-builder>
+    """
```
bradenmacdonald: If you want, you can put these templates into the `xml_templates` folder, and change the placeholders to `{{django_template syntax}}`, as some of the other tests do. But this way is fine too; personally I'd probably leave it as-is.

problem_builder/tests/integration/test_clarifications.py
```diff
@@ -0,0 +1,100 @@
...
+                <pb-choice value="6">More than 5 stars {clarify}</pb-choice>
+            </pb-rating>
+        </problem-builder>
+    """
+
+    long_answer_template = """
+        <problem-builder>
+            <pb-answer question="What did you think {clarify} of the ending?" />
+        </problem-builder>
+    """
```
bradenmacdonald: Can you please add a scenario template that includes `<html_demo>` tags, so that the test confirms the changes work in HTML children?

problem_builder/public/js/problem_builder.js
```diff
@@ -0,0 +1,40 @@
+window.ProblemBuilder = {
+
+    transformClarifications: function(element) {
+        element = $(element);
```
bradenmacdonald: Nit: I'd prefer `var $element = $(element);` for clarity. Actually, on second look, I think you should actually remove this line. The call to `transformExisting` will work fine with a DOM node rather than a jQuery object, and the only other use of `element` here actually needs a DOM node not a jQuery element.

bradenmacdonald:
@mtyaka This looks great. I wasn't able to think of a better approach than what you've done here. I've left some minor notes and a request for you to add a test with an HTML block. If you add that test I'll give you thumbs up; the rest of my notes are optional/nits.

Here's a screenshot of some tests I just did:
screen shot 2015-07-13 at 10 10 53 pm

One final question: When I edit an HTML block, it doesn't look so great because the `<span>` element has no special styling. The screenshot below is the WYSIWYG editor for the first block in the above screenshot. Did you look if there's any way to fix that?
screen shot 2015-07-13 at 10 13 54 pm

problem_builder/public/css/problem-builder.css
mtyaka: @bradenmacdonald That's a good idea, I moved the CSS rules into `mentoring.css` and renamed it to `problem_builder.css`.

problem_builder/public/js/problem_builder.js
mtyaka: @bradenmacdonald The problem is that we need to include the `transformClarifications` code in places where we don't need the `MentoringBlock`. I could add the `transformClarifications` code to `mentoring.js`, but since we keep code for different blocks in separate files, it doesn't feel like the best idea to combine this utility code with the `MentoringBlock` constructor.

I renamed `problem_builder.js` to `util.js` though, naming it `problem_builder.js` was a bad idea.

problem_builder/mcq.py
mtyaka: @bradenmacdonald Thanks for the idea, I pushed some changes to reduce the duplication.

problem_builder/tests/integration/test_clarifications.py
mtyaka: @bradenmacdonald Not sure how I forgot to add a test for the html block; thanks for noticing it's missing.

problem_builder/public/js/problem_builder.js
mtyaka: @bradenmacdonald Yeah, `var $element = $(element)` makes it more clear. The problem that I've run into is that xblock JS init functions are not invoked consistently in Studio/LMS/workbench. Sometimes they are initialized with a jquery object and sometimes with a plain DOM node. While `transformExisting` works with either, the mutation observer requires a plain DOM node, so the `var $element = $(element)` is just to ensure that `$element` is of consistent type.

This should probably be fixed in Studio/LMS to make xblock initialization consistent, but I didn't have time to debug this further.

mtyaka:
Thanks @bradenmacdonald! I addressed most of your comments. I also added a CSS file that gets included in the HTML editor to make the clarifications in the editor stand out a bit. Making the tooltips work inside the HTML editor would be a lot of work, so this only tweaks the style of `<span class="pb-clarification">` elements s bit.

Do you want to take another look?

problem_builder/public/js/problem_builder.js
bradenmacdonald: Ah, ok. I figured you probably had a good reason for it, but it doesn't hurt to ask :)

bradenmacdonald:
:shipit: ( :+1: )

PR #45, opened 2015-07-13
Instructor Tool XBlock - Polish (OC-751)

itsjeyd:
This PR renames the "Data Export" XBlock to "Instructor Tool". It also improves and extends existing functionality. It is now possible to:

1. View the data about to be exported directly within the XBlock. Answers are shown in a paginated table that includes the following columns: section name, subsection name, unit name, question type, question, answer, username. The content of exported files matches what is displayed on the page.
2. Filter answers by content. This works for MCQs, rating questions, and long answers.

Also, the UI has been overhauled completely. It now looks like this:

Instructor Tool with pagination

---

### Testing

From Studio:

1. Add `pb-instructor-tool` to advanced modules.
2. Add one or more MCQs/rating questions/long answer blocks to a unit.
3. Add an instructor tool block to the same or another unit.

From the LMS:

1. Provide answers to MCQ/rating questions/long answers blocks.
2. Switch to instructor tool and click "Search" (optionally using filtering options to narrow down results).


itsjeyd:
@bradenmacdonald Please review.

The last build has a single test that's failing. I'm not sure why -- I ran the full test suite locally and was unable to reproduce the failure.

CC @antoviaque

setup.py
```diff
@@ -54,7 +54,7 @@ def package_data(pkg, root_list):
     'pb-choice = problem_builder:ChoiceBlock',
 
     'pb-dashboard = problem_builder:DashboardBlock',
-    'pb-data-export = problem_builder:DataExportBlock',
```
bradenmacdonald: Can you please leave `pb-dashboard` in with a comment saying `deprecated`, so that existing courses will keep working?

LICENSE.MIT
```diff
@@ -0,0 +1,30 @@
+------------------------------------------------------------------------------
+This license applies to the following third-party libraries included
+in this repository:
+
+  - backbone.paginator
+  - Backbone.js
+  - Underscore.js
```
bradenmacdonald: @itsjeyd Is it necessary to include these? Backbone and Underscore should both be available in Studio and the LMS:

```
RequireJS.require(['backbone', 'underscore'], function(backbone, _) {
    console.log(backbone.VERSION, _.VERSION);
});
> 1.0.0, 1.4.4
```

bradenmacdonald:
Is "Student Answers Dashboard" the best name for this? I think it's confusing because there's another  "Dashboard" block in this repo already, which also displays a student's answers.

problem_builder/public/js/student_answers_dashboard.js
```diff
@@ -0,0 +1,299 @@
...
+
+    }, { properties: ['section', 'subsection', 'unit', 'type', 'question', 'answer', 'username'] });
+
+    var Results = Backbone.PageableCollection.extend({
+
+        model: Result,
+
+        getCurrentPage: function(object) {
+            var currentPage = this.state.currentPage;
+            if (object) {
```
bradenmacdonald: `object` is a confusing name for this parameter - it sounds like this method is expecting an object as its parameter. How about `fetchObject` or `returnObject` or `wantsObject`?

problem_builder/public/js/student_answers_dashboard.js
```diff
@@ -0,0 +1,299 @@
...
+            return this;
+        },
+
+        _insertRecords: function(records) {
+            var tbody = this.$('tbody');
+            tbody.empty();
+            records.each(function(result, index) {
+                var row = $('<tr>');
+                if (index % 2 === 0) {
+                    row.addClass('even');
```
bradenmacdonald: I think it would be better to use a `tr:nth-child(2)` CSS selector than to do even-row striping using JavaScript.

problem_builder/public/js/student_answers_dashboard.js
```diff
@@ -0,0 +1,299 @@
...
+        },
+
+        events: {
+            'click #first-page': '_firstPage',
+            'click #prev-page': '_prevPage',
+            'click #next-page': '_nextPage',
+            'click #last-page': '_lastPage'
+        },
+
+        _firstPage: function() {
```
bradenmacdonald: nit: Could these event handlers not be considered part of a public API? (i.e. do these four functions need to start with an underscore?)

problem_builder/public/js/student_answers_dashboard.js
```diff
@@ -0,0 +1,299 @@
...
+            } else {
+                if (currentPage === 1) {
+                    this._disable(backward);
+                    this._enable(forward);
+                } else if (currentPage === totalPages) {
+                    this._enable(backward);
+                    this._disable(forward);
+                } else {
+                    this._enable(all);
+                }
```
bradenmacdonald: Nit: If you get rid of `_disable` and make `_enable` accept a boolean parameter, you could simplify this:

```
this._enable(backward, currentPage > 1);
this._enable(forward, currentPage < totalPages);
```

bradenmacdonald:
FYI, I clicked the button to make Travis run the build again and now it's passing. That error you were seeing is a weird error that occurs randomly in the Travis environment and I haven't yet figured out the cause. Usually you can just run the test again and it will pass.

problem_builder/public/js/student_answers_dashboard.js
```diff
@@ -0,0 +1,299 @@
...
+
+    function showResults() {
+        $resultTable.show();
+    }
+
+    function maybeShowResults() {
+        if (status.last_export_result) {
+            showResults();
+        }
+    }
```
bradenmacdonald: Nit: I think you can combine this with `showResults`, since there doesn't seem to be any time when you want to force the resultTable to appear before the last_export_result is ready.

problem_builder/tasks.py
```diff
@@ -108,4 +106,60 @@ def scan_for_blocks(block):
...
+
+    # Username
+    user = user_by_anonymous_id(student_id)
+    username = user.username
+
+    # Question
+    student_item = StudentItem.objects.get(pk=submission['student_item'])
+
+    block_key = UsageKey.from_string(student_item.item_id)
+    block = modulestore().get_item(block_key)
```
bradenmacdonald: You should not use the `StudentItem` class since it's essentially a private class within the submissions API. I don't think you actually need any of the three lines above - if you look at where you're calling `_extract_data_for_display()`, you already have the XBlock loaded as `block`, so there is no need to go through the student_item>block_key>block lookup, which is inefficient. Just pass `block` in as a parameter to `_extract_data_for_display`.

problem_builder/tasks.py
```diff
@@ -108,4 +106,60 @@ def scan_for_blocks(block):
...
+
+    # Unit
+    mentoring_block = modulestore().get_item(block.parent)
+    unit = modulestore().get_item(mentoring_block.parent)
+
+    # Subsection
+    subsection = modulestore().get_item(unit.parent)
+
+    # Section
+    section = modulestore().get_item(subsection.parent)
```
bradenmacdonald: A couple of notes here:

1. This method is called for each block and for each student. So if you have one problem block that has 10,000 students, this method will be called 10,000 times. But the `mentoring_block`, `unit`, `subsection`, `section` data is the same for each block. Since the `get_item()` call is expensive, you should move all of these lookups to be outside of the loop, so they are only queried once per block, and not once per block-student pair.

2. The unit is not necessarily the direct parent of the problem block. Configurations like the following are possible:

    ```
Course > section (chapter) > subsection (sequential) > unit (vertical) > A/B Test > Problem Builder > [MCQ/Answer/etc.]

Course > section (chapter) > subsection (sequential) > unit (vertical) > vertical > vertical > Library Content Block > vertical > Problem Builder > [MCQ/Answer/etc.]
    ```
    A more reliable way to code this would be:
   
    ```python
    block_names_by_type = {}
    block_iter = block
    while block_iter:
        block_iter_type = block_iter.scope_ids.block_type
        block_names_by_type[block_iter_type] = block.display_name_with_default
        block_iter = block_iter.get_parent() if block_iter.parent else None
    section_name = block_names_by_type.get('chapter', '')
    subsection_name = block_names_by_type.get('sequential', '')
    unit_name = block_names_by_type.get('vertical', '')
    ```
   (and do that outside of this method)

problem_builder/tasks.py
```diff
@@ -108,4 +106,60 @@ def scan_for_blocks(block):
         "report_filename": filename,
         "start_timestamp": start_timestamp,
         "generation_time_s": generation_time_s,
+        "display_data": [] if len(rows) == 1 else rows[1:]
```
bradenmacdonald: If this is used in a course that has dozens of problems and thousands of students, large data sets are easily possible. I think you should add a maximum size limit here. If there are >1,000 rows, it's probably quite inefficient to send them via celery to the LMS to the browser and display them using JavaScript. In such a case, we should just show a preview of the data and get them to download the CSV in order to see the whole thing.

bradenmacdonald:
@itsjeyd Nice work! I couldn't find any bugs and it works very well. Please rebase this code (since there's now a merge conflict) and address the issues I commented on, then ping me for another review :)

LICENSE.MIT
itsjeyd: @bradenmacdonald It's not necessary to include Backbone and Underscore when running problem-builder inside LMS/Studio, I just checked. However, tests fail horribly without them. Is there a way to conditionally include them, i.e., load them iff problem-builder is running as a standalone XBlock?

problem_builder/public/js/student_answers_dashboard.js
itsjeyd: Agreed (I ended up using `tr:nth-child(odd)` to make sure the table layout matches the [mockup](https://tasks.opencraft.com/secure/attachment/10402/instructor_dashboard.png)).

problem_builder/public/js/student_answers_dashboard.js
itsjeyd: @bradenmacdonald That's an interesting question! :) I guess I tend think of the `events` hash as the public API for a `Backbone.View`. So if client code wants to effect a change in a view in most cases I would expect it to trigger an appropriate event (instead of calling one of its methods). That's why I added a leading `_` to these handlers.

But maybe that's just me -- if you disagree I'll remove the underscores :)

itsjeyd:
@bradenmacdonald Thanks a lot for your comments! The code is ready for another review. For the inline notes, if I didn't reply directly that means I just applied them.

> Is "Student Answers Dashboard" the best name for this? I think it's confusing because there's another "Dashboard" block in this repo already, which also displays a student's answers.

I agree that it's a tad confusing, but renaming from "Data Export" to "Student Answers Dashboard" was an explicit requirement for completing OC-751. It probably came directly from the client? In any case, I'm pretty sure it's not my call to change it :)

Maybe @antoviaque can comment on this?

> FYI, I clicked the button to make Travis run the build again and now it's passing. [...] Usually you can just run the test again and it will pass.

Did not notice there was a button that lets you restart a build without pushing. Thanks for prompting me to go hunt for that, I know where to look for it now ;)

problem_builder/tasks.py
```diff
@@ -113,4 +88,108 @@ def scan_for_blocks(block):
...
+    try:
+        choices = block.children
+    except AttributeError:
+        pass
+    else:
+        for choice in choices:
+            choice_block = modulestore().get_item(choice)
+            if choice_block.value == answer:
+                answer = choice_block.content
+                break
```
bradenmacdonald: @itsjeyd Nice improvements! This is great. I only have one final comment, which is about fetching the choices above. I believe the choices will always be the same for each block, so we should avoid loading all the choices from the modulestore for each student.

You could:
1. Move the code that looks for choice children up one level to be in `_extract_data` outside of the per-student loop.
or 
2. Keep it here but cache the result (this is probably more efficient since it avoids looking up answers that won't be needed).

Actually, after looking into this I found out that you might be able to take advantage of `QuestionnaireAbstractBlock.get_submission_display` and to make this cleaner:
```python
def _get_answer(block, submission, answer_cache):  # Answer cache is a dict that is reset for each block.
    answer = submission['answer']
    if isinstance(block, QuestionnaireAbstractBlock):
        # Convert from answer ID to answer label
        if answer not in answer_cache:
            answer_cache[answer] = block.get_submission_display(answer)
        return answer_cache[answer]
    return answer
```

antoviaque:
@itsjeyd @bradenmacdonald It's true that we already have a Dashboard XBlock. Any suggestions on a better name? This now does more than just exporting data, so it would be good to keep reflecting that. There is no precise requirement from the client on this, we can chose what we think is best.

bradenmacdonald:
@antoviaque @itsjeyd Maybe rename the other dashboard to "Answer Review" or something? Or call this "Instructor Tool" ?

antoviaque:
@bradenmacdonald Both approaches would work for me - @itsjeyd up to you : )

itsjeyd:
@antoviaque OK then! :)

My vote is for renaming "Student Answers Dashboard" to "Instructor Tool", as this emphasizes the fact that unlike the other dashboard, the *Instructor Tool* is only for instructors/not accessible to students.

problem_builder/tasks.py
itsjeyd: @bradenmacdonald Thanks! dae27b3 implements the optimizations you suggest.

itsjeyd:
@bradenmacdonald This is ready for you to have another look. Thanks!

bradenmacdonald:
Thanks @itsjeyd - looks great! :+1:

PR #46, opened 2015-07-15
Fix issue that can occur when upgrading split mongo courses

bradenmacdonald:
This fixes an error that was occurring when upgrading split mongo courses containing Mentoring v1 blocks:

```
Traceback (most recent call last):
  File "/usr/lib/python2.7/runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/edx/xblocks/xblock-problem-builder/problem_builder/v1/upgrade.py", line 196, in <module>
    upgrade_block(block)
  File "/edx/xblocks/xblock-problem-builder/problem_builder/v1/upgrade.py", line 88, in upgrade_block
    index = parent_children.index(old_usage_id)
ValueError: BlockUsageLocator(CourseLocator(u'HGSE', u'1x', u'15', None, None), u'mentoring', u'mentoring_first') is not in list
```

The problem was that properties like `block.location`, `block.scope_ids.usage_id`, `block.parent`, and `block.children` were returning UsageKeys that contained `version` and `branch` data. Since we modify the modulestore during the upgrade process (changing the version of the course), this was causing all kinds of issues and conflicts.

The simple fix is to always use the `MixedModulestore` class rather than using the course's modulestore directly, because `MixedModulestore` strips out the branch and version information for us, avoiding those conflicts.

### DIFF
--- a/problem_builder/v1/upgrade.py
+++ b/problem_builder/v1/upgrade.py
@@ -38,14 +38,13 @@ from .studio_xml_utils import studio_update_from_node
 from .xml_changes import convert_xml_v1_to_v2
 
 
-def upgrade_block(block):
+def upgrade_block(store, block):
     """
     Given a MentoringBlock "block" with old-style (v1) data in its "xml_content" field, parse
     the XML and re-create the block with new-style (v2) children and settings.
     """
     assert isinstance(block, (MentoringBlock, NewMentoringBlock))
     assert bool(block.xml_content)  # If it's a v1 block it will have xml_content
-    store = block.runtime.modulestore
     xml_content_str = block.xml_content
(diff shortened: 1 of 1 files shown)


Kelketek:
@bradenmacdonald This seems sensible to me. The ticket mentions setting up a sandbox for demonstration. Could you do so?

bradenmacdonald:
@Kelketek Ah, sorry, I forgot about that bit. I will do that tomorrow and ping you again.

Kelketek:
@bradenmacdonald Realized I forgot to give you this: :+1: . Shhh. You didn't see me.

PR #49, opened 2015-07-21
Visual improvements to Table block.

Kelketek:
@antoviaque Hopefully these should address your notes.

Note, instead of making an extra download button, I placed the link where you suggested. Adding a download button would have prompted a lot more work in terms of alignment of the pop-ups and notifications, and I felt that it would be better to just place it down below.

### DIFF
--- a/problem_builder/public/js/review_blocks.js
+++ b/problem_builder/public/js/review_blocks.js
@@ -97,7 +97,9 @@ function MentoringTableBlock(runtime, element, initData) {
         sharedRefresh(data);
         $element.find(".new-share-container").each(function(index, container) {
             if (index === 0) {
-                $(container).find('.add-share-username').val('');
+                var $container = $(container);
+                $container.find('.add-share-username').val('');
+                $container.find('.add-share-field').show();
                 return;
             }
             $(container).remove()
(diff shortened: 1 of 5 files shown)


Kelketek:
@bradenmacdonald And you should probably verify these as well, come to think of it.

problem_builder/public/js/review_blocks.js
```diff
@@ -174,7 +174,7 @@ function MentoringTableBlock(runtime, element, initData) {
         })
     });
 
-    function addShare() {
+    function addShare(event) {
```
bradenmacdonald: Did you intend to use this `event` parameter, or are you just adding it for consistency?

bradenmacdonald:
@Kelketek 

* I think maybe the "Share" button should be disabled when viewing another user's table.
* When I view another user's table and press "Download", it downloads my table instead. The download link should probably also be disabled.
* There are still too many "+" buttons visible:  
  screen shot 2015-07-21 at 12 16 43 pm
* I think it's slightly confusing to say "Enter usernames" when there is only one text field visible - users may try to enter multiple usernames into the field. Or maybe I'm being too pessimistic.
* Personally, I preferred this before when it had a bit of an explanation and didn't just say "Enter usernames:"

bradenmacdonald:
Looks good! :+1:  if you get tests passing. Failures look unrelated so maybe just run the builds again.

Kelketek:
@bradenmacdonald I've rerun the tests a few times. Each time the failure is somewhere else but not related to the current changes, usually appearing to be something with upstream libraries, like getting more than one block or a transaction error. I can't seem to restart it anymore. Since all the tests have passed individually, I'm going to go ahead and merge so @mtyaka can complete his end of things.

bradenmacdonald:
@Kelketek Can you please create a JIRA ticket then about the flaky tests?

Kelketek:
@bradenmacdonald A ticket has been added to the backlog.

PR #47, opened 2015-07-16
Instructor Tool: Drop-down for "Root block ID" section, pagination improvements (OC-783)

itsjeyd:
This PR improves UX for the "Instructor Tool" XBlock by introducing a dropdown menu for constraining the set of exported answers to a specific block in the course outline:

Instructor Tool with dropdown

It also enables client-server communication for pagination functionality introduced in #45. This means that instead of transferring all results to the client when the export process finishes and paginating them there, at most `PAGE_SIZE` items (currently: 15) are transferred at a time. This is not a user-facing change (except for the fact that navigating between result pages takes a little bit longer now that the client has to communicate with the server to obtain more data).

### Testing

From Studio:

1. Add `pb-instructor-tool` to advanced modules.
2. Add one or more MCQs/rating questions/long answer blocks to a unit.
3. Add an instructor tool block to the same or another unit.

From the LMS:

1. Provide answers to MCQ/rating questions/long answers blocks.
2. Switch to instructor tool and select a block from the "Section/Question" dropdown.
3. Click "Search".
4. Repeat for other blocks.
5. **Optional**: For each result set, use pagination controls to navigate between result pages.

PR #50, opened 2015-07-22
Add unit tests for instructor tool

itsjeyd:
This PR aims to extend unit test coverage for problem-builder by introducing unit tests for the Instructor Tool block.

It follows up on #47.

### DIFF
--- a/problem_builder/tests/unit/test_instructor_tool.py
+++ b/problem_builder/tests/unit/test_instructor_tool.py
@@ -1,0 +1,255 @@
+"""
+Unit tests for Instructor Tool block
+"""
+import ddt
+import unittest
+from mock import Mock, patch
+from xblock.field_data import DictFieldData
+from problem_builder import InstructorToolBlock
+
+
+@ddt.ddt
+class TestInstructorToolBlock(unittest.TestCase):
+    """
+    Test InstructorToolBlock with some mocked data.
+    """
+
+    def _get_block(self, block_info):
+        # Build spec
+        spec = ['scope_ids', 'runtime', 'has_children', 'children']
+        for attr in block_info.get('attrs', []):
+            spec.append(attr)
+        # Assemble block
+        block = Mock(spec=spec)
+        scope_ids_mock = Mock()
+        scope_ids_mock.usage_id = block_info.get('usage_id')
+        block.scope_ids = scope_ids_mock
+        block.runtime = self.runtime_mock
+        block.children = []
+        for attr, val in block_info.get('attrs', {}).items():
+            setattr(block, attr, val)
+        return block
+
+    def setUp(self):
+        self.runtime_mock = Mock()
+        self.runtime_mock.get_block = self._get_block
+        scope_ids_mock = Mock()
+        scope_ids_mock.usage_id = u'0'
+        self.block = InstructorToolBlock(
+            self.runtime_mock, field_data=DictFieldData({}), scope_ids=scope_ids_mock
+        )
+        self.block.children = [
+            # No attributes: Prefer usage_id
+            {'usage_id': u'1'},
+            # Single attribute: Prefer attribute that's present
+            {'usage_id': u'2', 'preferred_attr': 'question', 'attrs': {'question': 'question'}},
+            {'usage_id': u'3', 'preferred_attr': 'name', 'attrs': {'name': 'name'}},
+            {'usage_id': u'4', 'preferred_attr': 'display_name', 'attrs': {'display_name': 'display_name'}},
+            # Two attributes (question, name): Prefer question
+            {
+                'usage_id': u'5',
+                'preferred_attr':
+                'question', 'attrs': {'question': 'question', 'name': 'name'}
+            },
+            # Two attributes (question, display_name): Prefer question
+            {
+                'usage_id': u'6',
+                'preferred_attr': 'question',
+                'attrs': {'question': 'question', 'display_name': 'display_name'}
+            },
+            # Two attributes (name, display_name): Prefer name
+            {
+                'usage_id': u'7',
+                'preferred_attr': 'name',
+                'attrs': {'name': 'name', 'display_name': 'display_name'}
+            },
+            # All attributes: Prefer question
+            {
+                'usage_id': u'8',
+                'preferred_attr': 'question',
+                'attrs': {'question': 'question', 'name': 'name', 'display_name': 'display_name'}
+            },
+        ]
+
+    def test_build_course_tree_uses_preferred_attrs(self):
+        """
+        Check if `_build_course_tree` method uses preferred block
+        attributes for `id` and `name` of each block.
+
+        Each entry of the block tree returned by `_build_course_tree`
+        is a dictionary that must contain an `id` key and a `name`
+        key.
+
+        - `id` must be set to the ID (usage_id or block_id)
+          of the corresponding block.
+
+        - `name` must be set to the value of one of the following attributes
+           of the corresponding block:
+
+          - question
+          - name (question ID)
+          - display_name (question title)
+          - block ID
+
+          Note that the attributes are listed in order of preference;
+          i.e., if `block.question` has a meaningful value, that value
+          should be used for `name` (irrespective of what the values
+          of the other attributes might be).
+        """
+        block_tree = self.block._build_course_tree()
+
+        def check_block(usage_id, expected_name):
+            # - Does block_tree contain single entry whose `id` matches `usage_id` of block?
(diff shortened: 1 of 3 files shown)


itsjeyd:
@e-kolpakov Could you please have a look at what I have so far? With the unit test introduced by a909afa I'm verifying that individual questions represented correctly in the root block dropdown. The basic idea is to check whether the `Fragment` returned by `student_view` contains a line like this for each question that it's supposed to display:

    <option value="block_id" ...>... block_name ...</option>

... where `block_name` corresponds to the value of the block attribute that is preferred for display purposes. (For example, if the value of `block.question` is non-empty, we always want to display that instead of `block.name`, `block.display_name`, or `block_id`).

Maybe there's a simpler approach? I haven't used the `mock` library a lot, so I'm probably not using it to its full potential.

CC @antoviaque

e-kolpakov:
@itsjeyd I believe the test tests too much:

* `student_view` method builds correct render context
* template engine uses the context to render expected select options.

I can see two ways to improve (and potentially simplify) it:

1. Convert it to integration test, than use selenium stuff to check for option values/texts
2. Patch `loader.render_template` and make sure it get's the correct rendering context.

Since we have an idea to have less integration and more unit tests I would go for option 2. It lacks testing that template does what it supposed to do, but I don't know any way to test templates except integration tests (if you happen to know one - share the knowledge :))

itsjeyd:
@e-kolpakov Thanks for the feedback! I went for option 2. If you could have a quick look to see if the changes match what you had in mind, that would be great :)

itsjeyd:
@e-kolpakov I added six more unit tests and two integration tests addressing issues that came up for #47. Please review. :)

problem_builder/tests/unit/test_instructor_tool.py
```diff
@@ -0,0 +1,226 @@
...
+        num_blocks = len(self.block.children) + 1
+        self.assertEqual(len(block_tree), num_blocks)
+
+        # Check block_tree for root entry
+        # - Does block_tree contain single entry whose `id` matches usage_id of instructor block?
+        matching_blocks = [block for block in block_tree if block['id'] == self.block.scope_ids.usage_id]
+        self.assertEqual(len(matching_blocks), 1)
+        # # - Is `name` of that entry set to "All"?
+        matching_block = matching_blocks.pop()
+        self.assertEqual(matching_block['name'], 'All')
```
e-kolpakov: @itsjeyd nit: could use a helper method:

    def check_block(usage_id, expected_name):
        matching_blocks = [block for block in block_tree if block['id'] == self.block.scope_ids.usage_id]
        self.assertEqual(len(matching_blocks), 1)
        matching_block = matching_blocks[0]
        self.assertEqual(matching_block['name'], expected_name)

problem_builder/tests/unit/test_instructor_tool.py
```diff
@@ -0,0 +1,226 @@
...
+        """
+        Check if `student_view` method correctly marks MCQ, Rating, and
+        Answer blocks as eligible when building tree for render
+        context.
+
+        A block is eligible if its type is one of {'pb-mcq', 'pb-rating', 'pb-answer'}.
+        """
+        # Pretend that all blocks in self.block.children are eligible:
+        self.block.runtime.id_reader = Mock()
+        self.block.runtime.id_reader.get_block_type.return_value = random.choice(['pb-mcq', 'pb-rating', 'pb-answer'])
```
e-kolpakov: @itsjeyd this only tests for one of `('pb-mcq', 'pb-rating', 'pb-answer')` at a time. Better approach would be to use `ddt` and parameterized test:

    @ddt.data('pb-mcq', 'pb-rating', 'pb-answer')
    def test_student_view_block_tree_eligible(self, block_type):
        ...
        self.block.runtime.id_reader.get_block_type.return_value = block_type
        ...

problem_builder/tests/unit/test_instructor_tool.py
```diff
@@ -0,0 +1,226 @@
...
+
+        instructor_tool.loader = Mock()
+        instructor_tool.loader.render_template.return_value = u''
```
e-kolpakov: @itsjeyd looks like `instructor_tool.loader` is always patched - maybe it might make sense to set it up in `setUp`?

problem_builder/tests/unit/test_instructor_tool.py
```diff
@@ -0,0 +1,226 @@
...
+          - name (question ID)
+          - display_name (question title)
+          - block ID
+
+          Note that the attributes are listed in order of preference;
+          i.e., if `block.question` has a meaningful value, that value
+          should be used for `name` (irrespective of what the values
+          of the other attributes might be).
+        """
+        instructor_tool.loader = Mock()
```
e-kolpakov: @itsjeyd looks like you're monkey-patching `loader` in instructor_tool. I believe the correct way for this would be to use mock.patch as context manager:

    with mock.patch('instructor_tool.loader`) as patched_loader:
        ...

problem_builder/tests/unit/test_instructor_tool.py
```diff
@@ -0,0 +1,226 @@
...
+                preferred_attr = child.get('preferred_attr')
+                self.assertEqual(matching_block['name'], attrs[preferred_attr])
+
+    def test_student_view_block_tree_choice_blocks_excluded(self):
+        """
+        Check if `student_view` method excludes 'pb-choice' blocks when
+        building block tree for render context.
+        """
+        # Pretend that all blocks in self.block.children are of type 'pb-choice:
+        self.block.runtime.id_reader = Mock()
```
e-kolpakov: @itsjeyd nit: I would access runtime mock using `self.runtime_mock` to make clear that it's actually a mock.

problem_builder/tests/unit/test_instructor_tool.py
```diff
@@ -0,0 +1,226 @@
...
+        A block is eligible if its type is one of {'pb-mcq', 'pb-rating', 'pb-answer'}.
+        """
+        # Pretend that all blocks in self.block.children are eligible:
+        self.block.runtime.id_reader = Mock()
+        self.block.runtime.id_reader.get_block_type.return_value = random.choice(['pb-mcq', 'pb-rating', 'pb-answer'])
+
+        instructor_tool.loader = Mock()
+        instructor_tool.loader.render_template.return_value = u''
+
+        self.block.student_view(context={})
```
e-kolpakov: @itsjeyd I'm afraid I don't like the overall approach here - you're essentially testing `build_tree` function that is defined as local in `student_view`. As a result, you've got to patch all the other stuff and call `student_view`. It might make sense to make `build_tree` a method on a block than test it explicitly, without `studen_view` and intercepting template rendering call arguments.

If you do want to test that correct template arguments are passed, it would be simpler than to patch `build_tree` to return whatever you want, and than intercept rendering call.

problem_builder/tests/unit/test_instructor_tool.py
```diff
@@ -0,0 +1,226 @@
...
+
+    def test_student_view_block_tree_ineligible(self):
+        """
+        Check if `student_view` method correctly marks blocks that aren't
+        MCQ, Rating, or Answer blocks as ineligible when building tree
+        for render context.
+        """
+        # Pretend that none of the blocks in self.block.children are eligible:
+        self.block.runtime.id_reader = Mock()
+        self.block.runtime.id_reader.get_block_type.return_value = 'pb-dummy-type'
```
e-kolpakov: @itsjeyd nit: to be precise - it tests that it marks 'pb-dummy-type' block as ineligible. Might make sense to use parameterized test with a couple of more examples (2-3 would be enough I believe)

problem_builder/tests/unit/test_instructor_tool.py
```diff
@@ -0,0 +1,226 @@
...
+        Check if student_view correctly handles scenarios in which
+        old-style keys are not supported.
+        """
+        self.block.scope_ids.def_id = Mock()
+        self.block.scope_ids.def_id.block_type.side_effect = AttributeError()
+
+        try:
+            self.block.student_view(context={})
+        except AttributeError:
+            self.fail('student_view breaks if old-style keys are not supported.')
```
e-kolpakov: @itsjeyd I'm totally confused by this test. So it should fail if `def_id.block_type` is accessed, right? How is that connected to old-style keys? Some more verbose docstring would be nice to have

itsjeyd:
@e-kolpakov Thanks for the comments! I addressed them all. Please have another look at the code.

e-kolpakov:
@itsjeyd :+1:

PR #52, opened 2015-07-27
Fix two bugs (OC-839)

bradenmacdonald:
JIRA: OC-839

This fixes:

Bug 1
-------
The LMS sequential ("subsection") displays tooltips in its accordion unit selector according to a strange criterion: If there are zero or one child blocks in the "Unit" (vertical block) and there is no title set on that child, then the Unit's title is used. Otherwise, the recursively combined titles of all child blocks are shown.

This caused an apparent regression when upgrading from mentoring to problem builder, in a course with no titles set on any of the mentoring/pb blocks: Since Problem Builder now declares a title ("Mentoring Questions") by default, that title was being used instead of the unit's title.

To fix this, I have changed the `get_content_titles` method to only return a title to the LMS if a title is set explicitly.

Before:
screen shot

After:
screenshot

Bug 2
-------

When a text answer is provided in the same subsection as an answer recap (or another text answer with the same answer id), submitted answers aren't updated. This is because the LMS loads a whole sub-section at once - when navigating the different units, the initial answers values remain the ones from the initial loading of the subsection.

I have fixed this by refreshing the answers with JavaScript/AJAX every time the unit is activated. The answer values of an editable answer field are only updated if the user hasn't modified them, to avoid losing edits.

### DIFF
--- a/problem_builder/answer.py
+++ b/problem_builder/answer.py
@@ -85,6 +85,23 @@ class AnswerMixin(object):
         )
         return answer_data
 
+    @property
+    def student_input(self):
+        if self.name:
+            return self.get_model_object().student_input
+        return ''
+
+    @XBlock.json_handler
+    def answer_value(self, data, suffix=''):
+        """ Current value of the answer, for refresh by client """
+        return {'value': self.student_input}
+
+    @XBlock.json_handler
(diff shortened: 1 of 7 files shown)


Kelketek:
:+1:

PR #53, opened 2015-07-28
Add an extra selector to fix CSS rule ambiguity

bradenmacdonald:
JIRA: OC-839

On the Harvard GSE instance, Problem Builder's LMS theme was not rendering correctly. The issue turned out to be that some of the theme's CSS rules had the same specificity as the "unthemed" rules, so that which one would be used was ambiguous.

Screenshot showing [desired interpretation](https://cloud.githubusercontent.com/assets/945577/8925664/6188fa46-34be-11e5-91eb-b3ac7bd19f0f.png) and [valid but wrong interpretation](https://cloud.githubusercontent.com/assets/945577/8925665/6bf3de38-34be-11e5-998e-e3642589572b.png).

Fix was to add an extra CSS selector to each of the rules in question.

### DIFF
--- a/problem_builder/public/themes/lms.css
+++ b/problem_builder/public/themes/lms.css
@@ -31,7 +31,7 @@ div.course-wrapper section.course-content .themed-xblock.mentoring p:empty {
     margin-bottom: 1.41575em;
 }
 
-.themed-xblock.mentoring .choices-list {
+.themed-xblock.mentoring .questionnaire .choices-list {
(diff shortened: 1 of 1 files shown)


bradenmacdonald:
@mtyaka Since you're in the right timezone, you'll need this to deploy, and have you experience with this part of problem builder, can you please review this?

mtyaka:
:+1:

PR #55, opened 2015-07-30
Forward Port #54: Don't cause 500 error when get_block returns None

bradenmacdonald:
Forward port of #54, which is already reviewed for the `edx-release` branch.


bradenmacdonald:
Note: GitHub got confused; the only change merged here was 36cd10b93228caa9d8a9cc17c38fda782e48f101

PR #57, opened 2015-08-19
Instructor Tool: Support searching for answers from multiple users

itsjeyd:
This PR extends existing functionality of the instructor tool by allowing users to enter multiple usernames to search for answers from multiple users simultaneously.

The input field for entering usernames expects usernames to be separated by commas, but is not sensitive to whitespace (meaning that commas can be surrounded by arbitrary amounts of whitespace).

#### Testing

From Studio:

1. Add `pb-instructor-tool` to advanced modules.
2. Add one or more MCQs/rating questions/long answer blocks to a unit.
3. Add an instructor tool block to the same or another unit.

From the LMS:

1. Sign in as `staff` and provide answers to some of the MCQ/rating questions/long answers blocks.
2. Repeat previous step for `honor`, `audit`, etc.
3. Sign in as `staff` and switch to instructor tool. Enter usernames of one or more users and click "Search". Observe that results include answers provided by users specified in the input field.
4. Repeat previous step for other combinations of usernames.

### DIFF
--- a/problem_builder/instructor_tool.py
+++ b/problem_builder/instructor_tool.py
@@ -282,7 +282,7 @@ class InstructorToolBlock(XBlock):
     def start_export(self, data, suffix=''):
         """ Start a new asynchronous export """
         block_types = data.get('block_types', None)
-        username = data.get('username', None)
+        usernames = data.get('usernames', None)
         root_block_id = data.get('root_block_id', None)
         match_string = data.get('match_string', None)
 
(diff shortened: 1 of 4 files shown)


problem_builder/instructor_tool.py
```diff
@@ -295,12 +296,17 @@ def start_export(self, data, suffix=''):
...
-        if not username:
-            user_id = None
+        if not usernames:
+            user_ids = None
         else:
-            user_id = user_service.get_anonymous_user_id(username, unicode(self.runtime.course_id))
-            if user_id is None:
-                self.raise_error(404, _("Could not find the specified username."))
+            user_ids = []
+            for username in re.split(r',', usernames):
```
bradenmacdonald: I don't think you need `re.split` here - can't you just use `usernames.split(',')` ?

problem_builder/templates/html/instructor_tool.html
```diff
@@ -9,8 +9,8 @@
     <div class="data-export-field-container">
       <div class="data-export-field">
         <label>
-          <span>{% trans "Username:" %}</span>
-          <input type="text" name="username" />
+          <span>{% trans "Usernames:" %}</span>
```
bradenmacdonald: Nit: would be better to say "Username[s]:" in order to clarify that one username is fine.

problem_builder/templates/html/instructor_tool.html
```diff
@@ -9,8 +9,8 @@
...
+          <input type="text" name="usernames" />
```
bradenmacdonald: I think this should explain somehow that if the user is entering multiple usernames, they should be comma-separated. I thought of suggesting placeholder text but I found that it makes it look like the field is filled in, which is confusing. Maybe a tooltip?

screen shot 2015-08-20 at 4 19 47 pm

bradenmacdonald:
Left a few minor comments but otherwise this looks good :+1:

itsjeyd:
@bradenmacdonald Thanks! Comments addressed.

PR #58, opened 2015-09-01
Simplify get_message: Return message instead of template

itsjeyd:
`MentoringBlock.get_message` currently returns an unintuitive result which doesn't contain the message itself but a template of it. This PR modifies the method to return a message instead.


e-kolpakov:
:+1:

PR #59, opened 2015-09-02
Fix: Translate links in table column headers

itsjeyd:
In the LMS, links of the form `/jump_to_id/<unit identifier>` are currently not translated/expanded correctly when they are part of a table column header. For example,

    /jump_to_id/mentoring_reflecting_on_feedback

correctly expands to

    https://openedx.gse.harvard.edu/courses/course-v1:HGSE+IOCEX+2015_T1/jump_to_id/mentoring_reflecting_on_feedback

when added to an HTML block. But when added to the column header of an Answer Recap table the expanded URL ends up missing information about the parent course (leading to an invalid link):

    https://openedx.gse.harvard.edu/jump_to_id/mentoring_reflecting_on_feedback

This PR adds code that makes sure links to course units that are part of table column headers are expanded correctly inside the LMS.

**Testing**

1. Download a copy of the [`course-v1:HGSE+IOCEX+2015_T1` course](https://studio.openedx.gse.harvard.edu/course/course-v1:HGSE+IOCEX+2015_T1) from the Harvard instance and import it into your local devstack.
2. In the LMS, navigate to Courseware ⟶ Change Diary ⟶ Change Diary: Map and Fields and make sure you are viewing the second unit ("My Change Diary - Maps").
3. Verify that the links in the table column headers can be used to successfully navigate to the corresponding units of the course.

### DIFF
--- a/problem_builder/table.py
+++ b/problem_builder/table.py
@@ -113,7 +113,11 @@ class MentoringTableBlock(
         for child_id in self.children:
             child = self.runtime.get_block(child_id)
             # Child should be an instance of MentoringTableColumn
-            header_values.append(child.header)
+            header = child.header
+            # Make sure /jump_to_id/ URLs are expanded correctly
+            if getattr(self.runtime, 'replace_jump_to_id_urls', None):
+                header = self.runtime.replace_jump_to_id_urls(header)
+            header_values.append(header)
             child_frag = child.render('mentoring_view', context)
             content_values.append(child_frag.content)
         context['header_values'] = header_values if any(header_values) else None
--- a/problem_builder/tests/integration/test_table.py
+++ b/problem_builder/tests/integration/test_table.py
@@ -20,6 +20,8 @@
 
 # Imports ###########################################################
 
+from mock import patch
+from workbench.runtime import WorkbenchRuntime
 from .base_test import MentoringBaseTest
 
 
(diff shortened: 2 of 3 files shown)


Kelketek:
:+1: Works for me, code looks fine.

antoviaque:
Can we add a test for this?

itsjeyd:
@antoviaque The code conditionally uses functionality from the LMS runtime to translate `jump_to_id` URLs (as do other XBlocks). That functionality is not present in the workbench runtime, so the check that happens [here](https://github.com/open-craft/problem-builder/commit/b232a70dfb7bfb4f3a6e67915e89fff76b03e3b9#diff-fa54906560384d09d1e10a12d21d1f31R118) will fail when the code runs outside of the LMS. So at this stage I'm not sure how we could go about implementing a test for the fix.

antoviaque:
@itsjeyd Could you mock the `replace_jump_to_id_urls()` method for the test, if it's not present in the workbench? You don't need to duplicate strictly the functionality, but if it could ensure the URL gets through this method when available, that might ensure we don't forget it again in future refactorings.

itsjeyd:
@antoviaque Thanks for the suggestions, mocking `replace_jump_to_id_urls` worked. I extended the integration test for `MentoringTableBlock` to cover this fix.

@Kelketek Please have another look.

Kelketek:
:+1: if the tests pass. I've retriggered the failing build since it seems to be one of those random flaky ones.

PR #60, opened 2015-09-04
Making question checkmark display question-level feedback + tests

e-kolpakov:


problem_builder/tests/integration/test_assessment.py
```diff
@@ -197,11 +197,11 @@ def peek_at_multiple_response_question(
     ):
         question = self.expect_question_visible(number, mentoring)
         self.assert_persistent_elements_present(mentoring)
-        self._selenium_bug_workaround_scroll_to(mentoring, question)
```
e-kolpakov: This workaround includes clicking on things, and every click outside the feedback area closes it. But one of the aspects of the test below is to check if the feedback is shown when user opens the question for review. Hence, the workaround was removed. It worked in my dev environment, so if it works just fine on CI it might mean the workaround was not needed here at all.

Kelketek:
:+1:

PR #61, opened 2015-09-04
Assessment targeted review

bradenmacdonald:
This new feature allows content authors to set an optional "review message" on each MCQ/MRQ/Rating question.

The review messages are shown in assessment mode when the student is reviewing their grade before being allowed to try again. The review screen now includes a list of topics to review, based on which questions the student got wrong.

Screenshot:
screen shot 2015-09-04 at 4 28 44 pm

If the student got all the questions right or has no attempts remaining, the new review section is not shown.

Here's how it looks to course authors in Studio:
screen shot 2015-09-04 at 4 29 13 pm

### DIFF
--- a/problem_builder/step.py
+++ b/problem_builder/step.py
@@ -52,6 +52,12 @@ class StepParentMixin(object):
         """
         return [_normalize_id(child_id) for child_id in self.children if child_isinstance(self, child_id, StepMixin)]
 
+    def get_steps(self):
+        """ Get the step children of this block, cached if possible. """
+        if getattr(self, "_steps_cache", None) is None:
+            self._steps_cache = [self.runtime.get_block(child_id) for child_id in self.steps]
+        return self._steps_cache
+
 
 class StepMixin(object):
     """
--- a/problem_builder/step.py
+++ b/problem_builder/step.py
@@ -45,7 +45,7 @@ class StepParentMixin(object):
     An XBlock mixin for a parent block containing Step children
     """
 
-    @property
+    @lazy
     def steps(self):
         """
         Get the usage_ids of all of this XBlock's children that are "Steps"
--- a/problem_builder/mentoring.py
+++ b/problem_builder/mentoring.py
@@ -424,7 +424,7 @@ class MentoringBlock(XBlock, StepParentMixin, StudioEditableXBlockMixin, StudioC
         Get the message to display to a student following a submission in assessment mode.
         """
         if not self.max_attempts_reached:
-            return self.get_message_content('on-assessment-review')
+            return self.get_message_content('on-assessment-review', or_default=True)
         else:
             return None
 
@@ -712,12 +712,16 @@ class MentoringBlock(XBlock, StepParentMixin, StudioEditableXBlockMixin, StudioC
     def max_attempts_reached(self):
         return self.max_attempts > 0 and self.num_attempts >= self.max_attempts
 
-    def get_message_content(self, message_type):
+    def get_message_content(self, message_type, or_default=False):
         for child_id in self.children:
             if child_isinstance(self, child_id, MentoringMessageBlock):
                 child = self.runtime.get_block(child_id)
                 if child.type == message_type:
                     return child.content
+        if or_default:
+            # Return the default value since no custom message is set.
+            # Note the WYSIWYG editor usually wraps the .content HTML in a <p> tag so we do the same here.
+            return '<p>{}</p>'.format(MentoringMessageBlock.MESSAGE_TYPES[message_type]['default'])
 
     def validate(self):
         """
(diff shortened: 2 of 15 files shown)


problem_builder/step.py
```diff
@@ -52,6 +52,12 @@ def steps(self):
         """
         return [_normalize_id(child_id) for child_id in self.children if child_isinstance(self, child_id, StepMixin)]
 
+    def get_steps(self):
```
e-kolpakov: @bradenmacdonald nit: could be lazy - lazies are per-instance, so even if you're in mixin each object of every class will get it's own lazy property.

e-kolpakov:
@bradenmacdonald one minor nit: the formatting of the list is different from the design doc. Here's what I see:

e-kolpakov:
@bradenmacdonald one more thing :) It looks like the code does not support more than a single assessment message, but one can add any number of them in Studio. It would be nice to allow single instance only, similar to how it's made for mentoring-level messages.

Also, latest xblock-utils provide a mixin for handling nested blocks in a more "declarative" way - you might want to check it out: [StudioContainerWithNestedXBlocksMixin](https://github.com/edx/xblock-utils/blob/master/xblockutils/studio_editable.py#L398). Examples are in GP v2 [utils-update](https://github.com/open-craft/xblock-group-project-v2/tree/utils-update) branch (virtually everywhere, but GroupProjectXBlock is probably the richest example so far). 

Simply put: 
* Parent block need to inherit from the StudioContainerWithNestedXBlocksMixin (which inherits from StudioContainerXBlockMixin), and override `allowed_nested_blocks` property to contain a list of blocks that we want to add to it via Studio as either:
** XBlock class - simple mode, just allows unlimited children of the type
** [NestedXBlockSpec](https://github.com/edx/xblock-utils/blob/master/xblockutils/studio_editable.py#L333) instance - provides more config options including only allowing single instance.
* Nested Blocks (i.e. in this case Answer, Tip and Message) need to have the follwoing constants set: 
** CATEGORY - XBlock category as set on add button (and setup.py entrypoint)
** STUDIO_LABEL - human-readable name for an XBlock

If any css/js needs to be added you can just override the `author_preview_view` and `author_edit_view`. Also, for hiding the studio header in preview - I've took another option to not use any view that studio appends the header to, so [XBlockWithPreviewMixin](https://github.com/edx/xblock-utils/blob/master/xblockutils/studio_editable.py#L381) was created.

problem_builder/step.py
bradenmacdonald: @e-kolpakov Using `lazy` would be nice, but it would force this to become an attribute rather than a method, and I'd like this to be a method so that `block.steps` and `block.get_steps()` follows the same API pattern as `block.children / block.get_children()` and `block.parent / block.get_parent()`

bradenmacdonald:
@e-kolpakov I think I've addressed all your comments.

I tried using `StudioContainerWithNestedXBlocksMixin` and it looks really nice - I just ran into two issues:
1. Does it include any CSS by default? The buttons were the wrong shape:
screen shot 2015-09-07 at 10 52 50 pm
2. I can't use it with `pb-message`, because the `MentoringMessageBlock` has many different names in Studio - the `STUDIO_LABEL` depends on the `boilerplate` value. If you add some way to override the `label` when defining a `NestedXBlockSpec` then it could work.

I'd love to refactor the whole problem builder to use `StudioContainerWithNestedXBlocksMixin` but I think that's outside the scope of this PR.

problem_builder/step.py
e-kolpakov: @bradenmacdonald makes sense

e-kolpakov:
@bradenmacdonald the rules we used for those buttons were [xblock-specific](https://github.com/open-craft/problem-builder/blob/master/problem_builder/public/css/problem-builder-edit.css#L7). Also, preview and edit views usually need studio view rules applied in order to look more or less like the actual student content. So I decided it would make sense to not include the css into `StudioContainerWithNestedXBlocksMixin` - instead, inherting class would just build the fragment via single super call and append required css, [like this](https://github.com/open-craft/xblock-group-project-v2/blob/utils-update/group_project_v2/mixins.py#L157-L179). 

On the other hand, it actually might make sense to include basic rules for buttons. If you send a PR against utils I'll be happy to review it, otherwise (if you decide to go with adding css in descendant class) I'll include that next time I get to work on it.

Same thing about NestedXBlock. I believe just adding a kwarg `label` to override block label would just do the trick.

also, it might be a good idea to log this feature requests as issues on utils repo - I'll go create them. :)

e-kolpakov:
@bradenmacdonald positioning issue is fixed, but now "If you retake..." message persists when user is retaking the assessment:

retake_message

Is that intentional?

bradenmacdonald:
Good catch, thanks. I will fix.

e-kolpakov:
@bradenmacdonald :+1:

PR #62, opened 2015-09-10
New version of mentoring block that supports explicit steps

itsjeyd:
This PR introduces two new blocks and enables support for viewing and editing them in Studio. One of the blocks is a new version of `MentoringBlock` that supports explicit steps, the other one is a block that represents a step:

-  The new mentoring block only allows step blocks (any number) and mentoring-wide messages as direct children.
-   The step block hosts all blocks that can not be added to new mentoring block (MCQ, MRQ, Long Answer, etc.)

**Acceptance criteria**

Studio integration:

-   [X] Make it possible to add new mentoring block to unit
-   [X] Make it possible to **add** step blocks
-   [X] Make it possible to **edit** step blocks
-   [X] Make it possible to **remove** step blocks
-   [X] Make it possible to **add step contents**
-   [X] Previews at mentoring and step levels should function properly
    -   [X] inner blocks should not have a studio header (small glitch: at mentoring level, HTML block still has a header)
    -   [X] inner blocks should look more or less the same as they would in LMS
    -   [X] inner blocks should have no obvious styling glitches and pass other common-sense checks

**Testing**

1. Enable new mentoring block for a course in Studio by adding `pb-mentoring` to advanced modules.
2. Add new mentoring block to a (new or existing) unit by choosing "Mentoring Questions (with explicit steps)" from the list of "Advanced" blocks.
3. Click "View" to edit the components of the block. Observe that the options for blocks to add are restricted to step and message blocks.
4. Add a step block.
5. Click "View" to edit the components of the block. Observe that the options for blocks to add include all relevant blocks that can not be added to the top level of the new mentoring block.
6. Try adding and removing a few components to/from the step block.
7. On the previous (mentoring) level, click "Edit" to change e.g. the title of the step block.
8. Click the trash icon to remove the step block.

### DIFF
--- a/problem_builder/mentoring.py
+++ b/problem_builder/mentoring.py
@@ -24,6 +24,9 @@ import logging
 import json
 
 from collections import namedtuple
+from itertools import chain
+
+from lazy.lazy import lazy
 
 from xblock.core import XBlock
 from xblock.exceptions import NoSuchViewError, JsonHandlerError
@@ -31,12 +34,18 @@ from xblock.fields import Boolean, Scope, String, Integer, Float, List
 from xblock.fragment import Fragment
 from xblock.validation import ValidationMessage
 
-from .message import MentoringMessageBlock
-from .step import StepParentMixin, StepMixin
+from .message import (
+    MentoringMessageBlock, CompletedMentoringMessageShim, IncompleteMentoringMessageShim,
+    MaxAttemptsReachedMentoringMessageShim, OnAssessmentReviewMentoringMessageShim
+)
+from .mixins import _normalize_id, StepParentMixin, QuestionMixin, XBlockWithTranslationServiceMixin
 
 from xblockutils.helpers import child_isinstance
 from xblockutils.resources import ResourceLoader
-from xblockutils.studio_editable import StudioEditableXBlockMixin, StudioContainerXBlockMixin
+from xblockutils.studio_editable import (
+    NestedXBlockSpec, StudioEditableXBlockMixin, StudioContainerXBlockMixin, StudioContainerWithNestedXBlocksMixin
+)
+
 
 try:
     # Used to detect if we're in the workbench so we can add Font Awesome
@@ -70,7 +79,89 @@ PARTIAL = 'partial'
 
 @XBlock.needs("i18n")
 @XBlock.wants('settings')
-class MentoringBlock(XBlock, StepParentMixin, StudioEditableXBlockMixin, StudioContainerXBlockMixin):
+class BaseMentoringBlock(
+        XBlock, XBlockWithTranslationServiceMixin, StudioEditableXBlockMixin
+):
+    """
+    An XBlock that defines functionality shared by mentoring blocks.
+    """
+    # Content
+    show_title = Boolean(
+        display_name=_("Show title"),
+        help=_("Display the title?"),
+        default=True,
+        scope=Scope.content
+    )
+
+    has_children = True
+
+    icon_class = 'problem'
+    block_settings_key = 'mentoring'
+    theme_key = 'theme'
+
+    @property
+    def url_name(self):
+        """
+        Get the url_name for this block. In Studio/LMS it is provided by a mixin, so we just
+        defer to super(). In the workbench or any other platform, we use the usage_id.
+        """
+        try:
+            return super(BaseMentoringBlock, self).url_name
+        except AttributeError:
+            return unicode(self.scope_ids.usage_id)
+
+    def get_theme(self):
+        """
+        Gets theme settings from settings service. Falls back to default (LMS) theme
+        if settings service is not available, xblock theme settings are not set or does
+        contain mentoring theme settings.
+        """
+        settings_service = self.runtime.service(self, "settings")
+        if settings_service:
+            xblock_settings = settings_service.get_settings_bucket(self)
+            if xblock_settings and self.theme_key in xblock_settings:
+                return xblock_settings[self.theme_key]
+        return _default_theme_config
+
+    def include_theme_files(self, fragment):
+        theme = self.get_theme()
+        theme_package, theme_files = theme['package'], theme['locations']
+        for theme_file in theme_files:
+            fragment.add_css(ResourceLoader(theme_package).load_unicode(theme_file))
+
+    @XBlock.json_handler
+    def view(self, data, suffix=''):
+        """
+        Current HTML view of the XBlock, for refresh by client
+        """
+        frag = self.student_view({})
+        return {'html': frag.content}
+
+    @XBlock.json_handler
+    def publish_event(self, data, suffix=''):
+        """
+        Publish data for analytics purposes
+        """
+        event_type = data.pop('event_type')
+        self.runtime.publish(self, event_type, data)
+
+        return {'result': 'ok'}
+
+    def author_preview_view(self, context):
+        """
+        Child blocks can override this to add a custom preview shown to
+        authors in Studio when not editing this block's children.
+        """
+        fragment = self.student_view(context)
+        fragment.add_content(loader.render_template('templates/html/mentoring_url_name.html', {
+            "url_name": self.url_name
+        }))
+        fragment.add_css_url(self.runtime.local_resource_url(self, 'public/css/problem-builder-edit.css'))
+        self.include_theme_files(fragment)
+        return fragment
+
+
+class MentoringBlock(BaseMentoringBlock, StudioContainerXBlockMixin, StepParentMixin):
     """
     An XBlock providing mentoring capabilities
 
@@ -121,12 +212,6 @@ class MentoringBlock(XBlock, StepParentMixin, StudioEditableXBlockMixin, StudioC
         default='',
         scope=Scope.content,
         multiline_editor=True
-    )
-    show_title = Boolean(
-        display_name=_("Show title"),
-        help=_("Display the title?"),
-        default=True,
-        scope=Scope.content
     )
 
     # Settings
@@ -196,34 +281,13 @@ class MentoringBlock(XBlock, StepParentMixin, StudioEditableXBlockMixin, StudioC
         'display_name', 'mode', 'followed_by', 'max_attempts', 'enforce_dependency',
         'display_submit', 'feedback_label', 'weight', 'extended_feedback'
     )
-    icon_class = 'problem'
+
     has_score = True
-    has_children = True
-
-    block_settings_key = 'mentoring'
-    theme_key = 'theme'
-
-    def _(self, text):
-        """ translate text """
-        return self.runtime.service(self, "i18n").ugettext(text)
 
     @property
     def is_assessment(self):
         """ Checks if mentoring XBlock is in assessment mode """
         return self.mode == 'assessment'
-
-    def get_theme(self):
-        """
-        Gets theme settings from settings service. Falls back to default (LMS) theme
-        if settings service is not available, xblock theme settings are not set or does
-        contain mentoring theme settings.
-        """
-        settings_service = self.runtime.service(self, "settings")
-        if settings_service:
-            xblock_settings = settings_service.get_settings_bucket(self)
-            if xblock_settings and self.theme_key in xblock_settings:
-                return xblock_settings[self.theme_key]
-        return _default_theme_config
 
     def get_question_number(self, question_id):
         """
@@ -231,7 +295,7 @@ class MentoringBlock(XBlock, StepParentMixin, StudioEditableXBlockMixin, StudioC
         """
         for child_id in self.children:
             question = self.runtime.get_block(child_id)
-            if isinstance(question, StepMixin) and (question.name == question_id):
+            if isinstance(question, QuestionMixin) and (question.name == question_id):
                 return question.step_number
         raise ValueError("Question ID in answer set not a step of this Mentoring Block!")
 
@@ -272,12 +336,6 @@ class MentoringBlock(XBlock, StepParentMixin, StudioEditableXBlockMixin, StudioC
 
         return Score(score, int(round(score * 100)), correct, incorrect, partially_correct)
 
-    def include_theme_files(self, fragment):
-        theme = self.get_theme()
-        theme_package, theme_files = theme['package'], theme['locations']
-        for theme_file in theme_files:
-            fragment.add_css(ResourceLoader(theme_package).load_unicode(theme_file))
-
     def student_view(self, context):
         # Migrate stored data if necessary
         self.migrate_fields()
@@ -296,7 +354,7 @@ class MentoringBlock(XBlock, StepParentMixin, StudioEditableXBlockMixin, StudioC
                 child_content += u"<p>[{}]</p>".format(self._(u"Error: Unable to load child component."))
             elif not isinstance(child, MentoringMessageBlock):
                 try:
-                    if self.is_assessment and isinstance(child, StepMixin):
+                    if self.is_assessment and isinstance(child, QuestionMixin):
                         child_fragment = child.render('assessment_step_view', context)
                     else:
                         child_fragment = child.render('mentoring_view', context)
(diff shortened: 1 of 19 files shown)


problem_builder/mentoring.py
```diff
@@ -776,3 +780,570 @@ def workbench_scenarios():
         Scenarios displayed by the workbench. Load them from external (private) repository
         """
         return loader.load_scenarios_from_path('templates/xml')
+
+
+@XBlock.needs("i18n")
+@XBlock.wants('settings')
+class MentoringWithExplicitStepsBlock(XBlock, StudioEditableXBlockMixin, StudioContainerXBlockMixin):
```
antoviaque: Can we make this a block specifically for handling steps, rather than a copy of the all-encompassing mentoring block? Ie maybe StepManager or StepBuilder?

I haven't looked closely, but it seem there is currently a lot of duplicated code with mentoring, currently? Imho we should only duplicate code we expect to be able to remove from Mentoring once we remove the assessments from there, to handle it exclusively with steps. Can the rest be factorized? Otherwise maintaining twice the "almost the same code" is not going to be fun.

itsjeyd: @antoviaque Thanks for the feedback! I DRYed out the changeset based on your suggestions (removed most of the duplicate code and factored the remaining code that was shared between the two mentoring blocks out into a shared superclass).

problem_builder/mentoring.py
```diff
@@ -776,3 +789,98 @@ def workbench_scenarios():
...
+
+        fragment.initialize_js('MentoringBlock')
+
+        return fragment
+
+    def author_edit_view(self, context):
+        """
+        Add some HTML to the author view that allows authors to add child blocks.
+        """
+        fragment = Fragment(u'<div class="mentoring">')  # This DIV is needed for CSS to apply to the previews
```
e-kolpakov: @itsjeyd please use StudioContainerWithNestedXBlocksMixin - it would look something like that:

    context['wrap_children'] = {
         'head': u'<div class="mentoring">',
         'tail': u'</div>'
    }
    fragment = super(MentoringWithExplicitStepsBlock, self).author_edit_view(context)
    fragment.add_content(loader.render_template('templates/html/mentoring_url_name.html', {
            "url_name": self.url_name
     }))
     # css and js stuff

`mentoring_edit.js` is mostly not needed, except for `ProblemBuilderUtil.transformClarifications(element);` Not sure what does it do - @bradenmacdonald - some hint on that? Anyway, if you still need it, it would be better to have a new `mentoring_with_steps_edit.js` with the following contents:

    function MentoringWithStepsEdit(runtime, element) {
        ProblemBuilderUtil.transformClarifications(element);
        StudioEditableXBlockMixin(runtime, element);
    }

problem_builder/mentoring.py
```diff
@@ -776,3 +789,98 @@ def workbench_scenarios():
...
+    def student_view(self, context):
+        fragment = Fragment()
+        child_content = u""
+
+        for child_id in self.children:
+            child = self.runtime.get_block(child_id)
+            if child is None:  # child should not be None but it can happen due to bugs or permission issues
+                child_content += u"<p>[{}]</p>".format(self._(u"Error: Unable to load child component."))
+            elif not isinstance(child, MentoringMessageBlock):
+                try:
```
e-kolpakov: @itsjeyd this try-catch block can be replaced by [`StudioContainerWithNestedXBlocksMixin._render_child_fragment`](https://github.com/edx/xblock-utils/blob/8f4e7689367342f3637712fa871f870c66fed19b/xblockutils/studio_editable.py#L476):

    child_fragment = self._render_child_fragment(child, context, view='mentoring_view')

problem_builder/public/css/answer.css
```diff
@@ -1,6 +1,6 @@
 .mentoring .answer.editable {
     height: 250px;
-    width: 100%;
+    width: 99%;
```
e-kolpakov: @itsjeyd Doesn't seem right. If there are some issues with it in edit mode, add a css rule to mentoring_edit or some css file only applied in studio.

problem_builder/step.py
```diff
@@ -75,40 +83,109 @@ class StepMixin(object):
         scope=Scope.content
     )
 
+    editable_fields = ('display_name', 'show_title',)
+
+    def _(self, text):
```
e-kolpakov: @itsjeyd I would move that `_` into a common mixin - it is duplicated virtually everywhere

problem_builder/step.py
```diff
@@ -75,40 +83,109 @@ class StepMixin(object):
...
-        return len(self.get_parent().steps) == 1
+    def child_number(self):
+        return list(self.siblings).index(_normalize_id(self.scope_ids.usage_id)) + 1
+
+    @lazy
+    def lonely_child(self):
+        if _normalize_id(self.scope_ids.usage_id) not in self.siblings:
+            message = u"{child_caption}'s parent should contain {child_caption}".format(child_caption=self.CAPTION)
+            raise ValueError(message, self, self.siblings)
+        return len(self.siblings) == 1
```
e-kolpakov: @itsjeyd child_number and lonely_child are quite similar to what StepMixix provide - I remember extracting these into a common mixin. I believe it would make sense to extract it.

problem_builder/mixins.py
```diff
@@ -0,0 +1,98 @@
...
+
+    @property
+    def steps(self):
+        """
+        Get the usage_ids of all of this XBlock's children that are "Steps"
+        """
+        return [_normalize_id(child_id) for child_id in self.children if child_isinstance(self, child_id, StepMixin)]
+
+
+class StepMixin(object):
```
e-kolpakov: @itsjeyd I would advise renaming it into "QuestionMixin" to avoid name collisions and confusing "Get all children that are questions" using `StepMixin`.

problem_builder/step.py
```diff
@@ -75,40 +83,109 @@ class StepMixin(object):
...
+            AnswerRecapBlock, MentoringTableBlock,
+        ]
 
-    def author_preview_view(self, context):
-        context = context.copy() if context else {}
-        context['hide_header'] = True
-        return self.student_view(context)
+    @property
+    def steps(self):
+        """ Get the usage_ids of all of this XBlock's children that are "Questions" """
```
e-kolpakov: @itsjeyd they are questions, yes, but searched by `StepMixin` - a bit confusing. Supports suggestion to rename StepMixin to QuestionMixin

problem_builder/mentoring.py
```diff
@@ -776,3 +789,98 @@ def workbench_scenarios():
...
+        fragment.add_resource(loader.load_unicode('templates/html/mentoring_review_questions.html'), "text/html")
+
+        self.include_theme_files(fragment)
+        # Workbench doesn't have font awesome, so add it:
+        if WorkbenchRuntime and isinstance(self.runtime, WorkbenchRuntime):
+            fragment.add_css_url('//maxcdn.bootstrapcdn.com/font-awesome/4.3.0/css/font-awesome.min.css')
+
+        fragment.initialize_js('MentoringBlock')
+
+        return fragment
```
e-kolpakov: @itsjeyd this student view is overly complicated - it is copy-pasted from MentoringXBlock, and we don't need that much stuff here. We have a dedicated task to make everything work in LMS, but if it is used in Studio as a preview it would make sense to get it somewhat working now. Alternative approach would be to provide a dedicated preview_view and make studio use it for preview.

problem_builder/templates/html/mentoring-with-steps-add-buttons.html
```diff
@@ -0,0 +1,14 @@
...
+
+<div class="add-xblock-component new-component-item adding">
+  <div class="new-component">
+    <h5>{% trans "Add New Component" %}</h5>
+    <ul class="new-component-type">
+      <li><a href="#" class="single-template add-xblock-component-button" data-category="pb-mentoring-step">{% trans "Mentoring Step" %}</a></li>
+      <li><a href="#" class="single-template add-xblock-component-button" data-category="pb-message" data-boilerplate="completed">{% trans "Message (Complete)" %}</a></li>
+      <li><a href="#" class="single-template add-xblock-component-button" data-category="pb-message" data-boilerplate="incomplete">{% trans "Message (Incomplete)" %}</a></li>
+      <li><a href="#" class="single-template add-xblock-component-button" data-category="pb-message" data-boilerplate="max_attempts_reached">{% trans "Message (Max # Attempts)" %}</a></li>
+      <li><a href="#" class="single-template add-xblock-component-button" data-category="pb-message" data-boilerplate="on-assessment-review">{% trans "Message (Assessment Review)" %}</a></li>
```
e-kolpakov: @itsjeyd not needed if MentoringWithSteps uses `StudioContainerWithNestedXBlocksMixin`

problem_builder/mentoring.py
antoviaque: @itsjeyd Thanks! Definitely better : )

problem_builder/mentoring.py
bradenmacdonald: `transformClarifications` is important - it parses the text/html content and injects the HTML needed for https://github.com/open-craft/problem-builder/pull/44

problem_builder/public/css/answer.css
itsjeyd: @e-kolpakov Answer blocks look fine in edit mode. The issue that this is supposed to fix manifests itself at unit level and at mentoring level:

answer-block-broken-display

The size of the `textarea` exceeds the size of the containing `div`. With the patch applied:

answer-block-display-fixed

In both cases, `AnswerBlock.mentoring_view` is called which adds `answer.css` to the fragment. That overrides other rules (e.g., those in `problem-builder-edit.css`). Not sure how else to address this, do you have another idea?

problem_builder/mentoring.py
```diff
@@ -776,3 +793,98 @@ def workbench_scenarios():
...
+        default=_("Mentoring Questions (with explicit steps)"),
+        scope=Scope.settings
+    )
+
+    editable_fields = ('display_name',)
+
+    @lazy
+    def questions(self):
+        """ Get the usage_ids of all of this XBlock's children that are "Questions" """
+        return list(chain.from_iterable(self.runtime.get_block(step_id).steps for step_id in self.steps))
```
e-kolpakov: @itsjeyd nit: this is rather expensive operation - it queries mongo for each individual block. I'm not 100% sure there is a caching mechanism in place (probably it is) or if any "collection" methods are exposed to Xblocks, but at least `self.runtime.get_block(step_id) for step_id in self.steps` could be "cached" in a lazy property.

problem_builder/mentoring.py
```diff
@@ -776,3 +793,98 @@ def workbench_scenarios():
...
+        NestedXBlockSpec allows explicitly setting disabled/enabled state, disabled reason (if any) and single/multiple
+        instances
+        """
+        from .step import MentoringStepBlock  # Import here to avoid circular dependency
+        return [
+            MentoringStepBlock,
+            NestedXBlockSpec(CompletedMentoringMessageBlock, boilerplate='completed'),
+            NestedXBlockSpec(IncompleteMentoringMessageBlock, boilerplate='incomplete'),
+            NestedXBlockSpec(MaxAttemptsReachedMentoringMessageBlock, boilerplate='max_attempts_reached'),
+            NestedXBlockSpec(OnAssessmentReviewMentoringMessageBlock, boilerplate='on-assessment-review'),
```
e-kolpakov: @itsjeyd if I do recall right, we decided that messages should be at step level. Also, if the reason you have created a dedicated class for each message is to support names, it could be done "better" in two ways:

1. Create shims instead of actual subclasses
2. Even better: add ability to override block name in NestedXblockSpec in xblock-utils. It is edx repo, but (at least so far), the master is considered an active-development branch, so we don't go into upstream review with changes - we only send batches of changes to edx-release (which is what edx uses) branch from time to time.

problem_builder/mentoring.py
```diff
@@ -776,3 +793,98 @@ def workbench_scenarios():
...
+        fragment = super(MentoringWithExplicitStepsBlock, self).author_edit_view(context)
+        fragment.add_content(loader.render_template('templates/html/mentoring_url_name.html', {
+            "url_name": self.url_name
+        }))
+        fragment.add_css_url(self.runtime.local_resource_url(self, 'public/css/problem-builder.css'))
+        fragment.add_css_url(self.runtime.local_resource_url(self, 'public/css/problem-builder-edit.css'))
+        fragment.add_css_url(self.runtime.local_resource_url(self, 'public/css/problem-builder-tinymce-content.css'))
+        fragment.add_javascript_url(self.runtime.local_resource_url(self, 'public/js/util.js'))
+        fragment.add_javascript_url(self.runtime.local_resource_url(self, 'public/js/mentoring_with_steps_edit.js'))
+        return fragment
```
e-kolpakov: @itsjeyd probably you've forgot to call `fragment.initialize_js('MentoringWithStepsEdit')`

problem_builder/mixins.py
```diff
@@ -0,0 +1,123 @@
...
+    @lazy
+    def siblings(self):
+        return self.get_parent().steps
+
+    @lazy
+    def step_number(self):
+        return list(self.siblings).index(_normalize_id(self.scope_ids.usage_id)) + 1
+
+    @lazy
+    def lonely_step(self):
```
e-kolpakov: @itsjeyd nit: should be `lonely_child` now

problem_builder/mixins.py
```diff
@@ -0,0 +1,123 @@
...
+    show_title = Boolean(
+        display_name=_("Show title"),
+        help=_("Display the title?"),
+        default=True,
+        scope=Scope.content
+    )
+
+    @lazy
+    def siblings(self):
+        return self.get_parent().steps
```
e-kolpakov: @itsjeyd mixin should not contain `siblings` - classes that use this mixin should provide their own implementations, as their siblings are different (i.e. steps should do this, while questions should return `self.get_parent().questions`, or something).

problem_builder/mixins.py
```diff
@@ -0,0 +1,123 @@
...
+
+    @lazy
+    def step_number(self):
+        return list(self.siblings).index(_normalize_id(self.scope_ids.usage_id)) + 1
+
+    @lazy
+    def lonely_step(self):
+        if _normalize_id(self.scope_ids.usage_id) not in self.siblings:
+            message = u"{step_caption}'s parent should contain {step_caption}".format(step_caption=self.CAPTION)
```
e-kolpakov: @itsjeyd nit: probably should rename format placeholder to `child_caption`

problem_builder/mixins.py
```diff
@@ -0,0 +1,123 @@
...
+            raise ValueError(message, self, self.siblings)
+        return len(self.siblings) == 1
+
+    @property
+    def display_name_with_default(self):
+        """ Get the title/display_name of this question. """
+        if self.display_name:
+            return self.display_name
+        if not self.lonely_step:
+            return self._(u"{step_caption} {number}").format(
```
e-kolpakov: @itsjeyd same as above here.

problem_builder/mixins.py
```diff
@@ -0,0 +1,123 @@
...
+        return self.student_view(context)
+
+    def assessment_step_view(self, context=None):
+        """
+        assessment_step_view is the same as mentoring_view, except its DIV will have a different
+        class (.xblock-v1-assessment_step_view) that we use for assessments to hide all the
+        steps with CSS and to detect which children of mentoring are "Steps" and which are just
+        decorative elements/instructions.
+        """
+        return self.mentoring_view(context)
```
e-kolpakov: @itsjeyd current implementation of `siblings` won't work for questions; they should provide their own (and as I said already, `EnumerableChildMixin` should not provide default implementation for `siblings`)

problem_builder/public/css/answer.css
e-kolpakov: @itsjeyd the question is, why hadn't it worked like that before the changes? I would suggest investigating it - probably there's some regression added in this PR (or at least it doesn't work like before). This rule looks like it is always applied, so it is totally not an option to add it - in best case we'll have regressions coming from QA about "this used to be 100% width, but not anymore" or (more likely) "there's an additional horizontal spacing next to freeform answers". In any case, the fix for this should only be applied to studio preview views.

Idea: Add common_preview_view.css that is only added to preview view (at all levels and blocks) and apply this change there (probably will need a stronger css selector)

problem_builder/public/css/problem-builder-edit.css
```diff
@@ -11,6 +11,8 @@
 }
 
 /* Custom appearance for our "Add" buttons */
+.xblock[data-block-type=pb-mentoring-step] .add-xblock-component .new-component .new-component-type .add-xblock-component-button,
```
e-kolpakov: @itsjeyd nit: I wish we could use sass/less/whatever else here

problem_builder/public/js/mentoring_with_steps_edit.js
```diff
@@ -0,0 +1,4 @@
+function MentoringWithStepsEdit(runtime, element) {
+    ProblemBuilderUtil.transformClarifications(element);
```
e-kolpakov: @itsjeyd nit: `'use strict'`

e-kolpakov:
@itsjeyd There're some glitches of various severity and complexity:

* "url_name for linking to this mentoring question set ..." used to be italicized, now it's not (now = in mentoring with steps)
* Adding Steps and messages does not work - you've forgot to add `CATEGORY` constants
* Might be a good idea to add some padding in preview view (though there aren't any in old mentoring, so if this takes more than 5 minutes it's ok to ignore it for now)
* Multiple messages of the same type are allowed - should only allow one; xblock-utils does not support single-instance-per-boilerplate yet, so use the approach in old mentoring - probably in mentoring_edit.js

itsjeyd:
@e-kolpakov As discussed in the hangout:

- reverted `answer.css`
- renamed `lonely_step` to `lonely_child`
- moved `sibling` implementations for to subclasses of `EnumerableChildMixin` and added a note about providing a default implementation in the future

itsjeyd:
@e-kolpakov FYI: I squashed the commits before rebasing but did preserve a copy of them just in case.

e-kolpakov:
@itsjeyd :+1:

PR #63, opened 2015-09-18
Step navigation and submission for new mentoring block

itsjeyd:
This PR adds step navigation and submission functionality to the new blocks introduced in #62.

* New mentoring block now displays one step at a time.
* New step block allows to submit step results. Info about last step submitted is saved at mentoring level so users are taken to the correct step when resuming their work (or reloading the page).
* New mentoring block allows to navigate to next step after submitting step results.
* After submitting the last step, new mentoring block displays a "Try again" button that allows to reset the block and start from scratch. (This functionality will be moved to the review step as part of OC-913.)

**Screenshots**

New mentoring block displaying a single step (with multiple questions):

new-mentoring-block-displaying-single-step

After submitting the current step (all answers correct):

controls-after-submit-all-correct

After submitting the current step (some answers correct):

controls-after-submit-some-correct

After submitting the current step (no correct answers):

controls-after-submit-none-correct

### DIFF
--- a/problem_builder/public/js/mentoring_with_steps.js
+++ b/problem_builder/public/js/mentoring_with_steps.js
@@ -3,24 +3,24 @@ function MentoringWithStepsBlock(runtime, element) {
     var steps = runtime.children(element).filter(
         function(c) { return c.element.className.indexOf('pb-mentoring-step') > -1; }
     );
-    var step = $('.mentoring', element).data('step');
-    var active_child, checkmark, submitDOM, nextDOM, tryAgainDOM, submitXHR;
+    var activeStep = $('.mentoring', element).data('active-step');
+    var checkmark, submitDOM, nextDOM, tryAgainDOM, submitXHR;
 
-    function isLastChild() {
-        return (active_child === steps.length-1);
+    function isLastStep() {
+        return (activeStep === steps.length-1);
     }
 
-    function updateStep() {
-        var handlerUrl = runtime.handlerUrl(element, 'update_step');
-        $.post(handlerUrl, JSON.stringify(step+1))
+    function updateActiveStep(newValue) {
+        var handlerUrl = runtime.handlerUrl(element, 'update_active_step');
+        $.post(handlerUrl, JSON.stringify(newValue))
             .success(function(response) {
-                step = response.step;
+                activeStep = response.active_step;
             });
     }
 
     function handleResults(response) {
-        // Update step so next step is shown on page reload (even if user does not click "Next Step")
-        updateStep();
+        // Update active step so next step is shown on page reload (even if user does not click "Next Step")
+        updateActiveStep(activeStep+1);
 
         // Update UI
         if (response.completed === 'correct') {
@@ -36,16 +36,16 @@ function MentoringWithStepsBlock(runtime, element) {
         nextDOM.removeAttr("disabled");
         if (nextDOM.is(':visible')) { nextDOM.focus(); }
 
-        if (isLastChild()) {
+        if (isLastStep()) {
             tryAgainDOM.removeAttr('disabled');
             tryAgainDOM.show();
         }
     }
 
     function submit() {
-        // We do not handle submissions at this level, so just forward to "submit" method of current step
-        var child = steps[active_child];
-        child['submit'](handleResults);
+        // We do not handle submissions at this level, so just forward to "submit" method of active step
+        var step = steps[activeStep];
+        step.submit(handleResults);
     }
 
     function hideAllSteps() {
@@ -61,23 +61,21 @@ function MentoringWithStepsBlock(runtime, element) {
         hideAllSteps();
     }
 
-    function displayNextChild() {
+    function updateDisplay() {
         cleanAll();
-        findNextChild();
+        showActiveStep();
         nextDOM.attr('disabled', 'disabled');
         validateXBlock();
     }
 
-    function findNextChild() {
-        // find the next real child block to display. HTMLBlock are always displayed
-        ++active_child;
-        var child = steps[active_child];
-        $(child.element).show();
+    function showActiveStep() {
+        var step = steps[activeStep];
+        $(step.element).show();
     }
 
     function onChange() {
         // We do not allow users to modify answers belonging to a step after submitting them:
-        // Once an answer has been submitted (next button is enabled),
+        // Once an answer has been submitted ("Next Step" button is enabled),
         // start ignoring changes to the answer.
         if (nextDOM.attr('disabled')) {
             validateXBlock();
@@ -85,17 +83,17 @@ function MentoringWithStepsBlock(runtime, element) {
     }
 
     function validateXBlock() {
-        var is_valid = true;
-        var child = steps[active_child];
-        if (child) {
-            is_valid = child['validate']();
+        var isValid = true;
+        var step = steps[activeStep];
+        if (step) {
+            isValid = step.validate();
         }
-        if (!is_valid) {
+        if (!isValid) {
             submitDOM.attr('disabled', 'disabled');
         } else {
             submitDOM.removeAttr('disabled');
         }
-        if (isLastChild()) {
+        if (isLastStep()) {
             nextDOM.hide();
         }
     }
@@ -103,19 +101,16 @@ function MentoringWithStepsBlock(runtime, element) {
     function initSteps(options) {
         for (var i=0; i < steps.length; i++) {
             var step = steps[i];
-            step['initChildren'](options);
+            step.initChildren(options);
         }
     }
 
     function handleTryAgain(result) {
-        if (result.result !== 'success')
-            return;
-
-        active_child = -1;
-        displayNextChild();
+        activeStep = result.active_step;
+        updateDisplay();
         tryAgainDOM.hide();
         submitDOM.show();
-        if (! isLastChild()) {
+        if (! isLastStep()) {
             nextDOM.show();
         }
     }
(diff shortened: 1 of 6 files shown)


problem_builder/step.py
```diff
@@ -81,6 +83,13 @@ class MentoringStepBlock(
         scope=Scope.content
     )
 
+    # User state
+    student_results = List(
```
e-kolpakov: @itsjeyd Is there any particular benefit of not using `Dict` field?

problem_builder/step.py
```diff
@@ -104,13 +113,42 @@ def allowed_nested_blocks(self):
...
+        for child in self.get_steps():
+            if child.name and child.name in submissions:
+                submission = submissions[child.name]
+                child_result = child.submit(submission)
+                submit_results.append([child.name, child_result])
+                child.save()
+
+        # Update results stored for this step
+        while self.student_results:
+            self.student_results.pop()
```
e-kolpakov: @itsjeyd I believe just assigning empty list should do just fine. Like 90% sure it should work.

problem_builder/step.py
```diff
@@ -104,13 +113,42 @@ def allowed_nested_blocks(self):
...
+        for result in submit_results:
+            self.student_results.append(result)
+
+        # Compute "answer status" for this step
+        if all(result[1]['status'] == 'correct' for result in submit_results):
+            completed = 'correct'
+        elif all(result[1]['status'] == 'incorrect' for result in submit_results):
+            completed = 'incorrect'
+        else:
+            completed = 'partial'
```
e-kolpakov: @itsjeyd nit: slightly more concise:

    correct_count = sum(1 for result in submit_results if result[1]['status'] == 'correct' )
    if correct_count == len(self.questions): completed = 'correct'
    elif correct_count > 0: completed = 'partial'
    else: completed = 'incorrect'

Still three-branch if, but at least it goes over results only once. Those are a dict, so performance benefits should be negligible. Feel free to ignore this

problem_builder/step.py
```diff
@@ -104,13 +113,42 @@ def allowed_nested_blocks(self):
...
+
+        # Update results stored for this step
+        while self.student_results:
+            self.student_results.pop()
+        for result in submit_results:
+            self.student_results.append(result)
+
+        # Compute "answer status" for this step
+        if all(result[1]['status'] == 'correct' for result in submit_results):
+            completed = 'correct'
```
e-kolpakov: @itsjeyd please turn this 'correct', 'incorrect' and 'partial' into constant, i.e. like:

    class Correctness(object):
        COMPLETE = 'complete'
        ....

problem_builder/templates/html/mentoring_with_steps.html
```diff
@@ -0,0 +1,20 @@
...
+<div class="mentoring themed-xblock" data-step="{{ self.step }}">
+
+  {% if show_title and title %}
+  <div class="title">
+    <h2>{{ title }}</h2>
+  </div>
+  {% endif %}
+
+  <div class="assessment-question-block">
+    {{child_content|safe}}
```
e-kolpakov: @itsjeyd nit: I believe 

    {{ for child_content in children_contents }}
        {{ children_content|safe}}
    {{ endfor }}

would be slightly more flexible: if needed, each child could be wrapped into a dom element, or otherwise post-processed in template

problem_builder/public/js/mentoring_with_steps.js
```diff
@@ -0,0 +1,157 @@
+function MentoringWithStepsBlock(runtime, element) {
+
+    var steps = runtime.children(element).filter(
+        function(c) { return c.element.className.indexOf('pb-mentoring-step') > -1; }
+    );
+    var step = $('.mentoring', element).data('step');
```
e-kolpakov: @itsjeyd nit: technically, it's not a step, it's step number. Consider renaming to `step_number`

problem_builder/public/js/mentoring_with_steps.js
```diff
@@ -0,0 +1,157 @@
...
+
+    function isLastChild() {
+        return (active_child === steps.length-1);
+    }
+
+    function updateStep() {
+        var handlerUrl = runtime.handlerUrl(element, 'update_step');
+        $.post(handlerUrl, JSON.stringify(step+1))
+            .success(function(response) {
+                step = response.step;
```
e-kolpakov: @itsjeyd nit: same applies here - step_number and update_step_number might be a better names for that.

problem_builder/public/js/mentoring_with_steps.js
```diff
@@ -0,0 +1,157 @@
...
+    var step = $('.mentoring', element).data('step');
+    var active_child, checkmark, submitDOM, nextDOM, tryAgainDOM, submitXHR;
+
+    function isLastChild() {
+        return (active_child === steps.length-1);
+    }
+
+    function updateStep() {
+        var handlerUrl = runtime.handlerUrl(element, 'update_step');
+        $.post(handlerUrl, JSON.stringify(step+1))
```
e-kolpakov: @itsjeyd nit: it actually does slightly more than the name tells - it also advances to the next step. Consider replacing `step+1` with parameter.

problem_builder/public/js/mentoring_with_steps.js
```diff
@@ -0,0 +1,157 @@
...
+        findNextChild();
+        nextDOM.attr('disabled', 'disabled');
+        validateXBlock();
+    }
+
+    function findNextChild() {
+        // find the next real child block to display. HTMLBlock are always displayed
+        ++active_child;
+        var child = steps[active_child];
+        $(child.element).show();
```
e-kolpakov: @itsjeyd nit: I would expect that `findNextChild` would return next child, not show it right away. Consider either renaming or returning the child and delegating showing it to the caller.

problem_builder/public/js/mentoring_with_steps.js
```diff
@@ -0,0 +1,157 @@
...
+        if (nextDOM.attr('disabled')) {
+            validateXBlock();
+        }
+    }
+
+    function validateXBlock() {
+        var is_valid = true;
+        var child = steps[active_child];
+        if (child) {
+            is_valid = child['validate']();
```
e-kolpakov: @itsjeyd could just use `child.validate()` - it is exactly the same thing.

problem_builder/public/js/mentoring_with_steps.js
```diff
@@ -0,0 +1,157 @@
...
+        }
+        if (isLastChild()) {
+            nextDOM.hide();
+        }
+    }
+
+    function initSteps(options) {
+        for (var i=0; i < steps.length; i++) {
+            var step = steps[i];
+            step['initChildren'](options);
```
e-kolpakov: @itsjeyd could just use `step.initChildren(options)`

problem_builder/public/js/mentoring_with_steps.js
```diff
@@ -0,0 +1,157 @@
...
+        }
+    }
+
+    function handleTryAgain(result) {
+        if (result.result !== 'success')
+            return;
+
+        active_child = -1;
```
e-kolpakov: @itsjeyd what's the difference between step and active_child? Shouldn't it be the same?

problem_builder/public/js/mentoring_with_steps.js
```diff
@@ -0,0 +1,157 @@
...
+            submitXHR.abort();
+        }
+        submitXHR = $.post(handlerUrl, JSON.stringify({})).success(handleTryAgain);
+    }
+
+    function initXBlockView() {
+        checkmark = $('.assessment-checkmark', element);
+
+        submitDOM = $(element).find('.submit .input-main');
+        submitDOM.bind('click', submit);
```
e-kolpakov: @itsjeyd I believe `bind` is somewhat [deprecated](http://api.jquery.com/bind/)

    As of jQuery 1.7, the .on() method is the preferred method for attaching event handlers to a document.

e-kolpakov:
@itsjeyd there's one minor issue in studio editing - not sure if it slipped from the previous task or was added as part of this one:
screenshot from 2015-09-18 17 09 13

"Que?" there is a "Question" field of an MCQ - looks like student view is reused for preview view.

problem_builder/step.py
itsjeyd: @e-kolpakov Hm. Maybe for the odd case where a question (with a non-unique `name`) is used multiple times within the same step? With a `List` field we'd be able to retain results for all occurrences of the question in that case.

itsjeyd:
@e-kolpakov Thanks for the comments. I looked into the studio editing issue -- it's present in the existing mentoring block as well and it seems like it's caused by a peculiarity of `MCQBlock`: Unlike other types of question blocks it adds the value of `self.question` to its `Fragment` in [`get_author_edit_view_fragment`](https://github.com/open-craft/problem-builder/blob/master/problem_builder/mcq.py#L121):

```python
def get_author_edit_view_fragment(self, context):
    """
    The options for the 1-5 values of the Likert scale aren't child blocks but we want to
    show them in the author edit view, for clarity.
    """
    fragment = Fragment(u"<p>{}</p>".format(self.question))
    self.render_children(context, fragment, can_reorder=True, can_add=False)
    return fragment
```
This behavior was introduced way back in Feb -- let me know if you think I should still change it as part of this PR.

e-kolpakov:
@itsjeyd if it's an existing behavior - let's just keep it. I have saw it before for soem reason.

problem_builder/step.py
e-kolpakov: @itsjeyd this is unlikely to work at all - how do you than distinguish the questions when you need to display their values? By sequential number? I'm not sure it would work (but not sure it wouldn't as well). Let's keep it as is for now, but please keep in mind it might make sense to use Dict.

e-kolpakov:
@itsjeyd :+1:

PR #64, opened 2015-09-18
Make sure messages get jump_to_id URLs rewritten.

Kelketek:
@bradenmacdonald @e-kolpakov @antoviaque 

To test this:
1. Create an MCQ with a message sub-block that has a /jump_to_id/<id_hash> url link in it.
2. Add in a message feedback block that also has one.
3. Set the mode to assessment.
4. Fail to answer the MCQ correctly, see the results page, check the links to see if they work.


bradenmacdonald:
:+1:

PR #66, opened 2015-09-30
Step Builder: Add mentoring-level messages to review step and disable mode switching for Problem Builder

itsjeyd:
This PR adds three types of customizable messages to review step blocks:

* **Complete** - shown if all answers correct (and some attempts remaining)
* **Incomplete** - shown if one or more answers are incorrect or partially correct (and some attempts remaining)
* **Review** - shown when user ran out of attempts

As a first step towards deprecating assessment functionality of Problem Builder, this PR removes the ability to change the mode of Problem Builder blocks to assessment mode.

Additionally, it includes the following fixes/additions:

* Sequence navigation tooltips do not display Step Builder descendants anymore.
* Step Builder sends a `"grade"` event when user completes an attempt (submits the last step).

**Testing**

1. If you haven't already, enable Step Builder by adding `"step-builder"` to advanced modules.
2. Add Step Builder to a new or existing unit and set max attempts to some non-zero value.
3. Add some Mentoring Step blocks and a Review Step block.
4. Add some questions to each Mentoring Step block.
5. Add message blocks to Review Step block, optionally customizing their content. Observe that each type of message can only be added once.
6. Complete the Step Builder block in the LMS. Observe that review step will display "Complete" message if all answers are correct, "Incomplete" message if some answers are incorrect/partially correct, "Review" message when no attempts remain.


problem_builder/public/js/review_step.js
```diff
@@ -3,8 +3,29 @@ function ReviewStepBlock(runtime, element) {
     var gradeTemplate = _.template($('#xblock-feedback-template').html());
     var reviewStepsTemplate = _.template($('#xblock-step-links-template').html());
 
+    var assessmentMessageDOM = $('.assessment-message', element);
```
e-kolpakov: @itsjeyd nit: it's not an assessment message anymore - consider renaming

problem_builder/mentoring.py
```diff
@@ -948,6 +923,10 @@ def score(self):
         return Score(score, int(round(score * 100)), correct, incorrect, partially_correct)
 
     @property
+    def complete(self):
+        return not self.score.incorrect and not self.score.partially_correct
```
e-kolpakov: @itsjeyd no tests for this

problem_builder/mentoring.py
```diff
@@ -1040,6 +1018,16 @@ def update_num_attempts(self, data, suffix=''):
         }
 
     @XBlock.json_handler
+    def publish_attempt(self, data, suffix):
+        score = self.score
```
e-kolpakov: @itsjeyd no tests as far as I can tell (hint: patch runtime.publish and verify it is called with expected values values when this method is called).

problem_builder/public/js/mentoring_with_steps.js
```diff
@@ -79,7 +79,15 @@ function MentoringWithStepsBlock(runtime, element) {
...
             .success(function(response) {
                 attemptsDOM.data('num_attempts', response.num_attempts);
-                // Now that relevant info is up-to-date, get the latest grade
+                publishAttempt();
+            });
+    }
+
+    function publishAttempt() {
+        var handlerUrl = runtime.handlerUrl(element, 'publish_attempt');
+        $.post(handlerUrl, JSON.stringify({}))
```
e-kolpakov: @itsjeyd do we have an integration test for this? There's no easy way (if *any* way exists) to test js parts of XBlocks so far, but it is possible to create an integration tests that would test sending grades when user answered all the questions

problem_builder/step.py
```diff
@@ -240,6 +243,45 @@ class ReviewStepBlock(XBlockWithPreviewMixin, XBlock):
...
+        """
+        return [
+            NestedXBlockSpec(CompletedMentoringMessageShim, boilerplate='completed'),
+            NestedXBlockSpec(IncompleteMentoringMessageShim, boilerplate='incomplete'),
+            NestedXBlockSpec(OnReviewMentoringMessageShim, boilerplate='on-review'),
+        ]
+
+    @XBlock.json_handler
+    def get_assessment_message(self, grade, suffix):
+        complete = grade['complete']
```
e-kolpakov: @itsjeyd not sure if it is implicitly tested by `test_default_messages` - if not, could you please add the test?

problem_builder/step.py
```diff
@@ -240,6 +243,45 @@ class ReviewStepBlock(XBlockWithPreviewMixin, XBlock):
...
+        max_attempts_reached = grade['max_attempts_reached']
```
e-kolpakov: @itsjeyd nit: fragile - if I do understand right `grade` is got from another ajax call, passed through a couple of JS functinos and than sent here. Modifications at each step might cause it to not contain `max_attemts_reached` or `complete`, so it will crash. Suggesting using `grade.get('max_attempts_reached')`.

Also, might make sense to add where this comes from (i.e. which AJAX handler produces the results passed here as `grade`)

problem_builder/step.py
itsjeyd: @e-kolpakov I double-checked, this is covered by both `test_default_messages` and `test_step_builder`.

e-kolpakov:
@itsjeyd :+1: conditional notes are addressed.

problem_builder/public/js/mentoring_with_steps.js
itsjeyd: @e-kolpakov Logic for publishing `"grade"` event is now covered by the `test_step_builder` integration test.

problem_builder/mentoring.py
itsjeyd: @e-kolpakov Since this is nit-level: There is an upcoming task for expanding test coverage for Step Builder. That will be a good opportunity to address this. I'll add a note to the description of that task to make sure we won't forget.

PR #69, opened 2015-10-09
Fixed appearance of step links in extended feedback

bradenmacdonald:
OC-986

This fixes two issues with the appearance of the question links in the "review" step in Step Builder:
* When many questions were listed, the line break broke the layout
* There was no comma or "and" between the links.

**Before**:
before

**After (Apros)**:
after

**After (LMS)**:
after-lms


problem_builder/templates/html/review_step.html
```diff
@@ -57,11 +57,16 @@
...
   <!-- Template for extended feedback: Show extended feedback details when all attempts are used up. -->
   <script type="text/template" id="xblock-step-links-template">
-  <% var q, last_question; %>
+  <% var q, last_question, second_last_question; %>
   <ul class="review-list <%= correctness %>-list">
-      <% for (var question in questions) {{ q = questions[question]; last_question = question == questions.length - 1; %>
-      <li><a href="#" class="step-link" data-step="<%= q.step %>"><%= _.template(gettext("Question {number}"), {number: q.number}, {interpolate: /\{(.+?)\}/g}) %></a></li>
-      <% }} %>
+      <% for (var question in questions) { %>
+        <% q = questions[question]; last_question = question == questions.length - 1; second_last_question = question == questions.length - 2; %>
```
itsjeyd: @bradenmacdonald The code in this line is quite dense; could you break the assignments up into multiple lines?

problem_builder/templates/html/review_step.html
```diff
@@ -57,11 +57,16 @@
...
+        <li>
+          <a href="#" class="step-link" data-step="<%= q.step %>"><%= 
+            _.template(gettext("Question {number}"), {number: q.number}, {interpolate: /\{(.+?)\}/g})
+          %></a><% if (!last_question) { %><%= (questions.length > 2 ? ", " : "") %><%= (second_last_question ? gettext(" and"): "") %><% } %>
```
itsjeyd: @bradenmacdonald `gettext(" and")` -- does the space before `and` need to be translated? Or would it make sense to change the whole check to:

```
<%= (second_last_question ? " " + gettext("and") : "") %>
```

Just asking because I don't know :)

itsjeyd:
@bradenmacdonald Looks good! I tested both LMS and Apros, and it works well. Just had two small comments/questions about the code, :+1: otherwise.

problem_builder/templates/html/review_step.html
bradenmacdonald: Good point. Changed.

problem_builder/templates/html/review_step.html
bradenmacdonald: Sure.

bradenmacdonald:
Thanks @itsjeyd !

PR #73, opened 2015-10-14
Allow Videos and images to be added to Mentoring Step Block.

Kelketek:
@mtyaka 

Testing instructions:

1. Create a course, add 'step-builder' to the advanced modules.
2. Add step builder block, add Mentoring Step to that block.
3. You should see Video block is available for adding. Add one to see if it works.
4. Stop CMS, install https://github.com/Stanford-Online/xblock-image-modal
5. Go back to Mentoring Step block in the CMS
6. See Image Modal in the available blocks list for adding. Add one to see if it works.

### DIFF
--- a/problem_builder/step.py
+++ b/problem_builder/step.py
@@ -127,11 +127,27 @@ class MentoringStepBlock(
         NestedXBlockSpec allows explicitly setting disabled/enabled state, disabled reason (if any) and single/multiple
         instances
         """
+        additional_blocks = []
+        try:
+            from xmodule.video_module.video_module import VideoDescriptor
+            additional_blocks.append(NestedXBlockSpec(
+                VideoDescriptor, category='video', label=_(u"Video")
+            ))
+        except ImportError:
+            pass
+        try:
+            from imagemodal import ImageModal
+            additional_blocks.append(NestedXBlockSpec(
+                ImageModal, category='imagemodal', label=_(u"Image Modal")
+            ))
+        except ImportError:
+            pass
+
         return [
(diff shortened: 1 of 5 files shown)


problem_builder/step.py
```diff
@@ -121,11 +121,27 @@ def allowed_nested_blocks(self):
...
+            ))
+        except ImportError:
+            pass
+
         return [
             NestedXBlockSpec(AnswerBlock, boilerplate='studio_default'),
             MCQBlock, RatingBlock, MRQBlock, HtmlBlockShim,
             AnswerRecapBlock, MentoringTableBlock,
-        ]
+        ] + additional_blocks
```
mtyaka: @Kelketek Now that this method does more than simply returns a static list, do you think it would make sense to add a unit test for it? I'm not sure how easy it is, but I'm guessing you could [fake the modules](http://stackoverflow.com/questions/5122465/can-i-fake-a-package-or-at-least-a-module-in-python-for-testing-purposes) that contain `VideoDescriptor` and `ImageModal` in tests.

mtyaka:
@Kelketek I'm wondering whether we could add a test for the `allowed_nested_blocks` method, but otherwise the changes look good and work well both in Studio and the LMS :+1:

Kelketek:
@mtyaka I was able to get tests in. Could you take a look at the test code?

mtyaka:
@Kelketek Great, thanks! :+1:

One of the travis checks failed, but the failure looks weird, I"m guessing it was just a temporary issue?

Kelketek:
@mtyaka Yeah. Most likely. I'll force a rerun and merge when it passes.

PR #74, opened 2015-10-14
Make 'next' button text customizable in Mentoring Steps.

Kelketek:
@itsjeyd Ready for your review.

You should be able to test this by:

1. Adding 'step-builder' to the advanced modules.
2. Creating a step-builder block.
3. Adding a Mentoring step block to the step-builder block
4. Adding a question and review to the Mentoring step block
5. Editing the mentoring step block to change the next button label text.
6. Publishing
7. Trying the step in the LMS to see if the next button has your customized text

Heads up @e-kolpakov , this should be coming down the pipe soon, and needs to be included in the release.

### DIFF
--- a/problem_builder/public/js/mentoring_with_steps.js
+++ b/problem_builder/public/js/mentoring_with_steps.js
@@ -175,8 +175,14 @@ function MentoringWithStepsBlock(runtime, element) {
         reviewTipsDOM.empty().hide();
     }
 
+    function updateNextLabel() {
+        var step = steps[activeStep];
+        nextDOM.attr('value', step.getStepLabel());
+    }
+
     function updateDisplay() {
         cleanAll();
+
         if (atReviewStep()) {
             // Tell supporting runtimes to enable navigation between units;
             // user is currently not in the middle of an attempt
@@ -188,6 +194,7 @@ function MentoringWithStepsBlock(runtime, element) {
         } else {
             showActiveStep();
             validateXBlock();
+            updateNextLabel();
             nextDOM.attr('disabled', 'disabled');
             if (isLastStep() && reviewStep) {
                 reviewDOM.attr('disabled', 'disabled');
@@ -243,6 +250,7 @@ function MentoringWithStepsBlock(runtime, element) {
         activeStep = stepIndex;
         cleanAll();
         showActiveStep();
+        updateNextLabel();
 
         if (isLastStep()) {
             reviewDOM.show();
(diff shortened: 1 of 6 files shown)


problem_builder/public/js/mentoring_with_steps.js
```diff
@@ -177,6 +177,8 @@ function MentoringWithStepsBlock(runtime, element) {
 
     function updateDisplay() {
         cleanAll();
+        nextDOM.attr('value', $('div.sb-step', element).data('next-button-label'));
```
itsjeyd: @Kelketek When testing your changes I found that if there are multiple Mentoring Step blocks, setting different "Next step" labels doesn't work the way it's supposed to: The label doesn't change from step to step; it's always set to the `next-button-label` of the first step on the page.

One way of addressing this would be to add a function to `MentoringStepBlock` (in step.js) that returns the label of the current step, something like:

```js
        getStepLabel: function() {
            return $('.sb-step', element).data('next-button-label');
        }
```

... and then update via:

```js
        var step = steps[activeStep];
        nextDOM.attr('value', step.getStepLabel());
```

Also, I think you'll need to update the label not just from `updateDisplay` but also from `jumpToReview` (this is called from `reviewNextStep`, which acts as a click handler for `nextDOM` while reviewing individual steps). To test this (I didn't), set `max_attempts` and `extended_feedback` on the Step Builder block and complete the block in the LMS until you have no attempts left, then use one of the links to jump back to a specific question/step.

problem_builder/public/js/mentoring_with_steps.js
```diff
@@ -175,8 +175,15 @@ function MentoringWithStepsBlock(runtime, element) {
         reviewTipsDOM.empty().hide();
     }
 
+    function updateNext() {
```
itsjeyd: @Kelketek Nit: I think it would be good to rename this to something a little less generic, like `updateNextLabel` or even `updateNextButtonLabel`.

problem_builder/tests/integration/xml_templates/step_builder_next.xml
```diff
@@ -0,0 +1,52 @@
...
+  <sb-step display_name="Second step" next_button_label="Next Item">
+    <pb-mcq name="mcq_1_1" question="Do you like this MCQ?" correct_choices='["yes"]'>
+      <pb-choice value="yes">Yes</pb-choice>
+        <pb-choice value="maybenot">Maybe not</pb-choice>
+        <pb-choice value="understand">I don't understand</pb-choice>
+
+        <pb-tip values='["yes"]'>Great!</pb-tip>
+        <pb-tip values='["maybenot"]'>Ah, damn.</pb-tip>
+        <pb-tip values='["understand"]'><div id="test-custom-html">Really?</div></pb-tip>
+        {% if include_review_tips %}
```
itsjeyd: @Kelketek Nit: AFAICT, `test_next_label` doesn't really check for the presence of review tips, so this block (and the one below could be probably be removed).

itsjeyd:
@Kelketek Looks good, :+1: once the tests pass.

PR #76, opened 2015-10-15
Enable seamless use of steps that have no questions.

Kelketek:
To test:

1. Add 'step-builder' to advanced modules.
2. Create a step builder block.
3. Create a few steps within the step builder block.
4. Fill some with nothing but fluff, others with questions
5. Make sure you can move through the fluff steps without disrupting functionality of the questions
6. Make sure the button semantics make sense-- next or review where appropriate

@e-kolpakov

### DIFF
--- a/problem_builder/public/js/mentoring_with_steps.js
+++ b/problem_builder/public/js/mentoring_with_steps.js
@@ -56,6 +56,15 @@ function MentoringWithStepsBlock(runtime, element) {
         }
     }
 
+    function postUpdateStep(response) {
+        activeStep = response.active_step;
+        if (activeStep === -1) {
+            updateNumAttempts();
+        } else {
+            updateControls();
+        }
+    }
+
     function handleResults(response) {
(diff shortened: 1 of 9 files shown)


problem_builder/public/js/mentoring_with_steps.js
```diff
@@ -195,9 +205,18 @@ function MentoringWithStepsBlock(runtime, element) {
             showActiveStep();
             validateXBlock();
             updateNextLabel();
-            nextDOM.attr('disabled', 'disabled');
+            var step = steps[activeStep];
+            if (step.hasQuestion()) {
```
e-kolpakov: @Kelketek these two new blocks are almost identical - would be good to refactor them into method.

Kelketek:
@e-kolpakov Addressed notes, fixed quality issue. Could you take a further look?

e-kolpakov:
@Kelketek :+1:

PR #77, opened 2015-10-16
Step Builder: New block type: "Plot"

itsjeyd:
This PR adds a "plot" block to Step Builder that can be used to visualize answers to pairs of rating questions.

**Testing**

3. Add two Rating questions to the first Mentoring Step block.
5. For each of the Rating questions, grab the corresponding "Question ID" from the EDIT dialog in Studio (last field, underneath "Label for high ratings").

7. Publish the unit.
8. Complete the first step of the unit by submitting answers to the rating questions.
9. Navigate to the second step. The plot should show the answers you gave in the form of a single point.
10. Repeat the previous two steps as another user and make sure to choose different answers for the rating questions. When you get to the second step, the point representing the average value of all submissions for the rating questions should be located in a different place than the point representing your own answer.


Kelketek:
@itsjeyd The method of inputting the questions and labels is a bit overcomplicated. Consider making the specification a child block. Is it only allowed to have two questions, or are an arbitrary amount? You may wish to use a list field on the child block in such a case.

This PR also needs a rebase, and as you mentioned, more tests.

Kelketek:
@itsjeyd The plot points appear to be the same for me when I'm the staff or honor user even when I choose different answers.

problem_builder/plot.py
```diff
@@ -0,0 +1,258 @@
...
+
+    @property
+    def default_claims(self):
+        return self._get_claims(self._get_default_response)
+
+    @property
+    def average_claims(self):
+        return self._get_claims(self._get_average_response)
+
+    def _get_claims(self, response_function):
```
Kelketek: You'd be able to eliminate this function or significantly reduce it by using children as mentioned in the earlier comment.

problem_builder/plot.py
```diff
@@ -0,0 +1,258 @@
...
+        return json.dumps(self.average_claims)
+
+    @XBlock.json_handler
+    def get_data(self, data, suffix):
+        return {
+            'default_claims': self.default_claims,
+            'average_claims': self.average_claims,
+        }
+
+    def author_preview_view(self, context):
```
Kelketek: It would be really nice if some sort of preview could be added, even with phony data somehow. Or just showing the empty grid with buttons. The answer data should be available on the studio side, I'd think-- so someone in the studio should be able to test with it directly if they really wanted to.

problem_builder/plot.py
```diff
@@ -0,0 +1,258 @@
...
+        return Fragment(
+            u"<p>{}</p>".format(
+                _(u"This block displays a plot that summarizes answers to scale questions.")
+            )
+        )
+
+    def mentoring_view(self, context):
+        return self.student_view(context)
+
+    def student_view(self, context=None):
```
Kelketek: Does just using this and not the authoring view break anything?

problem_builder/plot.py
```diff
@@ -0,0 +1,258 @@
...
+    if hasattr(key, "for_branch"):
+        key = key.for_branch(None)
+    if hasattr(key, "for_version"):
+        key = key.for_version(None)
+    return key
+
+
+@XBlock.needs('i18n')
+@XBlock.wants('user')
+class PlotBlock(StudioEditableXBlockMixin, StudioContainerWithNestedXBlocksMixin, XBlockWithPreviewMixin, XBlock):
```
Kelketek: Note you used the NestedXBlocksMixin, though there are no children XBlocks specified for this one. If you're going to take my suggestion about children for the plot items, great. Otherwise the user is given a prompt to add in children blocks, but is presented with no options to actually do so if they follow that link.

problem_builder/public/js/plot.js
```diff
@@ -0,0 +1,218 @@
...
+        .attr("transform", "translate(0, " + plotHeight / 2 + ")")
+        .call(xAxis);
+
+    var yAxisGroup = svgContainer.append("g")
+        .attr("transform", "translate(" + plotWidth / 2 + ", 0)")
+        .call(yAxis);
+
+    // Buttons
+
+    var defaultButton = $('.plot-default', element);
```
Kelketek: Nit: With this many var statements in sequence, you may want to use the comma syntax:

```
var defaultButton = $('.plot-default', element),
    averageButton = $('.plot-average', element),
    var quadrantsButton = $('.plot-quadrants', element);
```

But that can mean diffs by adding a new line are slightly bigger. Up to you.

problem_builder/public/js/step.js
```diff
@@ -1,6 +1,15 @@
 function MentoringStepBlock(runtime, element) {
 
     var children = runtime.children(element);
+    var plots = [];
+    for (var i = 0; i < children.length; i++) {
```
Kelketek: Perhaps use the python-esque JS for loop:

```
for (var i in children) {
    var child = children[i]
...
```

problem_builder/public/js/step.js
```diff
@@ -81,6 +90,15 @@ function MentoringStepBlock(runtime, element) {
                     callIfExists(child, 'handleReview', result);
                 }
             }
+        },
+
+        updatePlots: function() {
+            if (plots) {
+                for (var i = 0; i < plots.length; i++) {
```
Kelketek: Same note here about the for loop

problem_builder/step.py
```diff
@@ -216,10 +217,18 @@ def _render_view(self, context, view):
...
                 child_contents.append(u"<p>[{}]</p>".format(self._(u"Error: Unable to load child component.")))
             else:
-                child_fragment = self._render_child_fragment(child, context, view)
-
-                fragment.add_frag_resources(child_fragment)
-                child_contents.append(child_fragment.content)
+                if rendering_for_studio and isinstance(child, PlotBlock):
+                    # Don't use view to render plot blocks in Studio.
+                    # This is necessary because:
+                    # - student_view of plot block uses submissions API to retrieve results,
```
Kelketek: Oh, ok. I see why you don't have the authoring view showing the plots now. That makes sense. Nevermind, then.

problem_builder/tests/integration/test_step_builder.py
```diff
@@ -463,3 +463,84 @@ def test_default_messages(self):
...
+
+    def check_quadrant_labels(self, step_builder, controls, hidden):
+        quadrant_labels = step_builder.find_elements_by_css_selector(".quadrant-label")
+        quadrants_button_border_colors = [
+            controls.quadrants_button.value_of_css_property('border-top-color'),
+            controls.quadrants_button.value_of_css_property('border-right-color'),
+            controls.quadrants_button.value_of_css_property('border-bottom-color'),
+            controls.quadrants_button.value_of_css_property('border-left-color'),
+        ]
+        print(quadrants_button_border_colors)
```
Kelketek: Leftover print statement

problem_builder/plot.py
itsjeyd: @Kelketek I added the mixin because `PlotBlock` will get children as part of OC-1014, which will add support for defining additional overlay types.

itsjeyd:
@Kelketek Thanks for the feedback!

> The method of inputting the questions and labels is a bit overcomplicated. Consider making the specification a child block. Is it only allowed to have two questions, or are an arbitrary amount?

I discussed the method for specifying claims to include in plots with @antoviaque in the comments of OC-1010; for the current MVP, the way it works is sufficient.

> The plot points appear to be the same for me when I'm the staff or honor user even when I choose different answers.

I can't reproduce this on my end. Can you try hovering over the points to see the data? Do the numbers that you see correspond to the values that you chose for the different users?

Kelketek:
@itsjeyd https://github.com/open-craft/problem-builder/pull/73 contains a commit that fixes the tests. You may wish to cherry-pick the commit to run the tests, and remove it in the resulting rebase. The issue appears to stem from installing from the head of the SDK, which has some new changes that include extra logging.

I'll be looking at this shortly.

itsjeyd:
@Kelketek Thanks for the pointers, I cherry-picked b3ea5e.

problem_builder/plot.py
Kelketek: @itsjeyd As long as it's not shipping to production with this oddity, I'm ok with that.

Kelketek:
@itsjeyd You've got a failing test which should be easy to fix. Fix that, and it has my :+1:

itsjeyd:
@Kelketek Thanks for the re-review :)

PR #78, opened 2015-10-17
Step Builder: New question type: "Scale" (OC-1009)

bradenmacdonald:
This adds a new "Scale" block (actually I called it "Ranged Value Slider" but can change if needed):

**Screenshot**:
screen shot

### DIFF
--- a/problem_builder/mixins.py
+++ b/problem_builder/mixins.py
@@ -1,5 +1,5 @@
 from lazy import lazy
-from xblock.fields import String, Boolean, Scope
+from xblock.fields import String, Boolean, Float, Scope, UNIQUE_ID
 from xblockutils.helpers import child_isinstance
 from xblockutils.resources import ResourceLoader
 
@@ -125,11 +125,25 @@ class QuestionMixin(EnumerableChildMixin):
     has_author_view = True
 
     # Fields:
+    name = String(
+        # This doesn't need to be a field but is kept for backwards compatibility with v1 student data
+        display_name=_("Question ID (name)"),
+        help=_("The ID of this question (required). Should be unique within this mentoring component."),
+        default=UNIQUE_ID,
+        scope=Scope.settings,  # Must be scope.settings, or the unique ID will change every time this block is edited
+    )
     display_name = String(
         display_name=_("Question title"),
         help=_('Leave blank to use the default ("Question 1", "Question 2", etc.)'),
         default="",  # Blank will use 'Question x' - see display_name_with_default
         scope=Scope.content
+    )
+    weight = Float(
+        display_name=_("Weight"),
+        help=_("Defines the maximum total grade of this question."),
+        default=1,
+        scope=Scope.content,
+        enforce_type=True
     )
 
     @lazy
--- a/problem_builder/mixins.py
+++ b/problem_builder/mixins.py
@@ -126,7 +126,6 @@ class QuestionMixin(EnumerableChildMixin):
 
     # Fields:
     name = String(
-        # This doesn't need to be a field but is kept for backwards compatibility with v1 student data
         display_name=_("Question ID (name)"),
         help=_("The ID of this question (required). Should be unique within this mentoring component."),
         default=UNIQUE_ID,
--- a/problem_builder/questionnaire.py
+++ b/problem_builder/questionnaire.py
@@ -24,7 +24,7 @@ from django.utils.safestring import mark_safe
 from lazy import lazy
 import uuid
 from xblock.core import XBlock
-from xblock.fields import Scope, String, Float
+from xblock.fields import Scope, String
 from xblock.fragment import Fragment
 from xblock.validation import ValidationMessage
 from xblockutils.helpers import child_isinstance
--- a/problem_builder/questionnaire.py
+++ b/problem_builder/questionnaire.py
@@ -24,7 +24,7 @@ from django.utils.safestring import mark_safe
 from lazy import lazy
 import uuid
 from xblock.core import XBlock
-from xblock.fields import Scope, String, Float, UNIQUE_ID
+from xblock.fields import Scope, String, Float
 from xblock.fragment import Fragment
 from xblock.validation import ValidationMessage
 from xblockutils.helpers import child_isinstance
@@ -61,13 +61,6 @@ class QuestionnaireAbstractBlock(
     values entered by the student, and supports multiple types of multiple-choice
     set, with preset choices and author-defined values.
     """
-    name = String(
-        # This doesn't need to be a field but is kept for backwards compatibility with v1 student data
-        display_name=_("Question ID (name)"),
-        help=_("The ID of this question (required). Should be unique within this mentoring component."),
-        default=UNIQUE_ID,
-        scope=Scope.settings,  # Must be scope.settings, or the unique ID will change every time this block is edited
-    )
     question = String(
         display_name=_("Question"),
         help=_("Question to ask the student"),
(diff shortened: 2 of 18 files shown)


problem_builder/mixins.py
```diff
@@ -125,12 +125,26 @@ class QuestionMixin(EnumerableChildMixin):
     has_author_view = True
 
     # Fields:
+    name = String(
+        # This doesn't need to be a field but is kept for backwards compatibility with v1 student data
```
itsjeyd: @bradenmacdonald Is this comment still accurate? AFAICT both Rating questions and Scale blocks need this field (now that we also have Plot blocks).

problem_builder/slider.py
```diff
@@ -0,0 +1,162 @@
...
+        if sub_api:
+            # Also send to the submissions API:
+            sub_api.create_submission(self.student_item_key, {'value': value, })
+        result = self.get_last_result()
+        log.debug(u'Slider submission result: %s', result)
+        return result
+
+    def get_author_edit_view_fragment(self, context):
+        """
+        The options for the 1-5 values of the Likert scale aren't child blocks but we want to
```
itsjeyd: @bradenmacdonald Since we're not dealing with a Likert scale here this docstring is a bit misleading :)

problem_builder/slider.py
```diff
@@ -0,0 +1,157 @@
...
+            return {}
+        return {
+            'submission': self.student_value,
+            'status': 'correct',
+            'tips': [],
+            'weight': self.weight,
+            'score': 1,
+        }
+
+    def get_results(self, _previous_result_unused=None):
```
bradenmacdonald: @itsjeyd Do you know why we have `get_results` and `get_last_result`? The API is a little confusing. Would be nice to combined them.

problem_builder/questionnaire.py
```diff
@@ -24,7 +24,7 @@
 from lazy import lazy
 import uuid
 from xblock.core import XBlock
-from xblock.fields import Scope, String, Float, UNIQUE_ID
+from xblock.fields import Scope, String, Float
```
itsjeyd: @bradenmacdonald `Float` import is unused.

problem_builder/slider.py
```diff
@@ -0,0 +1,157 @@
...
+
+# Imports ###########################################################
+
+import logging
+import uuid
+
+from xblock.core import XBlock
+from xblock.fields import Scope, String, Float
+from xblock.fragment import Fragment
+from xblock.validation import ValidationMessage
```
itsjeyd: @bradenmacdonald This is a leftover import that doesn't seem to be used anywhere.

problem_builder/slider.py
```diff
@@ -0,0 +1,157 @@
...
+
+# Classes ###########################################################
+
+
+@XBlock.needs("i18n")
+class SliderBlock(
+    SubmittingXBlockMixin, QuestionMixin, StudioEditableXBlockMixin, XBlockWithTranslationServiceMixin, XBlock,
+):
+    """
+    An XBlock used to ask multiple-choice questions
```
itsjeyd: @bradenmacdonald This docstring needs to be adjusted a little.

problem_builder/slider.py
itsjeyd: @bradenmacdonald I'm not sure why we have `get_result` and `get_last_result`, that part of the API predates me :) But I agree, if possible it would be good to combine them.

problem_builder/slider.py
```diff
@@ -0,0 +1,157 @@
...
+        return self.get_last_result()
+
+    def submit(self, value):
+        log.debug(u'Received Slider submission: "%s"', value)
+        if value < 0 or value > 1:
+            return {}  # Invalid
+        self.student_value = value
+        if sub_api:
+            # Also send to the submissions API:
+            sub_api.create_submission(self.student_item_key, {'value': value, })
```
itsjeyd: @bradenmacdonald Nit: Since the dict that's passed to `create_submission` fits on one line I'd probably remove the trailing comma and space: `{'value': value, }` ⟶ `{'value': value}`

problem_builder/templates/html/slider.html
```diff
@@ -0,0 +1,17 @@
+<div class="xblock-pb-slider">
+  {% if not hide_header %}<h3 class="question-title">{{ title }}</h3>{% endif %}
+  {% if question %}
+    <p><label for="{{slider_id}}">{{ question|safe }} <span class="sr">({{instructions_string}})</span></label></p>
```
itsjeyd: @bradenmacdonald Yay for a11y! :D

problem_builder/tests/integration/test_slider.py
```diff
@@ -0,0 +1,147 @@
...
+# General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program in a file in the toplevel directory called
+# "AGPLv3".  If not, see <http://www.gnu.org/licenses/>.
+#
+
+# Imports ###########################################################
+
+from .base_test import ProblemBuilderBaseTest, MentoringAssessmentBaseTest, CORRECT, GetChoices
```
itsjeyd: @bradenmacdonald `CORRECT` seems to be unused.

problem_builder/tests/integration/test_slider.py
```diff
@@ -0,0 +1,147 @@
...
+        self.click_submit(pb_wrapper)
+        # Now, we expect submit to be disabled and the checkmark to be visible:
+        self.expect_checkmark_visible(True)
+        self.assertFalse(self.submit_button.is_enabled())
+        # Now change the value, and the button/checkmark should reset:
+        self.set_slider_value(45)
+        self.assertTrue(self.submit_button.is_enabled())
+        self.expect_checkmark_visible(False)
+        # Now reload the page:
+        self.browser.execute_script("$(document).html(' ');")
```
itsjeyd: @bradenmacdonald Why do we need this line? Just asking because I'd like to improve my understanding of integration tests :)

problem_builder/tests/integration/test_slider.py
```diff
@@ -0,0 +1,147 @@
...
+        self.wait_for_init()
+        # The initial value should be 50 and submit should be disabled until an MCQ choice is selected
+        self.assertEqual(self.get_slider_value(), 50)
+        self.assertFalse(self.submit_button.is_enabled())
+        self.expect_checkmark_visible(False)
+        # Set the value to 15:
+        self.set_slider_value(15)
+        self.assertEqual(self.get_slider_value(), 15)
+        self.assertFalse(self.submit_button.is_enabled())
+        # Choose a choice:
```
itsjeyd: @bradenmacdonald Would it be possible to use `GetChoices` for this?

problem_builder/tests/integration/xml_templates/slider_step.xml
```diff
@@ -0,0 +1,16 @@
+<step-builder display_name="Step Builder" max_attempts="90">
```
itsjeyd: @bradenmacdonald Nit: No need to set `max_attempts` here, I think (tests don't use it).

problem_builder/tests/integration/test_slider.py
bradenmacdonald: @itsjeyd Good question. Since Selenium has no "reload" API method, what happens when we reload the page is we tell the browser to navigate to the same URL, and then we wait and watch for specific elements to appear on the page, indicating that the page is fully loaded (`self.wait_for_init()`).

If we don't clear the DOM first, then it's possible that the `wait_for_init()` method will see those elements are already on the page (while the browser is still downloading the new page from the server), and proceed with the test, which will then fail.

bradenmacdonald:
@itsjeyd I pushed 9d40359e2c0bfd89a546a4ce5be8cb895caa1321 to hide the Studio headers when previewing sliders within a Step block... But I think this makes it less convenient to edit slider blocks (since you now always have to click "View" before you can edit them), and it is also inconsistent with HTML blocks (over which we have no control):

screen shot 2015-10-20 at 11 45 30 am

itsjeyd:
@bradenmacdonald :+1:

PR #81, opened 2015-10-22
Making MCQ use message attribute the same way MRQ does

e-kolpakov:
**Description:** MCQ and MRQ were inconsistent in terms of showing feedback: 
* MRQ provided both per-choice feedback and question-level feedback, 
* MCQ, depsite having "message" attribute, only provided per-choice feedback.

**Testing instructions:**
1. Create problem builder/step builder with MCQ and MRQ.
2. Fill in "Message" fields for both MCQ and MRQ
3. Provide choices and (optional) tips
4. Open PB/SB in LMS/Apros
5. Answer MRQ. Observe that when feedback is displayed, question-level feedback is shown first, per-choice feedback is available by clicking on correctness icons next to choices
6. Answer MCQ. Observe that the behavior is now the same (i.e. message from "Message" attribute is shown first, per-choice feedback is available by clicking on correctness icon)

If you can't see the feedback you might want to turn PB into normal mode (no longer supported in studio edit; use existing PB or update in mongo manually), or enable extended feedback and exhaust all the available attempts.

### DIFF
--- a/problem_builder/public/js/questionnaire.js
+++ b/problem_builder/public/js/questionnaire.js
@@ -1,5 +1,20 @@
 // TODO: Split in two files
 
+function display_message(message, messageView, checkmark){
+    if (message) {
+        var msg = '<div class="message-content">' + message + '</div>' +
+                  '<div class="close icon-remove-sign fa-times-circle"></div>';
+        messageView.showMessage(msg);
+        if (checkmark) {
+            checkmark.addClass('checkmark-clickable');
+            checkmark.on('click', function(ev) {
+                ev.stopPropagation();
+                messageView.showMessage(msg);
+            })
+        }
+    }
+}
+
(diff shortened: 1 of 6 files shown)


problem_builder/public/js/questionnaire.js
```diff
@@ -102,12 +117,15 @@ function MCQBlock(runtime, element) {
...
 
-        handleSubmit: function(result) {
+        handleSubmit: function(result, options) {
 
-            mentoring = this.mentoring;
+            var mentoring = this.mentoring;
 
             var messageView = MessageView(element, mentoring);
             messageView.clearResult();
+            debugger;
```
Kelketek: Left over debugger statement

Kelketek:
@e-kolpakov :+1: once you remove that leftover debug statement

PR #84, opened 2015-10-23
Add instructional message to step blocks.

Kelketek:
To test:

1. Create a step builder block.
2. Add a step to the block.
3. Edit the block. Fill in the message section
4. Add a problem to the block.
5. Save, answer the problem, submit, see new popup thingy. Click anywhere and it should dismiss.

### DIFF
--- a/problem_builder/public/js/step.js
+++ b/problem_builder/public/js/step.js
@@ -10,7 +10,8 @@ function MentoringStepBlock(runtime, element) {
         }
     }
 
-    var submitXHR, resultsXHR;
+    var submitXHR, resultsXHR,
+        message = $(element).find('.sb-step-message');
 
     function callIfExists(obj, fn) {
         if (typeof obj !== 'undefined' && typeof obj[fn] == 'function') {
@@ -59,6 +60,12 @@ function MentoringStepBlock(runtime, element) {
             submitXHR = $.post(handlerUrl, JSON.stringify(data))
                 .success(function(response) {
                     resultHandler(response);
+                    if (message.length) {
+                        message.fadeIn();
+                        $(document).click(function() {
+                            message.fadeOut();
+                        });
+                    }
                 });
         },
 
--- a/problem_builder/tests/integration/test_step_builder.py
+++ b/problem_builder/tests/integration/test_step_builder.py
@@ -81,7 +81,9 @@ class StepBuilderTest(MentoringAssessmentBaseTest):
         runtime_patcher.start()
         self.addCleanup(runtime_patcher.stop)
 
-    def freeform_answer(self, number, step_builder, controls, text_input, result, saved_value="", last=False):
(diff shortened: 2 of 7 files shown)


Kelketek:
@itsjeyd

itsjeyd:
@Kelketek I tested and the feature is working well overall (love the box-shadow :)). 

There are a couple of minor glitches:

1. The fact that the background of the message is transparent can make some areas of it a little harder to read; this mostly affects longer messages. Example screenshots:

  message-readability-1
  message-readability-2

2. If I submit and then click "Next step" without explicitly dismissing the message, the message is visible when I restart Step Builder via "Try again".

Code looks good, just had a couple of minor comments.

problem_builder/public/js/step.js
```diff
@@ -59,6 +60,12 @@ function MentoringStepBlock(runtime, element) {
             submitXHR = $.post(handlerUrl, JSON.stringify(data))
                 .success(function(response) {
                     resultHandler(response);
+                    if (message.length) {
+                        message.fadeIn();
+                        $(document).click(function() {
+                            message.fadeOut();
+                        })
```
itsjeyd: @Kelketek This is missing a semicolon.

problem_builder/tests/integration/test_step_builder.py
```diff
@@ -765,3 +774,14 @@ def test_plot(self):
...
+
+    def test_instruction_message(self):
+        step_builder, controls = self.load_assessment_scenario("step_builder_instruction.xml", {})
+        # Step 1
+        # Submit free-form answer, go to next step
+        self.freeform_answer(None, step_builder, controls, 'This is the answer', CORRECT, hold=True, last=True)
+        message = step_builder.find_element_by_css_selector('.sb-step-message')
+        self.wait_until_visible(message)
+        self.assertEqual(message.text, 'Hello!')
+        controls.review.click()
```
itsjeyd: @Kelketek It might make sense to extend the test scenario to include two steps: That would allow to not only check if the message disappears when clicking one of the controls for navigating between steps (as you are doing here), but also when clicking anywhere, e.g. via something like

```js
self.browser.execute_script("$(document).trigger('click')")
```

The reason I bring this up is because both "Next" and "Review grade" implicitly call `hideAllSteps` when clicked, which also has the effect of hiding the message (at least until you come back to the same step).

Kelketek: @itsjeyd I'm going to make it do just the document click, because it should work no matter where you click on the document, and clicking on a next button would be adding unneeded variables for the test.

Kelketek:
@itsjeyd Ready for another look.

itsjeyd:
@Kelketek :+1: once you fix the failing build (it's just a leftover blank line at the end of test_step_builder.py).

PR #85, opened 2015-10-23
Fix various bugs found in QA

bradenmacdonald:
This fixes the following bugs:
* **Bug 83** (MCKIN-3583): Step Builder grades do not show up in progress reports
   - This was due to the block missing the `has_score` and `max_score` attribute/method. I added them plus tests.
   - I also fixed a vulnerability that could allow clever students to use an AJAX request to submit an arbitrary grade.
   - Before:  
      screen shot 2015-10-22 at 9 44 39 pm
   - After:  
       screen shot 2015-10-22 at 9 45 02 pm
   - Verification instructions: Add a Step Builder to a course that contains at least one step and question. Go to the settings its the "Subsection" of the course and set the "Grade as" to a graded type. Login to Apros or the LMS as a student and submit an answer and click through until you see the review screen that says what grade you achieved. Now the progress page (in the LMS) or the proficiency/progress (Apros) should be updated.  

* **Bug 89** (MCKIN-3589): Pre-selected choices shown inconsistently.
   - To reproduce: As a student, work through a Step Builder that contains an MCQ and/or an MRQ. Once complete, click "Try again". Once you get to the MCQ/MRQ step, note that the previous answers are *not* visible. Refresh the page. The previous answers will now be filled in.
   - Fix: I added `hide_prev_answer: True` to the template context when in Step Builder, and changed the various question templates so they wouldn't show the previous answer when that variable is set.
 
* **Bug 88** (MCKIN-3588): Images don't load in MRQ Review Tips
   - To reproduce: Upload an image to your course's static files. Copy the "Studio URL" of the image which should be like `/static/...`. Create a Step Builder with an MCQ and an MRQ. Edit each MxQ block and set the "Message (General feedback provided when submiting)" property in the XBlock settings to an HTML string containing the Studio URL:
    screen shot 2015-10-23 at 12 22 28 am
    Set the Step Builder's `max_attempts` to `1` and enable extended feedback. As a student, complete the Step Builder. Then from the review screen, click on the link to review the each question.  
   For the MCQ, no feedback will be visible.
   For the MRQ, the feedback popup will be visible but the image won't load.
   This fix should fix both of the above issues.
   - Fix: The reason URLs weren't being rewritten is that the message HTML is not rendered (via `student_view` or similar), but instead passed directly to the template. As a result, the runtime doesn't get to wrap the output with methods that process the HTML. So I added some code that will call the runtime's `replace_urls` method manually for such messages.
   
* **Bug 82** (MCKIN-3582): When user attempts for a question, submit button remains enabled for sometime, and the "Next question" button can take a long time to appear
   - Fix: I refactored the code so it will disable the "Submit" button as soon as it's clicked, and so that it can submit the user's response with a single AJAX request rather than 4+ roundtrip requests per step. I also added more validation and error handling.
   
* I also fixed some bad XML that would cause parsing errors and test failures with newer versions of XBlock


problem_builder/mentoring.py
```diff
@@ -214,6 +225,10 @@ def author_preview_view(self, context):
         self.include_theme_files(fragment)
         return fragment
 
+    def max_score(self):
+        """ Maximum score. We scale all scores to a maximum of 1.0 so this is always 1.0 """
+        return 1.0
```
smarnach: How does this relate to the weight field above  Is there a difference between score and grade?

problem_builder/mentoring.py
```diff
@@ -1003,36 +1028,45 @@ def allowed_nested_blocks(self):
         ]
 
     @XBlock.json_handler
-    def update_active_step(self, new_value, suffix=''):
+    def submit(self, data, suffix=None):
+        """
+        Called this when the user has submitted the answer[s] for the current step.
```
smarnach: I think "this" is redundant.

problem_builder/mentoring.py
```diff
@@ -1003,36 +1028,45 @@ def allowed_nested_blocks(self):
...
-        }
+            # Update the number of attempts, if necessary:
+            if self.num_attempts < self.max_attempts:
+                self.num_attempts += 1
+            response_data['num_attempts'] = self.num_attempts
+            # And publish the score:
+            score = self.score
+            grade_data = {
+                'value': score.raw,
+                'max_value': 1,
```
smarnach: Shouldn't this be `self.max_value` (or maybe `self.weight`)?

smarnach:
:+1: Code looks good (I added a few questions, but that's probably just me not understanding the code).  Manual testing went fine.

During manual testing i noticed that Problem Builder works with neither of the two different (?) versions of xblock-utils configured in requirements/edx/github.txt and requirements/edx/custom.txt in the solutions fork of edx-platform.  Switching to the master branch of xblock-utils made it work fine.

problem_builder/mentoring.py
bradenmacdonald: Sure, I can set this to `self.max_score()` as a DRY thing.

problem_builder/mentoring.py
bradenmacdonald: Score vs grade: They are used interchangeably in the edX platform, but they are [trying to move to a more consistent approach](https://openedx.atlassian.net/wiki/display/TNL/edX+Grading+Guide+for+Developers) where "score" means a numeric result returned by an individual component such as an XBlock, and "grade" refers to the user's overall standing in a course (as a percentage and a letter grade or pass/fail). So the "grade" is a function of the user's scores plus the rules and weightings configured for the course.

If the user has not yet generated a score for this block:
* The `max_score` method, if defined, will result in '0/x' being displayed on the progress page for this block. As far as I know, the return value is not used directly, as the total possible score is scaled by the platform to the value of `weight`. So if `weight` is `1`, then the progress page will always display `0/1`.

If the user has generated a score for this block (which is done, confusingly, by publishing a `grade` XBlock event):
* Then the platform displays the user's score for this block as (`value / max_value * weight`), where `weight` is the field of the XBlock and `value`/`max_value` come from the event itself.

bradenmacdonald:
Thanks @smarnach. I addressed your comments and rebased this since it was conflicting with a couple recent PRs. Will merge once the build passes.

Regarding xblock-utils, that's strange. @e-kolpakov or @Kelketek do you have any insight and do you know why xblock-utils is listed twice?

PR #86, opened 2015-10-23
Step Builder: Interoperability for scale and plot blocks

itsjeyd:
This PR follows up on #77 and #78 which introduced plot and scale blocks, respectively. More specifically:

1. The changes make sure that plots can access data from scale blocks.
2. The changes make the process for obtaining references to rating and scale blocks consistent in Studio. This means that instead of having to

  - bring up the "EDIT" dialog
  - scroll down to the "Question ID" field
  - copy the contents

  course authors can now grab IDs from previews of rating questions:

  rating-block-with-url-name

  For scale blocks this behavior was already in place.

**Testing**

3. Add two Scale questions ("Ranged Value Slider") to the first Mentoring Step block.

5. For each Scale question, grab corresponding ID from preview.

7. Publish the unit.

8. Complete the first step of the unit by submitting answers to the Scale questions.

9. Navigate to the second step. The plot should show the answers you gave in the form of a single point.

### DIFF
--- a/problem_builder/mcq.py
+++ b/problem_builder/mcq.py
@@ -185,7 +185,7 @@ class RatingBlock(MCQBlock):
         list_values_provider=QuestionnaireAbstractBlock.choice_values_provider,
         list_style='set',  # Underered, unique items. Affects the UI editor.
     )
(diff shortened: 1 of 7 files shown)


Kelketek:
:+1:

PR #88, opened 2015-10-28
Step Builder: Fix JS error that causes crash if no review step is present

itsjeyd:


bradenmacdonald:
:+1: Works for me.

If you have time though, I think this could be cleaned up a bit more: `gradeDOM` could be owned by the review step JS rather than the step builder JS. Also, `gradeDOM` does not need to be a DOM element at all, since it is invisible. It's only used as a hack for passing grade data to the JS. A better way would be to pass the grade data as the second argument to `fragment.initialize_js('MentoringWithStepsBlock', data={'grade': self.get_grade()})` and then store the data in a JavaScript data structure owned by the review step JS code.

PR #89, opened 2015-11-01
Refactored Step Builder Messages

bradenmacdonald:
This refactors step builder to:
* Make the Review step code simpler
* Split the parts of the review screen up into XBlocks that are more flexible and re-orderable.
* Make the Step Builder block preview/work in Studio as students would see it in the LMS

**Example: Editing a review step**:
screen shot 2015-11-01 at 8 36 53 pm

**Example: A Conditional Message**:
screen shot 2015-11-01 at 8 06 24 pm


problem_builder/mentoring.py
```diff
@@ -1101,9 +1115,7 @@ def try_again(self, data, suffix=''):
         }
 
     def author_preview_view(self, context):
-        context = context.copy() if context else {}
-        context['author_preview_view'] = True
-        return super(MentoringWithExplicitStepsBlock, self).author_preview_view(context)
+        return self.student_view(context)
```
smarnach: Isn't it the implicit behaviour to fall back to the student view?  (No need to change anything, it's fine to keep this explicit in any case.  I'm just curious.)

problem_builder/public/js/mentoring_with_steps.js
```diff
@@ -8,23 +8,22 @@ function MentoringWithStepsBlock(runtime, element) {
...
 
     var activeStep = $('.mentoring', element).data('active-step');
-    var reviewTipsTemplate = _.template($('#xblock-review-tips-template').html()); // Tips about specific questions the user got wrong
     var attemptsTemplate = _.template($('#xblock-attempts-template').html());
     var message = $('.sb-step-message', element);
-    var checkmark, submitDOM, nextDOM, reviewDOM, tryAgainDOM,
-        gradeDOM, attemptsDOM, reviewTipsDOM, reviewLinkDOM, submitXHR;
+    var checkmark, submitDOM, nextDOM, reviewButtonDOM, tryAgainDOM,
+        gradeDOM, attemptsDOM, reviewLinkDOM, submitXHR;
+    var reviewStepDOM = $("[data-block-type=sb-review-step]", element);
```
smarnach: It's probably more efficient to restrict this a bit further, like `"div.xblock[data-block-type=sb-review-step]"`.  Otherwise jQuery will have to iterate through all DOM objects and retrieve the `data-block-type` attribute.  (Not sure whether what I suggested would be correct in all cases.  Also not sure whether performance matters here.)

problem_builder/public/js/mentoring_with_steps.js
```diff
@@ -111,7 +93,8 @@ function MentoringWithStepsBlock(runtime, element) {
                 // We are now showing the review step / end
                 // Update the number of attempts.
                 attemptsDOM.data('num_attempts', response.num_attempts);
-                updateGrade(response.grade_data);
+                reviewStepDOM.html($(response.review_html).html());
```
smarnach: Again a question for my personal education: How is this different from `reviewStepDOM.html(response.review_html)`?  Does it just remove the outermost HTML tag?  If so, why is it included in `review_html` in the first place?

problem_builder/step_review.py
```diff
@@ -0,0 +1,291 @@
...
+            (self.score_condition == self.SCORE_IMPERFECT and perfect_score)
+        ):
+            return False
+
+        return True
+
+    def student_view(self, context=None):
+        """ Render this message. """
+        html = u'<div class="review-conditional-message">{content}</div>'.format(
+            content=self.content
```
smarnach: Do we need HTML escaping here?

problem_builder/step_review.py
```diff
@@ -0,0 +1,291 @@
...
+    STUDIO_LABEL = _("Score Summary")
+    has_author_view = True  # Without this flag, studio will use student_view on newly-added blocks :/
+
+    @property
+    def display_name_with_default(self):
+        return self._(self.STUDIO_LABEL)
+
+    def student_view(self, context=None):
+        """ Render the score summary message. """
+        html = loader.render_template("templates/html/sb-review-score.html", context.get("score_summary", {}))
```
smarnach: This won't work when `context` has its default value `None`.

problem_builder/step_review.py
```diff
@@ -0,0 +1,291 @@
...
+        return Fragment(html)
+
+    mentoring_view = student_view  # Same as student_view but Studio won't wrap it with the editing header/buttons
+
+    def author_view(self, context=None):
+        if not context.get("score_summary"):
```
smarnach: Same here.  (Not sure whether that ever happens, but then I wonder why we have a default value for `context` anyway.)

problem_builder/step_review.py
```diff
@@ -0,0 +1,291 @@
...
+
+                'is_example': True,
+            }
+        return self.student_view(context)
+
+
+@XBlock.needs("i18n")
+class PerQuestionFeedbackBlock(XBlockWithTranslationServiceMixin, XBlockWithPreviewMixin, XBlock):
+    """
+    Summaryize the score that the student earned.
```
smarnach: Spelling: "Summaryize"

problem_builder/step_review.py
```diff
@@ -0,0 +1,291 @@
...
+            })
+        else:
+            html = u""
+        return Fragment(html)
+
+    mentoring_view = student_view  # Same as student_view but Studio won't wrap it with the editing header/buttons
+
+    def author_view(self, context=None):
+        """ Show example content in Studio """
+        if not context.get("per_question_review_tips"):
```
smarnach: Check for `context is None` missing again.

problem_builder/templates/html/sb-review-per-question-feedback.html
```diff
@@ -0,0 +1,9 @@
+{% load i18n %}
+
+<!-- Tips about specific questions the student got wrong. From pb-message[type=on-assessment-review-question] blocks -->
```
smarnach: This comment will be included in the resulting HTML.  You can use

    {# Your comment here. #}

in Django templates to avoid that.

smarnach:
@bradenmacdonald Nice, this makes the code indeed more readable!  I've done a first past through the code, but haven't done any testing yet.  To facilitate getting this merged until tomorrow morning, I'll do the  testing after the meeting, so if anything comes up you will still have a bit of time to address any issues.

How much time will you need for the tests?  If you can finish them soon, I might be able to give a thumbs up today.  (I'll be available until about 22:00 UTC.)

problem_builder/mentoring.py
bradenmacdonald: Yes, but if it renders `student_view` as a fallback, Studio will wrap the XBlock in a bunch of HTML (header, edit controls) that we don't want. Studio looks at the name of the view that's being rendered when deciding to wrap the element or not.

problem_builder/step_review.py
bradenmacdonald: No, problem builder blocks generally allow the authors to enter custom HTML.

bradenmacdonald:
@smarnach Thanks for the initial review! Yes I'm happy with how this has improved the code :-)

I'll work on addressing your comments and updating the tests right after the meeting, but I'm not sure how long it will take. A couple hours perhaps?

smarnach:
@bradenmacdonald OK, we'll see then.  Either I'm still available to take a final look, or it will have to wait until tomorrow.

problem_builder/public/js/mentoring_with_steps.js
bradenmacdonald: I agree about the "DOM" name in the variables, but that's an existing convention in this code.

Also yeah, the problem with the selector is that in Workbench the XBlock DIVs have the class `xblock-v1` but in Studio they have the class `xblock` :-/ I can just double the selector though.

problem_builder/public/js/mentoring_with_steps.js
bradenmacdonald: @smarnach Yes, it removes the outermost HTML tag. It's included because the XBlock runtime puts it there and we can't control that. (I could bypass the normal block rendering pipeline which wraps it in that DIV but then we would miss other important steps in the render pipeline, such as link rewriting).

This could in theory be simplified to `reviewStepDOM.replaceWith($(response.review_html))` but that would invalidate the jQuery object / DOM reference `reviewStepDOM`, so I found this to be cleaner.

bradenmacdonald:
@smarnach  I've addressed your initial review comments and fixed a small bug. Now I'm just continuing to get the tests passing again.

smarnach:
When adding a new Per-Question Feedback block and cliking "Edit" on it, I get this error:

screenshot_2015-11-02_22-07-52

(This is on a relatively current version of the solutions devstack.)

bradenmacdonald:
@smarnach Good catch. Fix just pushed: ee49a27365576954b27adf71c68935b084595733

(I also just pushed two other fixes and will "soon" push a commit that makes the tests pass, hopefully)

smarnach:
Does the new version have to be backwards compatible to existing course data?  When loading my pre-existing Step Builder example in Studio, one of the messages was displayed with the heading "INVALID MESSAGE".  I also still had the old "Message shown when complete/incomplete" blocks, and things didn't properly work when tested in the LMS.  I'll try again with a newly created version.

bradenmacdonald:
@smarnach I'm aware of that, but I think it's fine. I don't believe this is yet in use in any courses. The important thing is that the authors are able to do the conversion.

smarnach:
If you add multilple conditional messages and view them from the Step Builder top level in studio, it's hard to see which condition belongs to which message, since the condition is separate from the message by a horizontal rule, while the messages aren't separated from each other at all:

screenshot_2015-11-02_22-26-45

smarnach:
I reviewed everything that's there up to now.  The code is looking good, and everything is working fine with the latest changes as well.  So the only thing left is to make the tests pass. :)

I'll take a final look tomorrow morning, and will merge and do the hash update on the Solutions fork of edx-platform if everything looks good.

bradenmacdonald:
Thanks a lot, @smarnach :)

smarnach:
All looking good, :+1:.

PR #92, opened 2015-11-17
Fix flaky tests

mtyaka:
This PR tries to fix the situation with flaky tests by:
- making a recently introduced flaky test (`test_scroll_into_view`) more robust by increasing the scroll check tolerance,
- enabling Circle CI builds in parallel mode (tests finish in 12-15 minutes compared to >30 minutes on Travis CI where they are not parallelized.
- Suppressing benign "broken pipe" errors that were polluting the logs. These errors are suppressed by default in django 1.8 (https://code.djangoproject.com/ticket/4444)
- Sleep before reloading xblock from the database in one of the flaky tests (`test_author_changes`). The block was reloaded immediately after the browser page was loaded in selenium. Selenium blocks while the page is loading, but it doesn't block on any ajax requests that happen after page load. The mentoring xblock requests some data at page load time via ajax (`initXBlockView`), which was probably causing a race condition where both the ajax request and the `reload_pb_block` test helper method were trying to `get_or_create` block instance data at the same time. Unfortunately I wasn't able to prove this, but the `MultipleObjectsReturned` hasn't occurred since this change.


bradenmacdonald:
Can switch the build status graphic in README.md to Circle CI, and remove `travis.yml`? Or is there some reason for keeping it around?

mtyaka:
@bradenmacdonald I switched the graphic and removed `travis.yml`, I don't think there's any reason to keep it.

bradenmacdonald:
@mtyaka Great :)

If you remove the old `is_broken_pipe_error` that you accidentally left in, this is :+1:

PR #95, opened 2016-01-21
Django migrations

Kelketek:
Move old migrations into south_migrations folder, create new folder with Django 1.8 migrations.

Use the same basic procedure as in https://github.com/open-craft/problem-builder/pull/94 but this time install the requirements.txt for problem-builder, since the master version requires a newer version of xblockutils.

In addition, check to see if you can get the answer sharing block to work between a couple of users, as that will verify the table is there and working.

@itsjeyd


itsjeyd:
@Kelketek Looks good, :+1: once you get a green build.

PR #97, opened 2016-02-04
Bump version number.

Kelketek:
@bradenmacdonald Who would we need to review this? It's needed for https://github.com/edx/edx-platform/pull/11312 All it does is bump the version number.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ install:
     - "pip install -r requirements.txt"
     - "pip install -r $VIRTUAL_ENV/src/xblock-sdk/requirements/base.txt"
     - "pip install -r $VIRTUAL_ENV/src/xblock-sdk/requirements/test.txt"
-    - "pip uninstall -y xblock-problem-builder && python setup.py sdist && pip install dist/xblock-problem-builder-2.0.tar.gz"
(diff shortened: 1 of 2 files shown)


bradenmacdonald:
:+1: 
It's ok that CircleCI failed since CircleCI is not enabled on this branch.
We should also get a plus 1 from any one of the reviewers on your other PR, then go ahead and merge this.

BTW In future, please don't create the git tag (`v2.0.1`) until after this version bump PR has merged. Then once it has merged, go ahead and tag the merge commit then push the tag. The main reason for this is that tags shouldn't be changed once published (similar to the reason we don't do force pushes to master), and you want to leave yourself some room to fix things in case the version bump PR has an issue.

Kelketek:
@nedbat FYI, version update PR to address the tagging issue for the problem builder hash bump.

nedbat:
I know this is a pain, but TBH I would not count on the moved tag at all.  It's safer to go to 2.0.2.

Kelketek:
@nedbat Very well. I'll update the version again. One moment...

Kelketek:
@nedbat Updated.

nedbat:
:+1:

PR #99, opened 2016-02-19
Fix exponential rerendering after hitting next button too many times.

Kelketek:
After each tap of the 'next' button, the updateDisplay function would be called and apply itself to the next button again, causing exponential rendering runs which would freeze things up after a certain number of questions.

@e-kolpakov This one's a critical hotfix. Please look at it first opportunity, if you can, so we can deploy it Monday.


e-kolpakov:
@Kelketek :+1: 

It would be helpful if you included a course export to replicate the issue - it would reduce the review time significantly.

PR #101, opened 2016-02-23
OC-1387 Always show tooltips feedback if tips are present.

jbzdak:
Testing instructions: 

# Problem Builder 

## MRQ 

Create following MRQ block, and verify that behavior is unchanged: 

1.That when you create submit you get question feedback: "Thank you for answering!" 
2. And after clicking check-mark you get tooltip. 

```
<problem-builder display_name="Default Title" weight="1" mode="standard">
    <html_demo>
        <p>Please answer the questions below.</p>
    </html_demo>

    <pb-mrq name="mrq_1_1" question="What do you like in this MRQ?" message="Thank you for answering!" required_choices='["gracefulness","elegance","beauty"]'>
        <pb-choice value="elegance">Its elegance</pb-choice>
        <pb-choice value="beauty">Its beauty</pb-choice>
        <pb-choice value="gracefulness">Its gracefulness</pb-choice>
        <pb-choice value="bugs">Its bugs</pb-choice>

        <pb-tip values='["gracefulness"]'>This MRQ is indeed very graceful</pb-tip>
        <pb-tip values='["bugs"]'>Nah, there aren't any!</pb-tip>
    </pb-mrq>

    <pb-message type="completed">
        <p>Congratulations!</p>
    </pb-message>
    <pb-message type="incomplete">
        <p>Still some work to do...</p>
    </pb-message>
</problem-builder>
```

## MCQ

Create following MCQ block, and verify that behavior is changed.

1. When you select a,b or d you get a tooltip straight away 
2. When you select c you get question level feedback

```
<vertical_demo>
    <html_demo><h1>Problem Builder Demo</h1></html_demo>

    <problem-builder display_name="MCQ With Tips" weight="1" mode="standard">
        <html_demo>
            <p>Like the MRQ above, multiple choice and rating questions can also provide feedback based on which answer was selected and whether the answer is correct or not. Here's an example:</p>
        </html_demo>

        <pb-mcq name="subjunctive"
          message="This is question level feedback"
          question="Which sentence correctly uses the subjunctive case?" correct_choices='["d"]'>
            <pb-choice value="a">To buy or not to buy, that is the question.</pb-choice>
            <pb-choice value="b">Renting gives you more flexibility.</pb-choice>
            <pb-choice value="c">If I was you, I'd buy the house.</pb-choice>
            <pb-choice value="d">If I were you, I'd rent the house.</pb-choice>
            <pb-tip values='["a","b"]'>This sentence is not discussing hypotheticals or impossibilities.</pb-tip>
            <pb-tip values='["d"]'>Correct. "was" has become "were" to indicate the subjunctive case.</pb-tip>
        </pb-mcq>

        <pb-message type="incomplete">
            <p>Read the tip above and try again.</p>
        </pb-message>

        <pb-message type="completed">
            <p>Nicely done. If I were you, I'd be proud of myself.</p>
        </pb-message>
    </problem-builder>
    
    <html_demo>
        <h1>See the details</h1>
        <p>Check out the XML source below to get an idea of how these mentoring questions were set up and what kind of options authors have available.</p>
    </html_demo>
</vertical_demo>
```

## Step Builder

It is not affected AFAIK, as doesn't display tooltips at all, and can optionally display question level feedback messages (on the review block).

### DIFF
--- a/problem_builder/mcq.py
+++ b/problem_builder/mcq.py
@@ -51,6 +51,16 @@ class MCQBlock(SubmittingXBlockMixin, QuestionnaireAbstractBlock):
     CATEGORY = 'pb-mcq'
     STUDIO_LABEL = _(u"Multiple Choice Question")
 
+    message = String(
+        display_name=_("Message"),
+        help=_(
+            "General feedback provided when submitting. "
+            "(This is not shown if there is a more specific feedback tip for the choice selected by the learner.)"
+        ),
+        scope=Scope.content,
+        default=""
+    )
+
     student_choice = String(
         # {Last input submitted by the student
         default="",
@@ -64,7 +74,7 @@ class MCQBlock(SubmittingXBlockMixin, QuestionnaireAbstractBlock):
         list_values_provider=QuestionnaireAbstractBlock.choice_values_provider,
         list_style='set',  # Underered, unique items. Affects the UI editor.
     )
-    editable_fields = QuestionnaireAbstractBlock.editable_fields + ('correct_choices', )
+    editable_fields = QuestionnaireAbstractBlock.editable_fields + ('message', 'correct_choices', )
 
     def describe_choice_correctness(self, choice_value):
         if choice_value in self.correct_choices:
--- a/problem_builder/public/js/questionnaire.js
+++ b/problem_builder/public/js/questionnaire.js
@@ -123,43 +123,30 @@ function MCQBlock(runtime, element) {
 
             var messageView = MessageView(element, mentoring);
 
-            if (result.message) {
-                var msg = '<div class="message-content">' + result.message + '</div>' +
-                          '<div class="close icon-remove-sign fa-times-circle"></div>';
-                messageView.showMessage(msg);
-            } else { messageView.clearResult(); }
+            messageView.clearResult();
 
-            display_message(result.message, messageView, options.checkmark);
+            var choiceInputDOM = $('.choice-selector input[value="'+ result.submission +'"]');
+            var choiceDOM = choiceInputDOM.closest('.choice');
+            var choiceResultDOM = $('.choice-result', choiceDOM);
+            var choiceTipsDOM = $('.choice-tips', choiceDOM);
 
-            var choiceInputs = $('.choice-selector input', element);
-            $.each(choiceInputs, function(index, choiceInput) {
-                var choiceInputDOM = $(choiceInput);
-                var choiceDOM = choiceInputDOM.closest('.choice');
-                var choiceResultDOM = $('.choice-result', choiceDOM);
-                var choiceTipsDOM = $('.choice-tips', choiceDOM);
-
-                if (choiceInputDOM.prop('checked')) {  // We're showing previous answers,
-                                                       // so go ahead and display results as well
-                    if (result.status === "correct" && choiceInputDOM.val() === result.submission) {
-                        choiceDOM.addClass('correct');
-                        choiceResultDOM.addClass('checkmark-correct icon-ok fa-check');
-                    }
-                    else if (choiceInputDOM.val() === result.submission || _.isNull(result.submission)) {
-                        choiceDOM.addClass('incorrect');
-                        choiceResultDOM.addClass('checkmark-incorrect icon-exclamation fa-exclamation');
-                    }
-
-                    if (result.tips && choiceInputDOM.val() === result.submission) {
-                        mentoring.setContent(choiceTipsDOM, result.tips);
-                    }
-
-                    choiceResultDOM.off('click').on('click', function() {
-                        if (choiceTipsDOM.html() !== '') {
-                            messageView.showMessage(choiceTipsDOM);
-                        }
-                    });
+            // We're showing previous answers, so go ahead and display results as well
+            if (choiceInputDOM.prop('checked')) {
+                display_message(result.message, messageView, options.checkmark);
+                if (result.status === "correct") {
+                    choiceInputDOM.addClass('correct');
+                    choiceResultDOM.addClass('checkmark-correct icon-ok fa-check');
+                } else {
+                    choiceDOM.addClass('incorrect');
+                    choiceResultDOM.addClass('checkmark-incorrect icon-exclamation fa-exclamation');
                 }
-            });
+
+                if (result.tips) {
+                    mentoring.setContent(choiceTipsDOM, result.tips);
+                    messageView.showMessage(choiceTipsDOM);
+                }
+            }
+
 
             if (_.isNull(result.submission)) {
                 messageView.showMessage('<div class="message-content">You have not provided an answer.</div>' +
--- a/problem_builder/tests/integration/test_mentoring.py
+++ b/problem_builder/tests/integration/test_mentoring.py
@@ -137,12 +137,16 @@ class ProblemBuilderQuestionnaireBlockTest(ProblemBuilderBaseTest):
         self.assertFalse(choice_input.is_selected())
 
     def _standard_filling(self, answer, mcq, mrq, rating):
+        self.scroll_to(answer)
         answer.send_keys('This is the answer')
+        self.scroll_to(mcq)
         self.click_choice(mcq, "Yes")
         # 1st, 3rd and 4th options, first three are correct, i.e. two mistakes: 2nd and 4th
+        self.scroll_to(mrq, 300)
         self.click_choice(mrq, "Its elegance")
         self.click_choice(mrq, "Its gracefulness")
         self.click_choice(mrq, "Its bugs")
+        self.scroll_to(rating)
         self.click_choice(rating, "4")
 
     # mcq and rating can't be reset easily, but it's not required; listing them here to keep method signature similar
@@ -153,8 +157,11 @@ class ProblemBuilderQuestionnaireBlockTest(ProblemBuilderBaseTest):
                 checkbox.click()
 
     def _standard_checks(self, answer, mcq, mrq, rating, messages):
+        self.scroll_to(answer)
         self.assertEqual(answer.get_attribute('value'), 'This is the answer')
+        self.scroll_to(mcq)
         self._assert_feedback_showed(mcq, 0, "Great!", click_choice_result=True)
+        self.scroll_to(mrq, 300)
         self._assert_feedback_showed(
             mrq, 0, "This is something everyone has to like about this MRQ",
             click_choice_result=True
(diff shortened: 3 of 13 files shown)


problem_builder/tests/integration/test_titles.py
```diff
@@ -72,7 +72,7 @@ class StepTitlesTest(SeleniumXBlockTest):
     )
 
     mcq_template = """
-        <problem-builder mode="{{mode}}">
```
jbzdak: This is unrelated, but made CircleCi tests to fail (strangely it passed on my system) --- ``{{`` is actually an escape sequence that means ``{`` in str.format.

problem_builder/tests/integration/test_questionnaire.py
```diff
@@ -115,17 +112,9 @@ def submit_answer_and_assert_messages(mcq1_answer, mcq2_answer, item_feedback1,
             mcq1_tips = mcq1.find_element_by_css_selector(".choice-tips .tip p")
             mcq2_tips = mcq2.find_element_by_css_selector(".choice-tips .tip p")
 
-            self.assertTrue(mcq1_feedback.is_displayed())
```
jbzdak: It's not that I just removed assertions :) Feedback is now hidden (so these assertions are not needed), and tips are displayed without the need to click on them.

problem_builder/mcq.py
```diff
@@ -51,6 +51,13 @@ class MCQBlock(SubmittingXBlockMixin, QuestionnaireAbstractBlock):
     CATEGORY = 'pb-mcq'
     STUDIO_LABEL = _(u"Multiple Choice Question")
 
+    message = String(
```
jbzdak: I moved the message from QuestionnaireAbstractBlock as MCQ and MRQ have different help text

conf/locale/en/LC_MESSAGES/django-partial.po
```diff
@@ -391,10 +391,14 @@ msgid ""
...
 
-#: mentoring/questionnaire.py:73
+#: mentoring/mrq.py:76
 msgid "General feedback provided when submiting"
 msgstr ""
 
+#: mentoring/mcq.py:56
+msgid "General feedback provided when submitting, it is not shown if question selected by user its own feedback tip."
+msgstr ""
+
```
bradenmacdonald: As discussed, you don't need to update the .po file, so I'd rather it not even show up in the diff.

problem_builder/mcq.py
```diff
@@ -51,6 +51,16 @@ class MCQBlock(SubmittingXBlockMixin, QuestionnaireAbstractBlock):
...
+        display_name=_("Message"),
+        help=_(
+            "General feedback provided when submitting, it is not shown if "
+            "question selected by user its own feedback tip."
```
bradenmacdonald: "General feedback shown to the learner when submitting. "
"(This is not shown if there is a more specific feedback tip for the choice selected by the learner.)"

(The text you used above is not grammatically correct because a comma "," should not be used to separate two independent clauses)

problem_builder/public/js/questionnaire.js
```diff
@@ -131,32 +131,22 @@ function MCQBlock(runtime, element) {
...
-                    mentoring.setContent(choiceTipsDOM, result.tips);
-                }
+            var choiceInputDOM = $('.choice-selector input[value="'+ result.submission +'"]');
+            var choiceDOM = choiceInputDOM.closest('.choice');
+            var choiceResultDOM = $('.choice-result', choiceDOM);
+            var choiceTipsDOM = $('.choice-tips', choiceDOM);
+            if (result.status === "correct"){
+                choiceInputDOM.addClass('correct');
+                choiceResultDOM.addClass('checkmark-correct icon-ok fa-check');
+            }else{
```
bradenmacdonald: Nit: Here, 3 lines above, and 5 lines below, there should be space before/after the `{` or `}`, for consistency with the rest of the file.

problem_builder/public/js/questionnaire.js
```diff
@@ -131,32 +131,22 @@ function MCQBlock(runtime, element) {
 
             display_message(result.message, messageView, options.checkmark);
 
-            var choiceInputs = $('.choice-selector input', element);
-            $.each(choiceInputs, function(index, choiceInput) {
```
bradenmacdonald: Nice cleanup of the code here :)

problem_builder/tests/integration/test_titles.py
bradenmacdonald: Yeah, I think these were originally copied from a django template. Not sure why they worked before this change.

Incidentally, the `mode` parameter is deprecated, and we will eventually remove support for it.

problem_builder/templates/xml/mentoring_demo.xml
```diff
@@ -1,68 +1,5 @@
 <vertical_demo>
     <html_demo><h1>Problem Builder Demo</h1></html_demo>
-    <problem-builder display_name="Basic Usage: Questions" weight="1" mode="standard">
```
bradenmacdonald: Why did you remove this?

problem_builder/templates/xml/mentoring_demo.xml
```diff
@@ -75,7 +12,6 @@
             <pb-choice value="c">If I was you, I'd buy the house.</pb-choice>
             <pb-choice value="d">If I were you, I'd rent the house.</pb-choice>
             <pb-tip values='["a","b"]'>This sentence is not discussing hypotheticals or impossibilities.</pb-tip>
-            <pb-tip values='["c"]'>This sentence should be using the subjunctive case since it's dicsussing a hypothetical, but "was" is the wrong case.</pb-tip>
```
bradenmacdonald: And why did you remove this + the other changes in this file?

problem_builder/tests/integration/test_questionnaire.py
bradenmacdonald: The test should be checking that if you choose an option with a tip, then `mcq1_tips.text` contains that tip. But if you choose an option without a tip, the test should make sure that `mcq1_tips.text` contains the general feedback. So your change to this method is fine, but we need to add some instances of

```python
submit_answer_and_assert_messages(..., "Feedback message 2")
```

to confirm the question-level feedback is working properly.

bradenmacdonald:
Tested and it works well, but I had some questions and think we need a bit more test coverage.

problem_builder/tests/integration/test_questionnaire.py
jbzdak: Makes sense! Thanks :)

problem_builder/templates/xml/mentoring_demo.xml
jbzdak: Commited by mistake, sorry for the noise.

problem_builder/tests/integration/xml/mcq_1.xml
```diff
@@ -6,7 +6,6 @@
             <pb-choice value="understand">I don't understand</pb-choice>
 
             <pb-tip values='["yes"]'>Great!</pb-tip>
-            <pb-tip values='["maybenot"]'>Ah, damn.</pb-tip>
```
jbzdak: These two were removed on purpose. To test questionaire I needed to have options with tooltips and without them. So I removed tooltips from selected choices. Other tests seem to be not affected.

problem_builder/mcq.py
```diff
@@ -51,6 +51,16 @@ class MCQBlock(SubmittingXBlockMixin, QuestionnaireAbstractBlock):
     CATEGORY = 'pb-mcq'
     STUDIO_LABEL = _(u"Multiple Choice Question")
 
+    message = String(
+        display_name=_("Message"),
+        help=_(
+            "General feedback provided when submitting, it is not shown if."
```
bradenmacdonald: "General feedback shown to the learner when submitting. "

bradenmacdonald:
@jbzdak Just had two minor notes. Also apparently this needs a rebase now :/  Please address those last two little things and rebase. I'll test the rebased version one last time and give you +1.

problem_builder/tests/integration/test_mentoring.py
```diff
@@ -165,18 +172,23 @@ def _standard_checks(self, answer, mcq, mrq, rating, messages):
...
         self._assert_feedback_showed(mrq, 3, "Nah, there aren't any!", click_choice_result=True, success=False)
+        self.scroll_to(rating)
         self._assert_feedback_showed(rating, 3, "I love good grades.", click_choice_result=True)
         self.assertTrue(messages.is_displayed())
+        self.scroll_to(messages)
         self.assertEqual(messages.text, "FEEDBACK\nNot done yet")
 
     def _feedback_customized_checks(self, answer, mcq, mrq, rating, messages):
         # Long answer: Previous answer and feedback visible
+        self.scroll_to(answer)
```
jbzdak: Unrelated tests failed locally, as there were trying to click on invisible elements, this fixed the issue for me, we can either leave it for the merint of making tests more robust, or just remove. Really I feel that selenium should be doing this scrolling.

bradenmacdonald:
@jbzdak The problem with the currently-failing tests is ultimately caused by the fact that you are calling `self.load_scenario(scenario)` twice within the same `test_feedback_persistence_tips` test case, which as you've found does not work. Unfortunately as is often the case with XBlock SDK, you'll see all kinds of random ORM/DB/txn error messages, so this type of thing is hard to diagnose.

All you need to do to fix the issue is replace:

```python
self.reload_student_view()
mentoring = self.load_scenario(scenario)
```

with

```python
mentoring = self.reload_student_view()
```

problem_builder/public/js/questionnaire.js
```diff
@@ -123,43 +123,30 @@ function MCQBlock(runtime, element) {
 
             var messageView = MessageView(element, mentoring);
 
-            if (result.message) {
```
jbzdak: Logic used to display message was changed, so take a second glance here.

problem_builder/tests/integration/test_mentoring.py
bradenmacdonald: @jbzdak Selenium *should* be doing this scrolling - is it possible that the elements in question were just hidden (in the process of becoming visible), and this scroll_to has a side effect of fixing it since it acts as a short delay?

These `scroll_to` calls seem very hacky, and I'd rather find a more direct way of addressing the problem. 

Please try this way of fixing it instead (which is a nice cleanup, if it works): 351705cb5caa8425858b013d958fba1b8b84b3c2

problem_builder/public/js/questionnaire.js
bradenmacdonald: Thanks - looks good.

bradenmacdonald:
Looking good! I just want a cleaner fix for those tests - I provided a suggestion that may do the trick. Then this will be good to go.

jbzdak:
@bradenmacdonald  I have integrated your fix, and everything works fine now.

bradenmacdonald:
:+1:

jbzdak:
Thanks! Merging.

PR #102, opened 2016-02-25
Problem Builder: Add instance-wide option to hide previous answers for MCQs

itsjeyd:
This PR introduces an instance-wide option to hide previous answers (and feedback) for MCQs in Problem Builder. When this option is enabled, students will still receive per-question feedback after providing answers to MCQs and submitting the containing block, but that feedback will not be visible when they return to the block, and all choices for a given MCQ will be unchecked.

**Screenshots: LMS**

Default settings:

lms-pb-option-disabled

Previous answers hidden:

lms-pb-option-enabled

---

**Screenshots: Apros**

Default settings:

apros-pb-option-disabled

Previous  answers hidden:

apros-pb-option-enabled

**Test instructions**

1. Create a subsection with two units and add a Problem Builder block to each one.

2. Add a long answer, MCQ, Rating, MRQ, and Slider to one of the Problem Builder blocks.

3. Publish the units.

4. In LMS/Apros, provide answers to all questions and click "Submit".

5. Navigate to the other unit, then come back to the first unit. Observe that Problem Builder shows previous answers and feedback for all questions.

6. In `lms.env.json`, enable the new option by modifying the `XBLOCK_SETTINGS` for `mentoring`:

   ```js
        "mentoring": {
            "theme": {
                // ...
            },
            "options": {
                "pb_mcq_hide_previous_answer": true
            }
        },
   ```

7. Restart LMS/Apros and reload the page showing the first unit. Observe that Problem Builder does not show previous answers and feedback for MCQs.

### DIFF
--- a/problem_builder/mentoring.py
+++ b/problem_builder/mentoring.py
@@ -42,6 +42,7 @@ from .step_review import ReviewStepBlock
 
 from xblockutils.helpers import child_isinstance
 from xblockutils.resources import ResourceLoader
+from xblockutils.settings import XBlockWithSettingsMixin, ThemableXBlockMixin
 from xblockutils.studio_editable import (
     NestedXBlockSpec, StudioEditableXBlockMixin, StudioContainerXBlockMixin, StudioContainerWithNestedXBlocksMixin,
 )
@@ -84,7 +85,8 @@ PARTIAL = 'partial'
 @XBlock.needs("i18n")
 @XBlock.wants('settings')
 class BaseMentoringBlock(
-        XBlock, XBlockWithTranslationServiceMixin, StudioEditableXBlockMixin, MessageParentMixin
+    XBlock, XBlockWithTranslationServiceMixin, XBlockWithSettingsMixin,
+    StudioEditableXBlockMixin, ThemableXBlockMixin, MessageParentMixin,
 ):
     """
     An XBlock that defines functionality shared by mentoring blocks.
@@ -161,45 +163,22 @@ class BaseMentoringBlock(
             return [self.display_name]
         return []
 
-    def get_settings(self, settings_key, default):
-        """
-        Get settings identified by `settings_key` from settings service.
-
-        Fall back on `default` if settings service is unavailable
-        or settings have not been customized.
-        """
-        settings_service = self.runtime.service(self, "settings")
-        if settings_service:
-            xblock_settings = settings_service.get_settings_bucket(self)
-            if xblock_settings and settings_key in xblock_settings:
-                return xblock_settings[settings_key]
-        return default
-
-    def get_theme(self):
-        """
-        Get theme settings for this block from settings service.
-
-        Fall back on default (LMS) theme if settings service is not available,
-        or theme has not been customized.
-        """
-        return self.get_settings(self.theme_key, _default_theme_config)
-
-    def include_theme_files(self, fragment):
-        theme = self.get_theme()
-        theme_package, theme_files = theme['package'], theme['locations']
-        for theme_file in theme_files:
-            fragment.add_css(ResourceLoader(theme_package).load_unicode(theme_file))
-
     def get_options(self):
         """
         Get options settings for this block from settings service.
 
-        Fall back on default options if settings service is not available
-        or options have not been customized.
+        Fall back on default options if xblock settings have not been customized at all
+        or no customizations for options available.
         """
-        return self.get_settings(self.options_key, _default_options_config)
+        xblock_settings = self.get_xblock_settings(_default_options_config)
+        if xblock_settings and self.options_key in xblock_settings:
+            return xblock_settings[self.options_key]
+        return _default_options_config
 
     def get_option(self, option):
+        """
+        Get value of a specific instance-wide `option`.
+        """
         return self.get_options()[option]
 
     @XBlock.json_handler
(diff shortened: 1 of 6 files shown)


problem_builder/public/js/mentoring.js
```diff
@@ -8,7 +8,9 @@ function MentoringBlock(runtime, element) {
     var attemptsTemplate = _.template($('#xblock-attempts-template').html());
     var data = $('.mentoring', element).data();
     var children = runtime.children(element);
-    var steps = runtime.children(element).filter(function(c) { return c.element.className.indexOf('assessment_step_view') > -1; });
+    var steps = runtime.children(element).filter(function(c) {
+        return $(c.element).attr("class").indexOf('assessment_step_view') > -1;
+    });
```
bradenmacdonald: In case you're wondering why you had to make this change: https://openedx.atlassian.net/browse/TNL-4161

I believe the earlier version is "better" but it doesn't work with the latest edx-platform master. The new version you have here should always work, so we may as well keep this change.

problem_builder/mentoring.py
```diff
@@ -156,18 +161,28 @@ def get_content_titles(self):
             return [self.display_name]
         return []
 
-    def get_theme(self):
+    def get_settings(self, settings_key, default):
```
Kelketek: Perhaps it would be best to take advantage of the XBlock settings mixin from utils.

https://github.com/edx/xblock-utils/blob/3b58c757f06943072b170654d676e95b9adb37b0/xblockutils/settings.py#L11

problem_builder/mentoring.py
```diff
@@ -156,18 +161,28 @@ def get_content_titles(self):
...
         if settings_service:
             xblock_settings = settings_service.get_settings_bucket(self)
-            if xblock_settings and self.theme_key in xblock_settings:
-                return xblock_settings[self.theme_key]
-        return _default_theme_config
+            if xblock_settings and settings_key in xblock_settings:
+                return xblock_settings[settings_key]
+        return default
+
+    def get_theme(self):
```
Kelketek: Likewise for get_theme here and all. I'm kind of surprised this isn't already done. Am I missing something or did we just make these mixins after the fact?

Kelketek:
@itsjeyd Just a couple of comments.

problem_builder/mentoring.py
itsjeyd: @Kelketek Not sure, but `BaseMentoringBlock` is now subclassing `XBlockWithSettingsMixin` and `ThemableXBlockMixin`.

Kelketek:
:+1: Though please squash at least some of the commits first.

itsjeyd:
@Kelketek Thanks, done.

problem_builder/mentoring.py
```diff
@@ -63,6 +64,10 @@
     'locations': ['public/themes/lms.css']
```
bradenmacdonald: @itsjeyd Is this `_default_theme_config` being used? Looks to me like with `ThemableXBlockMixin` you need to define `default_theme_config` as an attribute on the class instead.

problem_builder/tests/unit/test_problem_builder.py
```diff
@@ -1,9 +1,15 @@
...
+
 from mock import MagicMock, Mock, patch
+from random import random
+
 from xblock.field_data import DictFieldData
+
 from problem_builder.mcq import MCQBlock
-from problem_builder.mentoring import MentoringBlock, MentoringMessageBlock, _default_theme_config
+from problem_builder.mentoring import (
+    MentoringBlock, MentoringMessageBlock, _default_theme_config, _default_options_config
```
bradenmacdonald: `_default_theme_config` doesn't seem to be used - also that should have caused a pylint error, shouldn't it have?

problem_builder/tests/unit/test_problem_builder.py
```diff
@@ -71,76 +76,65 @@ def setUp(self):
...
 
     def test_author_preview_view_calls_include_theme_files(self):
+        self.service_mock.get_settings_bucket = Mock(return_value={})
         with patch.object(self.block, 'include_theme_files') as patched_include_theme_files:
             fragment = self.block.author_preview_view({})
             patched_include_theme_files.assert_called_with(fragment)
 
 
+@ddt.ddt
+class TestMentoringBlockOptions(unittest.TestCase):
```
bradenmacdonald: I think it would have been good to leave in one or two theming tests. I know the logic is moved to xblock-utils and tested there, but if someone else changes that API we need to know.

problem_builder/tests/unit/test_problem_builder.py
itsjeyd: @bradenmacdonald #103 removes `_default_theme_config` completely.

The `pylint` command we run on CircleCI currently [enables](https://github.com/open-craft/problem-builder/blob/master/circle.yml#L18) only a handful of checks. `unused-import` is not one of them, that's why `_default_theme_config` fell through the cracks here.

bradenmacdonald: @itsjeyd Ah. Would you mind enabling `unused-import` for us?

itsjeyd: @bradenmacdonald No problem, added to #103.

bradenmacdonald: @itsjeyd Thanks!

PR #103, opened 2016-03-01
Ensure default theme is used if theme has not been customized

itsjeyd:
Follows up on #102.

**Test instructions**

1. Remove theme settings from `lms.envs.json`.
2. Access unit containing Problem Builder block and check if rules in `public/themes/lms.css` are applied. For example, `.choice-result` should have `display: table-cell` (first rule in `public/themes/lms.css`).


Kelketek:
:+1:

PR #104, opened 2016-03-03
Problem Builder: Add instance-wide option to hide global feedback and results for Long Answer blocks

itsjeyd:
This PR introduces an instance-wide option to hide global feedback and results (correct/incorrect) icons for Long Answer blocks in Problem Builder. When this option is enabled (and there are some attempts left), students will still receive global and per-question feedback after providing answers and submitting the containing block, but that feedback will not be visible when they return to the block.

cf. [MCKIN-3858](https://edx-wiki.atlassian.net/browse/MCKIN-3858)

**Screenshots: LMS**

Default settings:

lms-pb-option-disabled

Feedback hidden:

lms-pb-option-enabled

---

**Screenshots: Apros**

Default settings:

apros-pb-option-disabled

Feedback hidden:

apros-pb-option-enabled

**Test instructions**

1. Create a subsection with two units and add a Problem Builder block to each one.

2. Add a long answer, MCQ, Rating, MRQ, and Slider to one of the Problem Builder blocks.

3. Publish the units.

4. In LMS/Apros, provide answers to all questions and click "Submit".

5. Navigate to the other unit, then come back to the first unit. Observe that Problem Builder shows global and per-question feedback for all questions.

6. In `lms.env.json`, enable the new option by modifying the `XBLOCK_SETTINGS` for `mentoring`:

   ```js
        "mentoring": {
            "theme": {
                // ...
            },
            "options": {
                "pb_hide_feedback_if_attempts_remain": true
            }
        },
   ```

7. Restart LMS/Apros and reload the page showing the first unit. Observe that Problem Builder does not show global feedback and results for the Long Answer are hidden.

### DIFF
--- a/problem_builder/mentoring.py
+++ b/problem_builder/mentoring.py
@@ -60,7 +60,8 @@ log = logging.getLogger(__name__)
 loader = ResourceLoader(__name__)
 
 _default_options_config = {
-    'pb_mcq_hide_previous_answer': False
+    'pb_mcq_hide_previous_answer': False,
+    'pb_hide_feedback_if_attempts_remain': False,
 }
 
 
@@ -174,11 +175,11 @@ class BaseMentoringBlock(
             return xblock_settings[self.options_key]
         return _default_options_config
 
-    def get_option(self, option):
+    def get_option(self, option, default=False):
         """
         Get value of a specific instance-wide `option`.
         """
-        return self.get_options()[option]
+        return self.get_options().get(option, default)
 
     @XBlock.json_handler
     def view(self, data, suffix=''):
@@ -563,9 +564,10 @@ class MentoringBlock(BaseMentoringBlock, StudioContainerXBlockMixin, StepParentM
         """
         results = []
         completed = True
-        show_message = bool(self.student_results)
+        hide_feedback = self.get_option("pb_hide_feedback_if_attempts_remain") and not self.max_attempts_reached
+        show_message = (not hide_feedback) and bool(self.student_results)
 
-        # In standard mode, all children is visible simultaneously, so need collecting responses from all of them
+        # In standard mode, all children are visible simultaneously, so need to collect results for all of them
         for child in self.steps:
             child_result = child.get_last_result()
             results.append([child.name, child_result])
--- a/problem_builder/mentoring.py
+++ b/problem_builder/mentoring.py
@@ -175,11 +175,11 @@ class BaseMentoringBlock(
             return xblock_settings[self.options_key]
         return _default_options_config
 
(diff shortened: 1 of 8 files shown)


problem_builder/tests/integration/test_mentoring.py
```diff
@@ -226,32 +290,40 @@ def test_feedbacks_and_messages_is_not_shown_on_first_load(self):
...
+        (
+            {
+                'pb_mcq_hide_previous_answer': True,
+                'pb_hide_feedback_if_attempts_remain': True
+            },
+            "_mcq_hide_previous_answer_hide_feedback_checks"
+        ),
+    )
+    @ddt.unpack
+    def test_persistence(self, options, after_reload_checks):
```
jbzdak: DDT stands for data driven testing, I think that names of methods that perform assertions don't qualify as "data" :). I think that splitting this into a helper method (very similar to ``test_persistence``) and four methods that launch it would be beneficial in this case to the overall clarity.

problem_builder/tests/integration/test_mentoring.py
```diff
@@ -203,19 +266,20 @@ def reload_student_view(self):
...
+
         wait.until(did_load_homepage, u"Workbench home page should have loaded")
         mentoring = self.go_to_view("student_view")
-        self.wait_until_visible(self._get_xblock(mentoring, "feedback_mcq_2"))
+        submit = self._get_submit(mentoring)
+        self.wait_until_visible(submit)
         return mentoring
 
-    def test_feedbacks_and_messages_is_not_shown_on_first_load(self):
+    def test_feedback_and_messages_not_shown_on_first_load(self):
```
jbzdak: I really appreciate cleaning some of the english names here!

problem_builder/mentoring.py
```diff
@@ -174,11 +175,11 @@ def get_options(self):
             return xblock_settings[self.options_key]
         return _default_options_config
 
-    def get_option(self, option):
+    def get_option(self, option, default=False):
```
jbzdak: NIT: I'd rather either have a default of ``None`` or no default for default parameter. ``None`` is a sensible default for every type of settings parameter, while ``False`` only sensible for bools. 

From these two options I'd rather have no default, and have the default specified when parameter is read.

jbzdak:
@itsjeyd It works, I have some minor remarks on code, so conditionally on fixing these/convincing me that status quo is better :+1: .

problem_builder/tests/integration/test_mentoring.py
itsjeyd: @jbzdak Fair enough :) Updated!

PR #107, opened 2016-03-14
Fix dashboard report image download

omarkhan:
The dashboard report image download can currently fail in 2 different ways:

- If the image is from a different domain and it has been cached by the browser, it won't be converted into a data URL due to cross-domain security issues. A javascript `SecurityError` is thrown, and nothing happens.
- If the image url starts with `//` (no explicit protocol) or the url is relative (e.g. `/static/...`), *and* it hasn't yet been cached by the browser, it will fail to show up in the downloaded report.

This pull request fixes both issues, by:

- Not trying to create a data URL for cross-domain requests
- Expanding relative urls and urls without an explicit protocol

This pull request also extends the integration tests to test report downloads.


Kelketek:
@omarkhan Test failing here, but it may just need to be rerun.

omarkhan:
@Kelketek the tests are passing now, yay

Kelketek:
:+1:

omarkhan:
@bradenmacdonald do you want to review this one as well?

bradenmacdonald:
@omarkhan Not necessary, thanks. I will review the edx-release hash update PR though.

bradenmacdonald:
@omarkhan Actually never mind I'm not going to review any of these PRs. I was only going to review this master PR if it was a merge from `edx-release` to `master`, which can be tricky`, but you just implemented it separately on each branch so that's fine.

omarkhan:
@bradenmacdonald do I need to get thumbs up from @tobz on this PR as well?

bradenmacdonald:
@omarkhan No, the `master` branch of Problem Builder is maintained by OpenCraft alone. EdX review is only required for the `edx-release` branch.

omarkhan:
@bradenmacdonald thanks that's what I figured, just wanted to make sure before merging

PR #109, opened 2016-03-17
Use dashboard image from master branch in tests

omarkhan:
So that the omar/report-download branch can be deleted


bradenmacdonald:
:+1:

PR #110, opened 2016-03-23
Enable freeform answer submit when feedback is hidden

mtyaka:
cf. [MKCIN-3890](https://edx-wiki.atlassian.net/browse/MCKIN-3890)

*Test Instructions*

1. Create a new subunit containing one problem builder freeform question unit, and another arbitrary unit.
2. Enable `pb_hide_feedback_if_attempts_remain` in your `XBLOCK_SETTINGS`.
3. Go to the unit with the freeform question, fill it in, and submit.
4. Go to the next unit, then come back to the unit with the freeform question.
5. Feedback should be hidden, but the submit button should be enabled

This PR also contains a patch that makes MRQ behavior when `pb_mcq_hide_previous_answer` option is enable more consistent with the way MCQs behave. To test:

1. Create a new subunit containing one problem builder MRQ question unit, and another arbitrary unit.
2. Enable `pb_mcq_hide_previous_answer` in your `XBLOCK_SETTINGS`.
3. Go to the unit with the MCQ question, fill it in, and submit.
4. Go to the next unit, then come back to the unit with the MCQ question.
5. Feedback should be hidden, checkmarks next to the choices should NOT be visible, and previously selected choices should not be selected.

### DIFF
--- a/problem_builder/mentoring.py
+++ b/problem_builder/mentoring.py
@@ -478,6 +478,10 @@ class MentoringBlock(BaseMentoringBlock, StudioContainerXBlockMixin, StepParentM
         """
         return '/jump_to_id/{}'.format(self.next_step)
 
+    @property
+    def hide_feedback(self):
+        return self.get_option("pb_hide_feedback_if_attempts_remain") and not self.max_attempts_reached
+
     def get_message(self, completed):
         """
         Get the message to display to a student following a submission in normal mode.
@@ -564,8 +568,7 @@ class MentoringBlock(BaseMentoringBlock, StudioContainerXBlockMixin, StepParentM
         """
         results = []
         completed = True
-        hide_feedback = self.get_option("pb_hide_feedback_if_attempts_remain") and not self.max_attempts_reached
-        show_message = (not hide_feedback) and bool(self.student_results)
+        show_message = (not self.hide_feedback) and bool(self.student_results)
 
         # In standard mode, all children are visible simultaneously, so need to collect results for all of them
         for child in self.steps:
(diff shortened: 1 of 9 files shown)


problem_builder/mentoring.py
```diff
@@ -478,6 +478,10 @@ def next_step_url(self):
         """
         return '/jump_to_id/{}'.format(self.next_step)
 
+    @property
+    def hide_feedback(self):
+        return self.get_option("pb_hide_feedback_if_attempts_remain") and not self.max_attempts_reached
+
```
itsjeyd: @mtyaka I like this refactoring :)

problem_builder/public/js/mentoring_with_steps.js
```diff
@@ -302,7 +302,8 @@ function MentoringWithStepsBlock(runtime, element) {
             var step = steps[i];
             var mentoring = {
                 setContent: setContent,
-                publish_event: publishEvent
+                publish_event: publishEvent,
+                step_builder: true
```
itsjeyd: @mtyaka Since this is a Boolean, could we call it `is_step_builder`?

mtyaka: @itsjeyd Good idea, I renamed the variable.

problem_builder/public/js/mentoring_standard_view.js
```diff
@@ -35,10 +35,13 @@ function MentoringStandardView(runtime, element, mentoring) {
             }
         }
 
+        // Data may have change, we have to re-validate.
+        validateXBlock();
+
```
itsjeyd: @mtyaka `change` ⟶ `changed`. Also, is this still necessary now that answer blocks are handling validation themselves?

itsjeyd:
:+1: once newest comment is addressed and build is green.

problem_builder/public/js/mentoring_standard_view.js
mtyaka: @itsjeyd I fixed the typo, thanks for noticing. This probably isn't necessary for answer blocks anymore (although it shouldn't do any harm), but I believe it still makes sense for other blocks.

mtyaka:
Thanks @itsjeyd!

PR #114, opened 2016-03-30
OC-1441 PR with a proper solution

jbzdak:
Testing instructions on #112.

PR #112, opened 2016-03-29
OC-1441 Detach step elements from DOM or stop playback of closed steps

jbzdak:
This is initial implementation of OC-1441, this stops playback of videos from previous steps. 

## Course set up

What you need is a course where on a step you have a video, and then you can to the next step (you should still hear the video on that next step --- which is the error). 

1. Create a course with two sections, each containing a single subsection and a single unit. 
2. Each unit contains a step-builder with three steps. First step contains a video and a question, second contains a question, and the last one is the review step. 
3. In one of the subsections use a normal Video XBlock, in second use HTML with OOyala player (OOyala set-up is described in the OC-1441 issue). 

Essentially you should have a course that looks like this: 
```
* Section 1
 * Subsection 1
   * Unit 1
    * Step Builder
     * Step 1
      * Video XBlock with a video
      * Question 
    * Step 2 
     * Question
    * Step 3
     * Review
* Section 2
 * Subsection 2
   * Unit 1
    * Step Builder
     * Step 1
      * HTML Xblock with OOyala video
      * Question 
    * Step 2 
     * Question
    * Step 3
     * Review
```

##  Verification instructions

1. Download the course attached to OC-1441 (for anyone from outside OpenCraft: this course just contains two StepBuilder units containing first step with a video, then step with a quiz and then step with a review; first block contains an Youtube video loaded using a Video XBlock, and second one contains an OOlaya video embedded using HTML Xblock). 
2. Please note that OOyala video needs to be manually started by pasting a snippet to a javascript console --- snippet is attached above the video, and looks like that: `OO.ready(function() { OO.Player.create('xxx', 'yyy_zzz'); });`. After you start the video it should work as it was started "normally" so tests are still valid. 
3. Go to video and block, start the video and wait until it loads, fill in the quiz and go to next step. Note that you still hear video music in the background. 
4. Check the same with the OOyala video subsection. 

## Testing instructions

1. Checkout this branch.
2. Go to video and block, start the video and wait until it loads, fill in the quiz and go to next step. Note that the video is no longer playing in the background. 
3. Fill in rest of the quiz, watch the review step, and go back to the video step. Verify that video is still working properly. 
4. Check the same with the OOyala video subsection.

PR #115, opened 2016-04-04
Stopping video playback for hidden steps

jbzdak:
This is a follow up on: #114 and #112. 

## Course set up

What you need is a course where on a step you have a video, and then you can to the next step (you should still hear the video on that next step --- which is the error). 

1. Create a course with three sections, each containing a single subsection and a single unit. 
2. First unit contains following steps: 

  a. A step with Video Block and a question
  b. A step with HTML Block with a youtube video and a question. Example embedding code: `<iframe width="420" height="315" src="https://www.youtube.com/embed/QH2-TGUlwu4" frameborder="0" allowfullscreen></iframe>`
 c. A step with a question (no videos!)
 d. A review step with a HTML with a youtube video, and other review elements. 

3. Second unit contains
 
 a. A step with OOyala video and a question (OOyala set-up is described in the OC-1441 issue)
 b. A step with a question 
 c. A review step  

4. Third unit contains (note that having the same Ooyala video on a two steps in single unit might not work. 

 a. A step with a question 
 a. A review step with a HTML block with ooyala video embedded. 

##  Verification instructions

1. Please note that OOyala video needs to be manually started by pasting a snippet to a javascript console --- snippet is attached above the video, and looks like that: `OO.ready(function() { OO.Player.create('xxx', 'yyy_zzz'); });`. After you start the video it should work as it was started "normally" so tests are still valid. 
3. Go to the block with video, start the video and wait until it loads, fill in the quiz and go to next step. Note that you still hear video music in the background. 
4. Check the same with the OOyala video subsection. 
5. Note that on re-entering the ooyala videos on review step the video doesn't display itself, this is due to the fact that html of [review elements is replaced when displaying this step](https://github.com/open-craft/problem-builder/blob/master/problem_builder/public/js/mentoring_with_steps.js#L92) (this is current behaviour in master), which doesn't play well with ooyala . 

## Testing instructions

For each of the units: 

1. Checkout this branch.
2. Go to video and block, start the video and wait until it loads, fill in the quiz and go to next step. Note that the video is no longer playing in the background. 
3. Fill in rest of the quiz, watch the review step, and go back to the video step. Verify that video is still working properly. 
4. Check if: 

 a. Reloading a page with a video works
 b. Video works on first visit
 c. Video works when re-visiting
 d. Video XBlock rememebers it's position 

5. Note that Ooyala videos in review steps are still broken.

### DIFF
--- a/problem_builder/public/js/lms_util.js
+++ b/problem_builder/public/js/lms_util.js
@@ -1,8 +1,8 @@
 (function () {
 
     /**
-     *  Manager for HTML XBlocks. These blocks are hid by detaching and shown
-     *  by re-attaching them to the DOM.This is only generic way to generically
+     *  Manager for HTML XBlocks. These blocks are hidden by detaching and shown
+     *  by re-attaching them to the DOM. This is only way to generically
      *  handle things like video players (they should stop playing when removed from DOM).
      *
      *  @param html an html xblock
@@ -20,31 +20,29 @@
     
     /**
      *
-     * Manager for HTML Video child. Videos are paused when hiding them, and re-sized when showing them.
+     * Manager for HTML Video child. Videos are re-sized when showing them.
      * @param video an video xblock
      *
      */
     function VideoManager(video) {
         this.show = function () {
-            if (typeof video.resizer === 'undefined'){
+            if (typeof video.resizer === 'undefined') {
                 // This one is tricky: but it looks like resizer is undefined only if the video is on the
                 // step that is initially visible (and then no resizing is necessary)
                 return;
             }
             video.resizer.align();
         };
-        this.hide = function () {
-            if (typeof video.videoPlayer === 'undefined'){
-                // If video player is undefined this means that the video isn't loaded yet, so no need to
-                // pause it.
-                return;
-            }
-            video.videoPlayer.pause();
-        };
+        /**
+         * Videos should be paused when user leaves a step containing a video. There is was a proposed implementation
+         * but since it didn't work on every system we decided to drop it (it was out of scope for current task
+         * nevertheless). See OC-1441 for details.
+         */
+        this.hide = function () {};
     }
     
     /**
-     *  Manager for Plot Xblocks. Handles updating a plot before displaying it.
+     * Manager for Plot Xblocks. Handles updating a plot before displaying it.
      * @param plot
      */
     function PlotManager(plot) {
@@ -65,48 +63,48 @@
         var children = runtime.children(xblock_element);
 
         /**
-         * A list of managers for children than need special care when showing or hiding.
+         * A list of managers for children that need special care when showing or hiding.
          *
          * @type {show, hide}[]
          */
         var managedChildren = [];
 
         /***
-         * This is a workaround for issue where jquery.xblock.Runtime doesn't return HTML Xblocks when querying
+         * This is a workaround for issue where jquery.xblock.Runtime doesn't return HTML blocks when querying
          * for children.
          *
          * This can be removed when:
          *
-         * * We allow inclusion of Ooyala blocks inside ProblemBuilder and our clients migrate to Ooyala, in this case
+         * * We allow inclusion of Ooyala blocks inside StepBuilder and our clients migrate to Ooyala, in this case
          *   we may drop special handling of HTML blocks. See discussions in OC-1441.
          * * We include HTML blocks in runtime.children for runtime of jquery.xblock, then just add
          *   `html: HtmlManager` to `Managers`, and remove this block.
          */
-        $("div.xblock.xblock-student_view.xmodule_HtmlModule", xblock_element).each(function (idx, element) {
-            managedChildren.push(new HtmlManager({element:element}));
+        $("div.xblock.xblock-student_view.xmodule_HtmlModule", xblock_element).each(function(idx, element) {
+            managedChildren.push(new HtmlManager({ element: element }));
         });
 
-        for (var idx = 0; idx < children.length; idx++){
+        for (var idx = 0; idx < children.length; idx++) {
             var child = children[idx];
-            // NOTE: While following assertion is true for e.g Video blocks:
-            // child.type == $(child.element).data('block-type') it is invalidated by for all sb-* blocks
+            // NOTE: While the following assertion is true for e.g Video blocks:
+            // child.type == $(child.element).data('block-type') it is invalid for all sb-* blocks
             var type =  $(child.element).data('block-type');
             var constructor = Managers[type];
-            if (typeof constructor === 'undefined'){
+            if (typeof constructor === 'undefined') {
                 // This block does not requires special care, moving on 
-                continue ;
+                continue;
             }
             managedChildren.push(new constructor(child));
         }
         
         this.show = function () {
-            for (var idx = 0; idx<managedChildren.length; idx++){
+            for (var idx = 0; idx < managedChildren.length; idx++) {
                 managedChildren[idx].show();
             }
         }; 
         
         this.hide = function () {
-            for (var idx = 0; idx<managedChildren.length; idx++){
+            for (var idx = 0; idx < managedChildren.length; idx++) {
                 managedChildren[idx].hide();
             }
         };
--- a/problem_builder/public/js/lms_util.js
+++ b/problem_builder/public/js/lms_util.js
@@ -1,0 +1,122 @@
+(function () {
+
+    /**
+     *  Manager for HTML XBlocks. These blocks are hid by detaching and shown
+     *  by re-attaching them to the DOM.This is only generic way to generically
+     *  handle things like video players (they should stop playing when removed from DOM).
+     *
+     *  @param html an html xblock
+     */
+    function HtmlManager(html) {
+        var $element = $(html.element);
+        var $anchor = $("<span>").addClass("sb-video-anchor").insertBefore($element);
+        this.show = function () {
+            $element.insertAfter($anchor);
+        };
+        this.hide = function () {
+            $element.detach()
+        };
+    }
+    
+    /**
+     *
+     * Manager for HTML Video child. Videos are paused when hiding them, and re-sized when showing them.
+     * @param video an video xblock
+     *
+     */
+    function VideoManager(video) {
+        this.show = function () {
+            if (typeof video.resizer === 'undefined'){
+                // This one is tricky: but it looks like resizer is undefined only if the video is on the
+                // step that is initially visible (and then no resizing is necessary)
+                return;
+            }
+            video.resizer.align();
+        };
+        this.hide = function () {
+            if (typeof video.videoPlayer === 'undefined'){
+                // If video player is undefined this means that the video isn't loaded yet, so no need to
+                // pause it.
+                return;
+            }
+            video.videoPlayer.pause();
+        };
+    }
+    
+    /**
+     *  Manager for Plot Xblocks. Handles updating a plot before displaying it.
+     * @param plot
+     */
+    function PlotManager(plot) {
+        this.show = function () {
+            plot.update();
+        };
+        this.hide = function () {};
+    }
+
+
+    function ChildManager(xblock_element, runtime) {
+
+        var Managers = {
+            'video': VideoManager,
+            'sb-plot': PlotManager
+        };
+
+        var children = runtime.children(xblock_element);
+
+        /**
+         * A list of managers for children than need special care when showing or hiding.
+         *
+         * @type {show, hide}[]
+         */
+        var managedChildren = [];
+
+        /***
+         * This is a workaround for issue where jquery.xblock.Runtime doesn't return HTML Xblocks when querying
+         * for children.
+         *
+         * This can be removed when:
+         *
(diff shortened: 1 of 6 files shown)


problem_builder/public/js/mentoring_with_steps.js
```diff
@@ -17,20 +18,57 @@ function MentoringWithStepsBlock(runtime, element) {
         }
     }
 
-    var activeStep = $('.mentoring', element).data('active-step');
+    var activeStepIndex = $('.mentoring', element).data('active-step');
```
jbzdak: This is kind of leftover from last PR, I left it in on the basis that this is a cleanup.

problem_builder/public/js/lms_util.js
```diff
@@ -0,0 +1,109 @@
+(function () {
```
jbzdak: This can be moved to ``step.js`` but I figured that it would be better to have it on separate file, to keep every file simple enough.

jbzdak:
@itsjeyd This is ready for review.

problem_builder/public/js/lms_util.js
itsjeyd: @jbzdak Is there a reason not to move the contents of this file to the existing `util.js` file? If they do need to be kept in a separate file, I think it would make sense to call it `step_util.js` to indicate the relationship more clearly. (This would also take into account the fact that usage of this file isn't limited to the LMS runtime.)

problem_builder/public/js/lms_util.js
```diff
@@ -0,0 +1,120 @@
...
+
+    /**
+     *  Manager for HTML XBlocks. These blocks are hid by detaching and shown
+     *  by re-attaching them to the DOM.This is only generic way to generically
+     *  handle things like video players (they should stop playing when removed from DOM).
```
itsjeyd: @jbzdak "hid" ⟶ "hidden"

"DOM.This is only generic way to generically handle" ⟶ "DOM. This is the only way to generically handle"

problem_builder/public/js/lms_util.js
```diff
@@ -0,0 +1,120 @@
...
+    
+    /**
+     *
+     * Manager for HTML Video child. Videos are re-sized when showing them.
+     * @param video an video xblock
+     *
+     */
+    function VideoManager(video) {
+        this.show = function () {
+            if (typeof video.resizer === 'undefined'){
```
itsjeyd: @jbzdak Missing space between `)` and `{`. There's a few more instances of this below, so it would make sense to do a search-and-replace.

problem_builder/public/js/lms_util.js
```diff
@@ -0,0 +1,120 @@
...
+                // This one is tricky: but it looks like resizer is undefined only if the video is on the
+                // step that is initially visible (and then no resizing is necessary)
+                return;
+            }
+            video.resizer.align();
+        };
+        /**
+         * Videos should be paused when user leaves, a step containing the video. There is was a proposed implementation
```
itsjeyd: @jbzdak "Videos should be paused when user leaves a step containing a video."

problem_builder/public/js/lms_util.js
```diff
@@ -0,0 +1,120 @@
...
+
+        var Managers = {
+            'video': VideoManager,
+            'sb-plot': PlotManager
+        };
+
+        var children = runtime.children(xblock_element);
+
+        /**
+         * A list of managers for children than need special care when showing or hiding.
```
itsjeyd: @jbzdak "than" ⟶ "that"

problem_builder/public/js/lms_util.js
```diff
@@ -0,0 +1,120 @@
...
+         *
+         * @type {show, hide}[]
+         */
+        var managedChildren = [];
+
+        /***
+         * This is a workaround for issue where jquery.xblock.Runtime doesn't return HTML Xblocks when querying
```
itsjeyd: @jbzdak "Xblocks" ⟶ "blocks"

problem_builder/public/js/lms_util.js
```diff
@@ -0,0 +1,120 @@
...
+         * for children.
+         *
+         * This can be removed when:
+         *
+         * * We allow inclusion of Ooyala blocks inside ProblemBuilder and our clients migrate to Ooyala, in this case
```
itsjeyd: @jbzdak "ProblemBuilder and StepBuilder" or just "StepBuilder" (I'm not sure if there are plans to actually support Ooyala blocks in ProblemBuilder).

problem_builder/public/js/lms_util.js
```diff
@@ -0,0 +1,120 @@
...
+         *   we may drop special handling of HTML blocks. See discussions in OC-1441.
+         * * We include HTML blocks in runtime.children for runtime of jquery.xblock, then just add
+         *   `html: HtmlManager` to `Managers`, and remove this block.
+         */
+        $("div.xblock.xblock-student_view.xmodule_HtmlModule", xblock_element).each(function (idx, element) {
```
itsjeyd: @jbzdak Remove extra space after `function`.

problem_builder/public/js/lms_util.js
```diff
@@ -0,0 +1,120 @@
...
+            managedChildren.push(new HtmlManager({element:element}));
```
itsjeyd: @jbzdak `managedChildren.push(new HtmlManager({ element: element }));`

problem_builder/public/js/lms_util.js
```diff
@@ -0,0 +1,120 @@
...
+        });
+
+        for (var idx = 0; idx < children.length; idx++){
+            var child = children[idx];
+            // NOTE: While following assertion is true for e.g Video blocks:
+            // child.type == $(child.element).data('block-type') it is invalidated by for all sb-* blocks
```
itsjeyd: @jbzdak "it is invalid for all..."

problem_builder/public/js/lms_util.js
```diff
@@ -0,0 +1,120 @@
...
+            var type =  $(child.element).data('block-type');
+            var constructor = Managers[type];
+            if (typeof constructor === 'undefined'){
+                // This block does not requires special care, moving on 
+                continue ;
```
itsjeyd: @jbzdak Remove extra space before `;`.

problem_builder/public/js/lms_util.js
```diff
@@ -0,0 +1,120 @@
...
+            }
+            managedChildren.push(new constructor(child));
+        }
+        
+        this.show = function () {
+            for (var idx = 0; idx<managedChildren.length; idx++){
```
itsjeyd: @jbzdak Can you try and make the spacing in these `for` loops consistent throughout this file? The following version

```js
(var idx = 0; idx < managedChildren.length; idx++)
```

is best for readability I think.

problem_builder/public/js/mentoring_with_steps.js
```diff
@@ -217,11 +251,20 @@ function MentoringWithStepsBlock(runtime, element) {
         reviewButtonDOM.hide();
         tryAgainDOM.show();
 
+        /**
+         * We detach review step from DOM, this is required to handle HTML
+         * blocks that can be added to the Review step.
```
itsjeyd: @jbzdak "HTML blocks with embedded videos" perhaps? Also, it would be good to point out where the detaching happens - this comment had me confused for a second because L261 below doesn't do what it says.

problem_builder/public/js/mentoring_with_steps.js
```diff
@@ -217,11 +251,20 @@ function MentoringWithStepsBlock(runtime, element) {
...
+         *
+         * NOTE: This is handled differently than step js. As the html contents
+         * of review step are replaced with fresh contents in submit function.
```
itsjeyd: @jbzdak "Review steps are handled differently than "normal" steps: The HTML contents of a review step are replaced with ..."

problem_builder/public/js/step.js
```diff
@@ -110,20 +105,15 @@ function MentoringStepBlock(runtime, element) {
...
-                    break;
-                case 'sb-plot':
-                    updatePlot(child);
-                    break;
-                }
-            });
-        }
+        show: function () {
+            $(element).show();
+            childManager.show();
```
itsjeyd: @jbzdak Based on the way `show` and `hide` are used here, it might make sense to call them `showAll` and `hideAll` instead.

problem_builder/tests/integration/test_step_builder.py
```diff
@@ -624,9 +624,7 @@ def test_review_tips(self):
 
     @data(True, False)
     def test_conditional_messages(self, include_messages):
-        """
-        Test that conditional messages in the review step are visible or not, as appropriate.
-        """
+        # Test that conditional messages in the review step are visible or not, as appropriate.
         max_attempts = 3
         extended_feedback = False
         params = {
```
itsjeyd: @jbzdak Do you see a way to verify that videos stop playing by means of an integration test?

itsjeyd:
@jbzdak Reviewed! Please ping me when you're done addressing the comments so I can have another quick look.

problem_builder/public/js/lms_util.js
jbzdak: @itsjeyd  `utils.js` is used only in studio (or `cms`), and this one is used only in `lms` and while this would have negligible effect on performance, i decided that adding unneeded code would be bad, so I decided to split them. 

I'll rename them `step_utils.js`.

problem_builder/tests/integration/test_step_builder.py
jbzdak: Not really, we may check whether HTML blocks are detached. 

However I'd like to have some tests for the Video XBlock functionality (but these are definitely out of scope for now).

problem_builder/public/js/step.js
jbzdak: There is a ``MentoringWithStepsBlock.hideAllSteps`` function so plain ``hideAll`` would be misleading, I decided to rename this one to ``hideStep`` (other option --- like: ``hideAllChildren`` or ``hideStepAndChildren`` --- would be too verbose.)

problem_builder/tests/integration/test_step_builder.py
jbzdak: @bradenmacdonald  

Tim asked me whether it's possible to add regression tests to this feature, and that is not easy for the following two reasons: 

1. Testing whether video is playing is non trivial, especially for videos embedded in HTML. 
2. Testing whether ``Html`` is detached is hard, as the workbench doesn't know how to handle ``html`` XModule, but rather uses ``html_demo`` xblock. For now only a ``html`` Xmodule can be added to step builder using studio interface. As of today I don't know how to fix this issue. 

So would it be sensible to push this without integration tests? 

CC: @itsjeyd

problem_builder/public/js/step.js
```diff
@@ -4,6 +4,8 @@ function MentoringStepBlock(runtime, element) {
 
     var submitXHR, resultsXHR,
         message = $(element).find('.sb-step-message');
+    
+    var childManager = new ProblemBuilderUtilLMS.ChildManager(element, runtime);
```
itsjeyd: @jbzdak Now that the utils file has been renamed, `ProblemBuilderUtilLMS` should be renamed as well I think. How about just `StepUtil`?

jbzdak: Done. I went for ``ProblemBuilderStepUtil`` this is a global name attached to window object, so I'd rather be verbose than have a clash.

itsjeyd:
@jbzdak :+1: once the build is green and @bradenmacdonald confirms that we can go ahead without adding integration tests for now.

bradenmacdonald:
@jbzdak As long as your code is well commented (looks like it is - thanks!), writing integration tests would require enormous effort (which seems to be the case), and all the other non-video-specific refactoring you've done is covered by tests, that's fine with me in this case.

The only thing I would suggest is that you first have a look to see if you can [use `XBlock.register_temp_plugin`](https://github.com/edx/XBlock/blob/5bd3985/xblock/test/test_parsing.py#L140-L142) to temporarily register a *mock* HTML or Video XBlock which offers a JS pause method, then do an acceptance test with that, to make sure that the pause method is being called. But again if you look into that and it's going to be an enormous amount of work, we can go without. Then we can revisit testing when we implement first-class support for ooyala blocks.

jbzdak:
@bradenmacdonald 

Using XBlock.register_temp_plugin would work, [if not for this](https://github.com/open-craft/problem-builder/pull/115/files#diff-fd31c4ba0e50d2d68fe7be4476736298R83). JQuery-xblock doesnt return Html XModules as children of the step (this runtime is used on Apros), so I added a workaround of querying dom for HTML modules. . . and this workaround would not work in XBlock sdk as the dom elements generated by sdk and lms are different. 

In LMS I can select html xblock using `div.xblock[data-module-type="html"]` and in SDK I'd have to use  `div.xblock-v1[data-module-type="html"]`, of course I could add some more or less ugly hack, to make it work. But I'd suggest rather pushing for inclusion of Ooyala block as a first class citizen, and then removing the "detach html nodes from DOM" logic.

bradenmacdonald:
@jbzdak Sounds good, let's push for that.

jbzdak:
Merging.

PR #116, opened 2016-04-12
Hide MRQ question-level feedback on return.

mtyaka:
When the pb_mcq_hide_previous_answer is set to True, question-level feedback should not be displayed when user returns to the page.

cf. [MCKIN-3999](https://edx-wiki.atlassian.net/browse/MCKIN-3999)

*Test Instructions*

1. Create a new subunit containing one problem builder MRQ question unit, and another arbitrary unit. Make sure to put some content into the "Message (General feedback provided when submitting)" field of the MRQ.
2. Enable `pb_mcq_hide_previous_answer` in your `XBLOCK_SETTINGS`.
3. Go to the unit with the MRQ question, fill it in, and submit. You should see the general feedback message.
4. Go to the next unit, then come back to the unit with the MRQ question.
5. General feedback message should be hidden.

### DIFF
--- a/problem_builder/tests/integration/test_mentoring.py
+++ b/problem_builder/tests/integration/test_mentoring.py
@@ -442,26 +442,87 @@ class ProblemBuilderQuestionnaireBlockTest(ProblemBuilderBaseTest):
 
     @ddt.unpack
     @ddt.data(
-        # MCQ with tips
-        ("feedback_persistence_mcq_tips.xml", '.choice-tips'),
-        # Like the above but instead of tips in MCQ
+        # Questionnaire with tips.
+        ("feedback_persistence_mcq_tips.xml", 'MCQ', True, False),
+        ("feedback_persistence_mrq_tips.xml", 'MRQ', True, False),
+        # Like the above but instead of tips in MCQ/MRQ
         # has a question level feedback. This feedback should also be suppressed.
-        ("feedback_persistence_mcq_no_tips.xml", '.feedback')
+        ("feedback_persistence_mcq_general_feedback.xml", 'MCQ', False, True),
+        ("feedback_persistence_mrq_general_feedback.xml", 'MRQ', False, True),
+        # These examples have both the choice tips and the general feedback.
+        ("feedback_persistence_mcq_general_feedback_and_tips.xml", 'MCQ', True, True),
+        ("feedback_persistence_mrq_general_feedback_and_tips.xml", 'MRQ', True, True),
+        # And these have neither the tips nor the general feedback.
+        ("feedback_persistence_mcq_no_feedback.xml", 'MCQ', False, False),
+        ("feedback_persistence_mrq_no_feedback.xml", 'MRQ', False, False),
     )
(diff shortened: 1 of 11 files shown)


problem_builder/tests/integration/test_mentoring.py
```diff
@@ -442,22 +442,32 @@ def assert_state(answer, mcq, mrq, rating, messages):
...
-        # MCQ with tips
+        # Questionnaire with tips
         ("feedback_persistence_mcq_tips.xml", '.choice-tips'),
-        # Like the above but instead of tips in MCQ
+        ("feedback_persistence_mrq_tips.xml", '.choice-tips'),
+        # Like the above but instead of tips in MCQ/MRQ
         # has a question level feedback. This feedback should also be suppressed.
-        ("feedback_persistence_mcq_no_tips.xml", '.feedback')
+        ("feedback_persistence_mcq_no_tips.xml", '.feedback'),
+        ("feedback_persistence_mrq_no_tips.xml", '.feedback'),
```
itsjeyd: @mtyaka I think it would be worth extending this test to also cover cases where an MCQ/MRQ has:

* no question-level feedback and no choice-level feedback (no message displayed when clicking checkmark for individual choice after submitting answer)
* both question-level feedback and choice-level (general message displayed after submitting, choice-level message displayed after clicking corresponding checkmark)

Aside from making the code even more robust against regressions, a test that covers all combinations could also serve to document the expected behavior for us.

itsjeyd:
@mtyaka Works for me! Tested on both LMS and Apros (thanks for the course export).

Aside from the comment I left about the integration tests, maybe we should add a comment to the code that explains that `pb_mcq_hide_previous_answer` also controls previous answers for MRQs. [This line](https://github.com/open-craft/problem-builder/blob/hide-mrq-feedback/problem_builder/mentoring.py#L63) would be a good place for such a comment I think.

mtyaka:
@itsjeyd Thanks for the good suggestions! I implemented the changes, please take another look.

problem_builder/tests/integration/xml_templates/feedback_persistence_mcq_general_feedback.xml
```diff
@@ -0,0 +1,11 @@
+<vertical_demo>
+    <problem-builder url_name="mcq_general_feedback" enforce_dependency="false">
```
itsjeyd: @mtyaka Nit: Can we do without `enforce_dependency="false"` in the templates that are used by `test_feedback_persistence_tips`? It seems to be redundant since `enforce_dependency` has a default value of `False`.

itsjeyd:
@mtyaka Looking great, thanks for the updates! I had one more nit (sorry for not noticing sooner), but other than that this is :+1:

problem_builder/tests/integration/xml_templates/feedback_persistence_mcq_general_feedback.xml
mtyaka: Thanks @itsjeyd, I removed `enforce_dependency`. It was redundant indeed.

PR #117, opened 2016-07-11
Fix mako template context error

bradenmacdonald:
This fixes https://openedx.atlassian.net/browse/TNL-4901 by backporting the already-existing fix that was on the `master` branch.

The problem was that `mentoring_view` was inadvertently modifying its `context` parameter.

Note 1: @adampalay tested and confirmed this fixes the TNL-4901 issue.

Note 2: on this branch, CircleCI is expected to fail (it's only configured on master), but TravisCI is required to pass - so the tests are fine.

### DIFF
--- a/problem_builder/answer.py
+++ b/problem_builder/answer.py
@@ -164,7 +164,7 @@ class AnswerBlock(AnswerMixin, StepMixin, StudioEditableXBlockMixin, XBlock):
 
     def mentoring_view(self, context=None):
         """ Render this XBlock within a mentoring block. """
-        context = context or {}
+        context = context.copy() if context else {}
         context['self'] = self
         context['hide_header'] = context.get('hide_header', False) or not self.show_title
         html = loader.render_template('templates/html/answer_editable.html', context)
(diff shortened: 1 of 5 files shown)


adampalay:
@bradenmacdonald 👍

problem_builder/answer.py
```diff
@@ -164,7 +164,7 @@ def student_input(self):
 
     def mentoring_view(self, context=None):
         """ Render this XBlock within a mentoring block. """
-        context = context or {}
+        context = context.copy() if context else {}
         context['self'] = self
```
adampalay: I trust you that this fixes the issue, but I'm confused by why having `"self"` as a parameter in `context` doesn't mess up mako's `render` function

bradenmacdonald: This XBlock uses Django to render its templates, not mako. The problem only occurs if this block unintentionally modifies the context dict that was passed as an argument, so that the 'self' gets added to a different scope/context entirely (e.g. it gets added to the context of the parent block).

bradenmacdonald: Example:

```python
>>> from django.template import Context, Template
>>> Template(u"hello").render(Context({'self': 'present'}))
u'hello'
```

haikuginger:
@bradenmacdonald 👍 Verified that with this change pulled in, the `edx-release` branch seems to run fine; I was able to create and work on a problem using this XBlock.

PR #120, opened 2016-09-12
Update xblock-utils to v1.0.3.

itsjeyd:
New version includes fixes from https://github.com/edx/xblock-utils/pull/38 and https://github.com/edx/xblock-utils/pull/39.

Follow-up PR for [OC-1819](https://tasks.opencraft.com/browse/OC-1819) (candidate task).

Also fixes an issue affecting dependency installation for CircleCI builds for problem-builder. See the commit message of https://github.com/open-craft/problem-builder/pull/120/commits/51277a34fb426724616618c1afdb893ab2de4c6b for details.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -4,6 +4,8 @@ machine:
 dependencies:
   override:
     - "pip install -U pip wheel"
+    # Temporarily pin setuptools to a specific version.
+    # See commit message of https://github.com/open-craft/problem-builder/commit/51277a34fb426724616618c1afdb893ab2de4c6b for more info:
     - "pip install setuptools==24.3.1"
     - "pip install -e git://github.com/edx/xblock-sdk.git@bddf9f4a2c6e4df28a411c8f632cc2250170ae9d#egg=xblock-sdk"
     - "pip install -r requirements.txt"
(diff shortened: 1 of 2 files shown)


circle.yml
```diff
@@ -3,7 +3,8 @@ machine:
     version: 2.7.10
 dependencies:
   override:
-    - "pip install -U pip wheel setuptools"
+    - "pip install -U pip wheel"
+    - "pip install setuptools==24.3.1"
```
bradenmacdonald: Nit: `yml` files support comments; can you please add a link here to https://github.com/open-craft/problem-builder/commit/51277a34fb426724616618c1afdb893ab2de4c6b in case people are wondering why this is pinned and don't do a git blame?

bradenmacdonald:
Hmm, that setuptools bug looks annoying. Thanks for these fixes, Tim.

:+1: if you add the requested comment.

- [x] I tested this: Installed this on devstack, did `pip install -r requirements.txt`, then started Studio with `NO_PREREQ_INSTALL=1 paver devstack studio --fast` and confirmed it worked.
- [x] I read through the code
- [x] Includes documentation: Nice explanation in commit message; just needs a link.

itsjeyd:
Thanks @bradenmacdonald! I addressed your nit. Merging now.

PR #121, opened 2016-09-16
Refactor MentoringBlock to use StudioContainerWithNestedXBlocksMixin

haikuginger:
Since we initially created ProblemBuilder, the StudioContainerWithNestedXBlocksMixin class has been added to the XBlock runtime. Since this is now the standardized way of dealing with nested XBlocks, we want to move MentoringBlock to use that structure and gain any existing benefits, as well as any benefits for the future.

**Dependencies**: None

**Screenshots**:
screen shot 2016-09-28 at 7 38 21 am

screen shot 2016-09-28 at 7 38 10 am

**Sandbox URL**:
- LMS : http://mentoring-block-refactor.opencraft.hosting
- Studio: http://studio-mentoring-block-refactor.opencraft.hosting

**Partner information**: Hosted on edx.org (Davidson College)

**Testing instructions**:

1. In either the sandbox or a devstack, verify that you are able to create a mentoring block.
2. Verify that subelements of the mentoring block can be created as expected.
3. Verify that you are able to drag and drop subelements to reorder them (not possible previously)

1. To prevent a circular import, we had to revert a performance enhancement change that imported MentoringBlock directly into the Questionnaire module. Because the QuestionnaireBlock is now used directly by the MentoringBlock, this resulted in a loop. Suggestions are welcome for a way we can keep the performance enhancement.

**Reviewers**
- [ ] @mtyaka 
- [ ] edX reviewer[s] TBD


problem_builder/mentoring.py
```diff
@@ -817,13 +858,9 @@ def author_edit_view(self, context):
         """
         Add some HTML to the author view that allows authors to add child blocks.
         """
-        fragment = Fragment(u'<div class="mentoring">')  # This DIV is needed for CSS to apply to the previews
-        self.render_children(context, fragment, can_reorder=True, can_add=False)
-        fragment.add_content(u'</div>')
-        fragment.add_content(loader.render_template('templates/html/mentoring_add_buttons.html', {}))
-        fragment.add_content(loader.render_template('templates/html/mentoring_url_name.html', {
```
mtyaka: Nit: I'm not sure how important this is, but I think we can put the URL name back in to make it consistent with Step Builder block (see screenshot).

screen shot 2016-09-29 at 10 08 54

problem_builder/mentoring.py
```diff
@@ -817,13 +858,9 @@ def author_edit_view(self, context):
         """
         Add some HTML to the author view that allows authors to add child blocks.
         """
-        fragment = Fragment(u'<div class="mentoring">')  # This DIV is needed for CSS to apply to the previews
-        self.render_children(context, fragment, can_reorder=True, can_add=False)
-        fragment.add_content(u'</div>')
-        fragment.add_content(loader.render_template('templates/html/mentoring_add_buttons.html', {}))
```
mtyaka: We should remove the `mentoring_add_buttons.html` template if we're not using it anymore.

problem_builder/mentoring.py
```diff
@@ -322,6 +329,40 @@ class MentoringBlock(BaseMentoringBlock, StudioContainerXBlockMixin, StepParentM
...
+                ImageModal, category='imagemodal', label=_(u"Image Modal")
+            ))
+        except ImportError:
+            pass
+
+        return [
+            NestedXBlockSpec(AnswerBlock, boilerplate='studio_default'),
+            MCQBlock, RatingBlock, MRQBlock,
+            NestedXBlockSpec(None, category="html", label=self._("HTML")),
+            AnswerRecapBlock, MentoringTableBlock, PlotBlock, SliderBlock
```
mtyaka: It looks like there are three types of children missing from this list: "Message (Complete)", "Message (Incomplete)", and "Message (Max # Attempts)" (see screenshot). It's not visible in the screenshot, but I think there is also "Message (Assessment Review)" when using the deprecated assessment mode.

screen shot 2016-09-29 at 10 04 22

problem_builder/questionnaire.py
```diff
@@ -93,13 +92,8 @@ def student_view(self, context=None):
         context['hide_header'] = context.get('hide_header', False) or not self.show_title
 
         fragment = Fragment(loader.render_template(template_path, context))
-        # If we use local_resource_url(self, ...) the runtime may insert many identical copies
-        # of questionnaire.[css/js] into the DOM. So we use the mentoring block here if possible
-        block_with_resources = self.get_parent()
-        if not isinstance(block_with_resources, MentoringBlock):
-            block_with_resources = self
```
mtyaka: We normally work around circular dependencies by deferring imports until they are needed, for example: https://github.com/open-craft/problem-builder/blob/4ffa21c36b534f9b86f50b128815800a33b69899/problem_builder/mentoring.py#L418

Would that work here? If it works, it would be great if we could apply the same optimization to Step Builder.

problem_builder/mentoring.py
mtyaka: Now that were' using `StudioContainerWithNestedXBlocksMixin`, do we still need the code in this file: https://github.com/open-craft/problem-builder/blob/4ffa21c36b534f9b86f50b128815800a33b69899/problem_builder/public/js/mentoring_edit.js or is that handled by the mixin now? It might be worth checking and remove that file if it's not needed anymore.

haikuginger: It looks like we're still using it in the questionnaire block; I haven't taken a long look to determine what it's being used for, though.

haikuginger:
@mtyaka, I'm updating the sandbox now - care to take another look?

problem_builder/mentoring.py
```diff
@@ -355,12 +355,27 @@ def allowed_nested_blocks(self):
         except ImportError:
             pass
 
+        message_block_shims = [
+            NestedXBlockSpec(
+                MentoringMessageBlock,
+                category='pb-message',
+                boilerplate=message_type,
+                label=get_message_label(message_type),
```
mtyaka: The Message block specs should have `single_instance=True`, because we only allow a single instance of each message type inside a Mentoring block, *BUT* I tried that and it looks like `single_instance` functionality is currently [implemented in a rather hackish way](https://github.com/edx/xblock-utils/blob/7e7e6fc20e2de3ecb0c882db2bde45b5e4d1ccbb/xblockutils/public/studio_container.js#L40-L48) and it can only limit blocks by category. Here we still want to allow multiple `pb-message` blocks, as long as they are of different `message_type`, but it doesn't look like we can do that currently.

If you set `single_instance=True`, the button that you click gets correctly disabled when adding a Message type block to a mentoring container, but if you refresh the page, *all* `pb-message` buttons are disabled rather than just the one type that you added to the container.

One way to fix this would be to introduce subclasses of `MentoringMessageBlock`, so that each message type would be a separate class.
Another way would be to expand the `single_instance` feature of nested xblocks in xblock-utils so that it would be able to toggle the button on parameters other than just the block category.

I think I prefer the second option, but neither of these options are simple, so in order to not block this PR, I think I am fine with leaving `single_instance=True` out for now, and create a task to handle this in next sprint.

haikuginger:
@mtyaka, those changes have been made.

mtyaka:
Thanks @haikuginger! Once the build is green:

:+1:

- [x] I tested this: Adding child items to the mentoring block container works correctly after refactoring.
- [x] I read through the code
- [x] ~~Includes documentation~~ (this is internal refactoring, no documentation needed)

problem_builder/mentoring.py
antoviaque: @mtyaka Given the larger work that both options would represent, which we don't have time to address in the scope here, could we try to find a third option?

What would happen if two messages of the same type were added? Would there be an explicit validation error appearing for the bloc? If so, would it make sense to simply let the validation error inform the author that he needs to fix it - we can't prevent every editing issue proactively anyway, so this is already a pattern the author should be familiar with. And improve this case when Studio and xblock-utils support the use case?

mtyaka: @antoviaque If two messages of the same type are added, there is an explicit error message on the parent (mentoring) block, although it's only visible after you update the page. It's not a big problem, so I think it would be fine to leave it for when Studio/xblock-utils support the use case.

haikuginger: Okay, sounds like I should just go ahead and merge? @antoviaque, @mtyaka?

PR #124, opened 2016-10-04
Handle unresolvable anonymous ids gracefully.

smarnach:
In some cases it is possible that the anonymous `student_id` of a submission can't be reversed to the original user.  In this case, the function `user_by_anonymous_id()` returns `None`, resulting in a crash when building reports.

This change fixes the crash by returning `student_id` instead of the actual user name in case it cannot be reconstructed.


mtyaka:
👍

PR #131, opened 2016-11-23
Introduce Answer.course_key and deprecate course_id.

mtyaka:
This patch introduces a new 255-char `course_key` field to the `Answer` model. The old 50-char `course_id` field is still present, but is deprecated and will be removed in next release.

It removes the previous migration (which was never deployed to production) which extended the existing `course_id` field to 255 characters, which was found to not be the optimal approach on the edx.org production database due to the relevant table's size. We cannot simply extend the existing `course_id` field because we have some large `problem_builder_answer` tables in production and the migration to extend the column would lock the table causing issues in production.

The code is updated so that it uses the new `course_key` column, but falls back to `course_id`.
In next release, a data migration to copy course_id data to `course_key` will be provided and the `course_id` column dropped.

Below is the output of `sqlmigrate` for the new migration:

```
BEGIN;
ALTER TABLE `problem_builder_answer` ADD COLUMN `course_key` varchar(255) NULL;
ALTER TABLE `problem_builder_answer` ADD CONSTRAINT `problem_builder_answer_student_id_2ce682a818c95cbc_uniq` UNIQUE (`student_id`, `course_key`, `name`);
CREATE INDEX `problem_builder_answer_c8235886` ON `problem_builder_answer` (`course_key`);

COMMIT;
```

Next release will add a data migration to copy contents of `course_id` to the new `course_key` column. EdX will fake this migration and copy data in batches instead.
Next release will also drop the deprecated `course_id` column.

See https://github.com/edx/edx-platform/pull/14013#issuecomment-261635454 for more information.

**Environments**:  edx.org and edge.edx.org

**Merge deadline**: ASAP - this is blocking a course on Edge

**JIRA tickets**: [TNL-5932](https://openedx.atlassian.net/browse/TNL-5932)

**Discussions**: See https://github.com/edx/edx-platform/pull/14013 and JIRA ticket.

**Sandbox URL**:

- LMS: http://pr14044.sandbox.opencraft.hosting/
- Studio: http://studio-pr14044.sandbox.opencraft.hosting/

**Partner information**: hosted on edX Edge (Davidson College)

**Testing instructions**:

Check that you can successfully add a Problem Builder "Long Answer" component to a course with a course key that is longer than 50 character and that answers answered on an older version of problem-builder continue to work correctly with this patch.

1. Install problem-builder v2.6.1
1. Create a course with a combined course key less than 50 characters.
1. Add a Problem Builder component, then add the Long Answer component.
1. Answered several questions.
1. Create a course with a combined course key *more* than 50 characters.
1. Add Unit with Problem Builder and a Long Answer element - watched it fail.
1. Installed problem-builder from this PR.
1. Run migrations.
1. Verify that the answers are still present in the first course.
1. Verify that the Long Answer element in the course with the very long course ID is working now, and you can submit answers to it.

**Reviewers**
- [x] @pomegranited

### DIFF
--- a/problem_builder/answer.py
+++ b/problem_builder/answer.py
@@ -23,6 +23,10 @@
 import logging
 from lazy import lazy
 
+from django.core.exceptions import ValidationError
+from django.db import IntegrityError
+from django.utils.crypto import get_random_string
+
 from .models import Answer
 
 from xblock.core import XBlock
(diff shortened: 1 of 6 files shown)


problem_builder/models.py
```diff
@@ -35,11 +35,19 @@ class Answer(models.Model):
...
 
     name = models.CharField(max_length=50, db_index=True)
     student_id = models.CharField(max_length=32, db_index=True)
-    course_id = models.CharField(max_length=255, db_index=True)
+    # course_id is deprecated; it will be removed in next release.
+    course_id = models.CharField(max_length=50, db_index=True)
+    # course_key is the new course_id replacement with extended max_length.
+    # We need to allow NULL values during the transition period,
+    # but will be set to null=False in next release.
+    course_key = models.CharField(max_length=255, db_index=True, null=True)
```
mtyaka: No, multiple `NULL` values are not considered to be identical to each other so they don't violate the uniqueness constraint.

mtyaka:
Rebased from f0d0ce517b756c892c4eff424c260063af2a9572.

mtyaka:
Thank you @pomegranited!

problem_builder/answer.py
```diff
@@ -64,6 +68,40 @@ def _get_student_id(self):
         except AttributeError:
             return self.scope_ids.user_id
 
+    @staticmethod
+    def _fetch_model_object(name, student_id, course_id):
+        sql_query = '''SELECT * FROM problem_builder_answer
+                         WHERE name = %s
+                           AND student_id = %s
+                           AND (course_key = %s
+                                OR course_id = %s)'''
```
mtyaka: @maxrothman Only because I was not familiar with django `Q` objects and didn't know django ORM supports `OR` queries. Thanks for the hint, I opened #136 to convert the raw SQL query to `Q` objects - can you please review?

If it looks good, I'll update the version number at https://github.com/edx/edx-platform/pull/14044 and we can finally get that one merged.

PR #132, opened 2016-11-23
Implement Completion XBlock

itsjeyd:
This PR adds a new question type to Problem Builder that allows course authors to ask learners whether they completed an offline activity.

cf. [OC-2086](https://tasks.opencraft.com/browse/OC-2086)

**Screenshots**

- LMS: completion-lms-full

- Apros: completion-apros-full

**Test instructions** (to be completed on Solutions devstack)

1. Create a new section and subsection.

2. Add two units to the subsection created in the previous step.

3. Add a Problem Builder component to the first unit.

4. Add a Completion block to the Problem Builder component and publish the unit.

5. In the LMS/Apros, navigate to the unit containing the Problem Builder component. Observe that on first load, the checkbox to indicate completion is unchecked.

6. Tick the checkbox.

7. Navigate to the second unit, then navigate back to the first unit. Observe that the checkbox reverts to the default value (unchecked).

8. Tick the checkbox.

9. Reload the page. Observe that the checkbox reverts to the default value (unchecked).

10. Tick the checkbox and click "Submit".

11. Navigate to the second unit, then navigate back to the first unit. Observe that the checkbox retained the submitted value.

12. Reload the page. Observe that the checkbox retained the submitted value.

13. Add a second completion block to the first unit. Make sure you can update them independently (i.e., ticking checkbox for first block and clicking "Submit" should not result in checkbox for second block being checked, and vice versa).

**Reviewers**

- [ ] @mtyaka


problem_builder/completion.py
```diff
@@ -0,0 +1,137 @@
...
+    editable_fields = ('display_name', 'show_title', 'question', 'answer')
+
+    def mentoring_view(self, context):
+        """
+        Main view of this block.
+        """
+        context = context.copy() if context else {}
+        context['question'] = self.question
+        context['answer'] = self.answer
+        context['completion_id'] = 'pb-completion-{}'.format(uuid.uuid4().hex[:20])
```
mtyaka: nit: The `completion_id` value is used for the HTML `id` attribute of the checkbox, but it doesn't look like we need the `id` attribute at all? I think we could just remove it (and the line above).

itsjeyd:
Thanks for the review, @mtyaka! I addressed your comment in ba5071419205cd7ac22d9f27bdcfdec8f134ff7a, and will merge once I get a green build. Version bump PR against the Solutions fork forthcoming.

itsjeyd:
Rebased from ba5071419205cd7ac22d9f27bdcfdec8f134ff7a (squashed review changes into appropriate commit).

itsjeyd:
Forgot to remove unused import in ba5071419205cd7ac22d9f27bdcfdec8f134ff7a. Fixed in e32a341f1fa64edbd5a75f4ea230e351eb5ba83b.

itsjeyd:
Rebased from e32a341f1fa64edbd5a75f4ea230e351eb5ba83b (squashed additional changes into appropriate commit).

bradenmacdonald:
> Documentation is not included, but other sub-blocks such as Plot and Slider are not documented either, so I'm not sure documentation is required here.

Yes, documentation is always expected, hence this review step :-p
I would have liked to see something added to https://github.com/open-craft/problem-builder/blob/master/doc/Questions.md

PR #133, opened 2016-11-24
Remove migration that extends course_id column

mtyaka:
Migration locks the table, which is not desired since the production table on edx.org is quite large.

The course_id field will be extended in a multiple steps instead.

For more info see:

- https://github.com/edx/edx-platform/pull/14013
- https://github.com/open-craft/problem-builder/pull/131
- https://github.com/edx-solutions/edx-platform/pull/757

**Reviewers**:

- [ ] @itsjeyd


mtyaka:
Thanks @itsjeyd, I'll merge once the build succeeds

PR #138, opened 2017-01-12
Copy course_id into course_key and remove temporary transition code

mtyaka:
This is a follow-up to #131. In #131, we deprecated the `Answer.course_id` field and replaced it with a new `Answer.course_key` field with extended length to accommodate courses with course keys longer than 50 characters (for example some CCX courses on edx.org have course keys longer than 50 characters).

 This PR includes a migration that copies contents of `course_id` column into `course_key`, for any `Answer` object where `course_key` is `NULL` (these are `Answer` objects that were created before problem-builder v2.6.5 was deployed).

**NOTE: Before deploying v2.6.6 to production, edX will have to fake this migration and manually copy `course_id` values into `course_key` in batches because the production `problem_builder_answer` table is large and performing the migration in one go would degrade performance!**

Next, this PR also includes a migration that makes the `course_key` mandatory and removes transition code that supports both `course_id` and `course_key` columns.

The `course_id` column is only deprecated at this point, but could safely be dropped in the future.

**Environments**:  edx.org and edge.edx.org

**JIRA tickets**: [TNL-5932](https://openedx.atlassian.net/browse/TNL-5932)

**Discussions**: See https://github.com/edx/edx-platform/pull/14013 and JIRA ticket.

**Sandbox URLs**:

Built with c733391cee6c6e34d2b3b74aa95a6e908132b138.

- LMS: http://pr14327.sandbox.opencraft.hosting/
- Studio: http://studio-pr14327.sandbox.opencraft.hosting/

**Partner information**: hosted on edX Edge (Davidson College)

**Testing instructions**:

This patch should not change functionality in any way. Verify that long Answer objects still work correctly on the sandbox.

**Reviewers**
- [x] OpenCraft:  @bdero 
- [x] @maxrothman
- [x] @adampalay

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -11,7 +11,7 @@ dependencies:
     - "pip install -r requirements.txt"
     - "pip install -r $VIRTUAL_ENV/src/xblock-sdk/requirements/base.txt"
     - "pip install -r $VIRTUAL_ENV/src/xblock-sdk/requirements/test.txt"
-    - "pip uninstall -y xblock-problem-builder && python setup.py sdist && pip install dist/xblock-problem-builder-2.6.5.tar.gz"
+    - "pip uninstall -y xblock-problem-builder && python setup.py sdist && pip install dist/xblock-problem-builder-2.6.6.tar.gz"
     - "pip install -r test_requirements.txt"
     - "mkdir var"
 test:
--- a/problem_builder/answer.py
+++ b/problem_builder/answer.py
@@ -23,11 +23,6 @@
 import logging
 from lazy import lazy
 
-from django.core.exceptions import ValidationError
-from django.db import IntegrityError
-from django.db.models import Q
-from django.utils.crypto import get_random_string
-
 from .models import Answer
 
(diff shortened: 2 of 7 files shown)


bdero:
Looks good @mtyaka. :) :+1:

- [x] I tested this: Long answers are still working as expected with course names > 50 characters (on the sandbox).
- [x] I read through the code
- [ ] ~~I checked for accessibility issues~~ There are no user facing changes.
- [ ] ~~Includes documentation~~ Not necessary - this PR is a code base touch up that eliminates tech debt. It doesn't change anything external facing.

mtyaka:
@edx/devops Can somebody please review this follow-up to the hotfix we did for problem-builder. It removes the temporary transition code and includes a migration that copies values from the deprecated column into the new, extended `course_key` column - **this migration will have to be faked and performed in batches when deploying to edx.org**!.

/cc @adampalay

adampalay:
code looks ok to me! But this should definitely get ops approval :)

PR #125, opened 2016-10-05
Accessibility updates

pomegranited:
Addresses the following accessibility issues:

* Encloses all `<input>` and `<textarea>` fields inside a `<label>`, as per edx-platfomr's CAPA problem layout.
* Associates question and feedback text with the corresponding choice labels, matching edx-platform's CAPA problem layout.
* Ensures that all `<input>` elements have a `<label>`.
* Ensures all table header cells use `<th>` elements.
* Associates feedback icons and choice-level feedback messages with the corresponding choice.
* Adds `aria-live="polite"` attributes to the blocks whose content changes without page reload.

**Partner information**: affects edX-hosted clients who use this XBlock.

**Testing instructions**:

1. Install this PR's branch of the problem-builder xblock on your devstack.  (Ensure that any other versions are removed.)
1. Create a course, and import [course.pr125.tar.gz](https://github.com/open-craft/problem-builder/files/513109/course.pr125.tar.gz).
1. Visit the Step Builder subsection to see the Plot Preview table header changes.
1. Login to the LMS as staff to see the Instructor Tool subsection.  Ensure that your screen reader is alerted to the dynamic page content changes as you change the filters.
1. Register for the course as a student, and browse through to test:
    * All exercises can be completed using your keyboard.
    * All status icons have a text label.
    * All input items have a label.
    * After clicking "Submit" on Problem Builder exercises, or "Next" on Step Builder, your screen reader is alerted to the dynamic page content changes.

**Author Notes and Concerns**

1. The test course also includes a CAPA section, which contains a checkbox problem with hints, and a multiple choice problem with hints, if you want to compare against their accessibility layout.
1. Internationalization text changes will be addressed by OC-1856.
1. The Plot XBlock needs substantial work to ensure that the plot data is understandable to non-sighted learners. 

**Reviewers**
- [ ] @mtyaka

PR #139, opened 2017-01-24
Fix display of MCQ/MRQ choices with long text.

itsjeyd:
Reported by McKinsey.

cf. [OC-2264](https://tasks.opencraft.com/browse/OC-2264)

**Screenshots** (Apros)

MCQ:

mcq

MCQ during review (with tip):

mcq-review-tip

Rating question:

rating

Rating question during review (without tip):

rating-review

MRQ:

mrq

MRQ during review (without tip):

mrq-review

**Test instructions**

1. Add a Step Builder block to a unit, setting "Max. attempts allowed" to a non-zero value, and "Extended Feedback" to True.

2. Add a Mentoring Step block and a Review Step block to the Step Builder block.

3. Add an MCQ, Rating, and MRQ to the Mentoring Step block, setting the text of some choices to a long string.

4. Add tips for individual choices to the MCQ.

5. Add a Score Summary block to the Review Step block.

6. Publish the unit and navigate to it in the LMS/Apros. Observe that alignment of choice texts matches screenshots.

7. Answer the questions, submit, and review grade until you have no attempts left. Use links to individual questions to review answers. Observe that alignment of choice texts matches screenshots.

**Reviewers**

- [ ] @mtyaka

### DIFF
--- a/problem_builder/public/css/questionnaire.css
+++ b/problem_builder/public/css/questionnaire.css
@@ -2,17 +2,21 @@
     display: table;
     position: relative;
     width: 100%;
-    border-spacing: 0 6px;
+    border-spacing: 0 2px;
     padding-top: 10px;
     margin-bottom: 10px;
 }
 
 .mentoring .questionnaire .choice-result {
-    display: inline-block;
+    display: table-cell;
     width: 34px;
     vertical-align: top;
     cursor: pointer;
     float: none;
+}
+
+.mentoring div[data-block-type=pb-mrq] .questionnaire .choice-result {
+    display: inline-block;
 }
 
 .mentoring .questionnaire .choice {
@@ -78,7 +82,12 @@
 }
 
 .mentoring .choices-list .choice-selector {
-    display: inline-block;
+    display: table-cell;
+}
+
+.mentoring .choices-list .choice-label-text {
+    display: table-cell;
+    padding-left: 0.4em;
 }
 
 .mentoring .choice-tips-container,
--- a/problem_builder/public/css/questionnaire.css
+++ b/problem_builder/public/css/questionnaire.css
@@ -88,6 +88,7 @@
 .mentoring .choices-list .choice-label-text {
     display: table-cell;
     padding-left: 0.4em;
+    padding-right: 0.4em;
 }
 
 .mentoring .choice-tips-container,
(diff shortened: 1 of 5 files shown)


bradenmacdonald:
@itsjeyd Looks much better - I'm just wondering; did the MCQ question text used to be full width or have they always been half-width like this?

itsjeyd:
@bradenmacdonald I'm not sure if they've always *been* half-width, but they have been in the more recent past. This PR does not change that particular aspect of the layout.

problem_builder/public/css/questionnaire.css
```diff
@@ -78,7 +82,12 @@
 }
 
 .mentoring .choices-list .choice-selector {
-    display: inline-block;
+    display: table-cell;
+}
+
+.mentoring .choices-list .choice-label-text {
+    display: table-cell;
+    padding-left: 0.4em;
```
mtyaka: We should add some padding to the right as well - currently there's no padding between the choice text and the line that divides the text from the tip:

screen shot 2017-01-27 at 09 51 01

Ideally it would be the same amount of padding on both sides, to make it easier to support RTL.

mtyaka:
@itsjeyd While testing this, I found that the layout is currently somewhat broken in RTL mode, but it's also broken on master, so I'm just including the screenshots here for reference, but fixing this is out of scope of this PR. I don't really remember whether problem-builder was ever RTL compatible.

screen shot 2017-01-27 at 09 53 37


screen shot 2017-01-27 at 09 54 36

itsjeyd:
@mtyaka Thanks for the review, and for catching the padding issue. I was using the Apros theme, which made it difficult/impossible to notice. Fixed in the most recent commit, please have another look.

padding-lms

mtyaka:
👍 thank you

bradenmacdonald:
> @bradenmacdonald I'm not sure if they've always been half-width, but they have been in the more recent past. This PR does not change that particular aspect of the layout.

@itsjeyd As I feared, QA reviewed this and said the half-width MCQ text is not correct.

Form further investigation I think it comes from #125. @pomegranited was this change deliberately made for a11y or is this a side effect?

@itsjeyd  Can we add a CSS rule to the Apros CSS to override that?

pomegranited:
@bradenmacdonald  CC @itsjeyd  IIRC, the [choices width was adjusted](https://github.com/open-craft/problem-builder/pull/125/files#diff-89b628fea6b5e792171bdc91e6f8bc70R88) simply to make tips appear alongside the choices, not as a specific a11y change.

PR #140, opened 2017-01-31
Do not show completion checkmark when hide_feedback is enabled

mtyaka:
This makes the Completion block consistent with the Answer block.

When 'pb_hide_feedback_if_attempts_remain' is set to True in
XBLOCK_SETTINGS, once the user submits an answer to an Answer (freeform
long answer), the green checkmark is not visible after refreshing the
page, even though the given answer/text is still visible.

This commit makes Completion blocks consistent with that behaviour.

**JIRA ticket**: https://edx-wiki.atlassian.net/browse/MCKIN-4707

**Reviewers**:

- [x] @smarnach

### DIFF
--- a/problem_builder/public/js/completion.js
+++ b/problem_builder/public/js/completion.js
@@ -16,10 +16,12 @@ function CompletionBlock(runtime, element) {
             return $completion.is(':checked');
         },
 
-        handleSubmit: function(result) {
+        handleSubmit: function(result, options) {
             if (typeof result.submission !== 'undefined') {
                 this.updateCompletion(result);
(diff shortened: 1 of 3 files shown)


smarnach:
:+1:

- [x] I tested this: I tested this on my local devstack.  I set `pb_hide_feedback_if_attempts_remain` to `true` and added a Problem Builder component with a free form answer and a completion check mark, and set maximum answers to 3.  The answer and the completion check mark behave consistently with the patch applied.
- [x] I read through the code
- [x] ~~Includes documentation~~ bugfix, no documentation necessary.

PR #143, opened 2017-02-21
Add management command to manually migrate course_id field.

mtyaka:
Normally the value of deprecated `Answer.course_id` will be copied into the new `Answer.course_key` field by a migration, but if your answer table is very large, you might want to use this management command to perform the copying in batches instead.

The new management command added by this patch lets you copy the column in batches, with a configurable batch size and sleep time between each batch.

**Discussion**: https://github.com/edx/edx-platform/pull/14327#issuecomment-280043022

**Testing instructions**:

1. Install this branch of problem builder. Add some Answer type blocks to a unit in studio and fill them in the LMS to populate some entries in the `problem_builder_answer` table.
2. Current release of problem builder does not allow `NULL` values in the `course_key` field, so you'll have to temporarily modify the table to be able to nullify the values via mysql console:
```sql
ALTER TABLE problem_builder_answer MODIFY course_key VARCHAR(255);
UPDATE problem_builder_answer SET course_id=course_key;
UPDATE problem_builder_answer SET course_key=NULL;
```
3. Make sure `course_key` field of all rows in the table is empty (`SELECT * FROM problem_builder_answer`).
4. Run the management command as the edxapp user:
```
python manage.py lms copy_deprecated_course_id --settings=devstack 
```
5. Make sure that the values of `course_id` have been copied into `course_key` (`SELECT * FROM problem_builder_answer`).
6. Repeat steps 2 - 5 while playing with different `--batch-size` and `--sleep` values.
7. Put your answer table back into the correct state:
```sql
ALTER TABLE problem_builder_answer MODIFY course_key VARCHAR(255) NOT NULL;
```

### DIFF
--- a/problem_builder/management/commands/copy_deprecated_course_id.py
+++ b/problem_builder/management/commands/copy_deprecated_course_id.py
@@ -1,0 +1,46 @@
+import math
+import time
+from django.core.management.base import BaseCommand
+from problem_builder.models import Answer
+
+
+class Command(BaseCommand):
+    """
+    Copy content of the deprecated Answer.course_id column into Answer.course_key in batches.
+    The batch size and sleep time between batches are configurable.
+    """
+    help = 'Copy content of the deprecated Answer.course_id column into Answer.course_key in batches'
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--batch-size',
+            help='The size of each batch of records to copy (default: 5000).',
+            type=int,
+            default=5000,
+        )
+        parser.add_argument(
+            '--sleep',
+            help='Number of seconds to sleep before processing the next batch (default: 5).',
+            type=int,
+            default=5
+        )
+
+    def handle(self, *args, **options):
+        batch_size = options['batch_size']
+        sleep_time = options['sleep']
+        queryset = Answer.objects.filter(course_key__isnull=True)
+        batch_count = int(math.ceil(queryset.count() / float(batch_size)))
+
+        self.stdout.write(
+            "Copying Answer.course_id field into Answer.course_key in batches of {}".format(batch_size)
+        )
+
+        for batch in xrange(1, batch_count + 1):
+            for answer in queryset[:batch_size]:
+                answer.course_key = answer.course_id
+                answer.save()
+            self.stdout.write("Processed batch {} of {}".format(batch, batch_count))
+            if batch != batch_count:
+                time.sleep(sleep_time)
+
+        self.stdout.write("Successfully copied Answer.course_id into Answer.course_key")
--- a/circle.yml
+++ b/circle.yml
@@ -11,7 +11,7 @@ dependencies:
     - "pip install -r requirements.txt"
     - "pip install -r $VIRTUAL_ENV/src/xblock-sdk/requirements/base.txt"
     - "pip install -r $VIRTUAL_ENV/src/xblock-sdk/requirements/test.txt"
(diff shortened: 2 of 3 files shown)


problem_builder/management/commands/copy_deprecated_course_id.py
```diff
@@ -0,0 +1,45 @@
...
+            help='Number of seconds to sleep before processing the next batch (default: 5).',
+            type=int,
+            default=5
+        )
+
+    def handle(self, *args, **options):
+        batch_size = options['batch_size']
+        sleep_time = options['sleep']
+        queryset = Answer.objects.filter(course_key__isnull=True)
+        batch_count = (queryset.count() / batch_size) + 1
```
bdero: Nit: This should do a floating point division with a ceiling operation instead - otherwise cases where the queryset length is divisible by the batch_size will result in one additional batch count than necessary. E.g.:

```python
batch_count = int(math.ceil(queryset.count() / float(batch_size)))
```

bdero:
This looks good @mtyaka - I just had one nit about the batch_count calculation, which you can choose to ignore if you want since I verified it won't cause an error either way. :+1:

- [x] I tested this by creating a bunch of problem builder problems with this version checked out, moved all of the course_key fields to course_id + nulled out the course_key field, and repeated for several different combinations of `--batch-size` and `--sleep`.
- [x] I read through the code
- [ ] ~I checked for accessibility issues~ N/A.
- [ ] ~Includes documentation~ Not necessary. This is a one-time tool to help edX migrate problem builder. The doc string is informative enough.

mtyaka:
Rebased from 768c171957848e6899492c6bdc3af6e86a713bcc

problem_builder/management/commands/copy_deprecated_course_id.py
mtyaka: @bdero Thanks, done!

problem_builder/management/commands/copy_deprecated_course_id.py
```diff
@@ -0,0 +1,46 @@
...
+            help='Number of seconds to sleep before processing the next batch (default: 5).',
+            type=int,
+            default=5
+        )
+
+    def handle(self, *args, **options):
+        batch_size = options['batch_size']
+        sleep_time = options['sleep']
+        queryset = Answer.objects.filter(course_key__isnull=True)
+        batch_count = int(math.ceil(queryset.count() / float(batch_size)))
```
maxrothman: Does this actually do the equivalent of `SELECT count(*)`?

maxrothman:
Thanks! This looks good.

problem_builder/management/commands/copy_deprecated_course_id.py
mtyaka: Yes, `queryset.count()` performs a `SELECT count(*)`.

maxrothman: That's going to be a performance issue for large tables, isn't it? And large tables is how we got down this road in the first place.

bradenmacdonald: @mtyaka Can we just avoid counting how many batches there are, and run the batches one by one in a `while` loop until the batch happens to be empty?

mtyaka: @bradenmacdonald Sure, I opened a PR to do that at https://github.com/open-craft/problem-builder/pull/145, but I am away until March 5th so if this needs to land before then somebody else will have to take care of merging & releasing this.

PR #144, opened 2017-02-22
Heading restructure

MHaton:
As discussed with @pomegranited , have had a first pass at amending headings within the templates to allow for the heading structure change from Eucalyptus. For the most part it's just bumping them down a level, with one being bumped down from 2 to 4 (as unless it's not the element I'm thinking of, its heading level was too high to begin with).

Note that there's references to other H2+H3's etc. contained within things like the python tests that I've not touched, so presumably those will fail and also need updating, I just didn't want to mess with anything I don't understand. I've also not amended the apros theme CSS. Basically, half the task still to do.

### DIFF
--- a/problem_builder/tests/integration/test_titles.py
+++ b/problem_builder/tests/integration/test_titles.py
@@ -48,12 +48,12 @@ class TitleTest(SeleniumXBlockTest):
         self.set_scenario_xml(xml)
         pb_element = self.go_to_view()
         if expected_title is not None:
-            h2 = pb_element.find_element_by_css_selector('h2')
-            self.assertEqual(h2.text, expected_title)
+            h3 = pb_element.find_element_by_css_selector('h3')
+            self.assertEqual(h3.text, expected_title)
         else:
-            # No <h2> element should be present:
-            all_h2s = pb_element.find_elements_by_css_selector('h2')
-            self.assertEqual(len(all_h2s), 0)
+            # No <h3> element should be present:
+            all_h3s = pb_element.find_elements_by_css_selector('h3')
+            self.assertEqual(len(all_h3s), 0)
 
 
 class StepTitlesTest(SeleniumXBlockTest):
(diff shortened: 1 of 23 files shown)


pomegranited:
@MHaton If you could go ahead and update the apros.css and the python tests, that would be great.

I'll run the tests and make sure they pass, and we'll test this in the Apros environment too.

MHaton:
@pomegranited Updated what I think is the last few places h2 & old h3's appear. Needs testing, my workbench has stopped working.

pomegranited:
Thank you for this, @MHaton !  I've asked to get this into next week's sprint, but it might end up being the week after that.

bradenmacdonald:
@MHaton With the ongoing release of Ficus we're pretty busy this week so we've scheduled a review of this PR for the week of March 6. Sorry for the delay.

problem_builder/tests/integration/test_titles.py
```diff
@@ -145,9 +145,9 @@ def test_all_the_things(self):
...
                     pb_element = self.go_to_view()
                     if expected_title:
-                        h3 = pb_element.find_element_by_css_selector('h3')
+                        h4 = pb_element.find_element_by_css_selector('h4')
                         self.assertEqual(h3.text, expected_title)
                     else:
-                        # No <h3> element should be present:
-                        all_h3s = pb_element.find_elements_by_css_selector('h3')
-                        self.assertEqual(len(all_h3s), 0)
+                        # No <h4> element should be present:
```
pomegranited: 2 lines above needs to be updated too:

```
self.assertEqual(h4.text, expected_title)
```

MHaton: Sure thing, will do this on Monday :)

MHaton:
Version number updated (now conflicting), test sorted

Take it away @pomegranited  :)

pomegranited:
Tagged as v2.6.8.

PR #145, opened 2017-02-25
Avoid count(*) query in management command.

mtyaka:
This modifies the management command introduced in https://github.com/open-craft/problem-builder/pull/143 to avoid a `COUNT(*)` query for performance reasons.

**Testing instructions**: Please see https://github.com/open-craft/problem-builder/pull/143

**Reviewers**:

- [x] @maxrothman


maxrothman:
This looks good assuming you've tested it.

PR #150, opened 2017-06-29
Add API for native mobile Problem Builder + MCQ view to Problem Builder

replaceafill:
This PR adds `student_view_data` methods to the Problem Builder and MCQ blocks so their data can be exported as JSON and queried through the Courses API.

**Testing instructions**
1. Install this branch into your edx-platform virtual environment.
2. In Studio, as `staff`, add a Problem Builder block to the demo course and set the following attributes:
- Max. attempts allowed
- Feedback Header
- extended_feedback
3. Check the ID of the block in the message:
`url_name for linking to this mentoring question set: ...`
4. Add a Multiple Choice Question to the PB Block setting the following attributes:
- Question
- Weight
-  Message
5. Add a few choices to the Multiple Choice Question, setting tips for some of them.
6. Add one of each type of message to the PB Block (Complete, Incomplete, Max # Attempts) 
7. In the LMS, as `staff`, query the Courses API for the block using the ID of step 3 (replace BLOCK_ID_GOES_HERE in the following URL):

http://localhost:8000/api/courses/v1/blocks/block-v1:edX+DemoX+Demo_Course+type@problem-builder+block@BLOCK_ID_GOES_HERE?all_blocks=true&depth=all&requested_fields=student_view_data

8. In the result mapping look for the `blocks` key. Find the `problem-builder+block` and `pb-mcq+block` 
 blocks and verify that the values in their `student_view_data` mappings match the attributes and messages you set.

### DIFF
--- a/problem_builder/mcq.py
+++ b/problem_builder/mcq.py
@@ -167,6 +167,27 @@ class MCQBlock(SubmittingXBlockMixin, QuestionnaireAbstractBlock):
                 self._(u"A choice value listed as correct does not exist: {choice}").format(choice=choice_name(val))
             )
 
+    def student_view_data(self):
+        """
+        Returns a JSON representation of the student_view of this XBlock,
+        retrievable from the Course Block API.
+        """
+        return {
+            'id': self.name,
+            'type': self.CATEGORY,
+            'question': self.question,
+            'message': self.message,
+            'choices': [
+                {'value': choice['value'], 'content': choice['display_name']}
+                for choice in self.human_readable_choices
+            ],
+            'weight': self.weight,
+            'tips': [
+                {'content': tip.content, 'for_choices': tip.values}
+                for tip in self.get_tips()
+            ],
+        }
+
 
 class RatingBlock(MCQBlock):
     """
--- a/problem_builder/mentoring.py
+++ b/problem_builder/mentoring.py
@@ -905,6 +905,31 @@ class MentoringBlock(BaseMentoringBlock, StudioContainerWithNestedXBlocksMixin, 
         """
         return loader.load_scenarios_from_path('templates/xml')
 
+    def student_view_data(self):
+        """
+        Returns a JSON representation of the student_view of this XBlock,
+        retrievable from the Course Block API.
+        """
+        components = []
+        for child_id in self.children:
+            block = self.runtime.get_block(child_id)
+            if hasattr(block, 'student_view_data'):
+                components.append(block.student_view_data())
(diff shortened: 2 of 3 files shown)


problem_builder/mcq.py
```diff
@@ -167,6 +167,42 @@ def choice_name(choice_value):
                 self._(u"A choice value listed as correct does not exist: {choice}").format(choice=choice_name(val))
             )
 
+    def student_view_data(self):
+        """
+        Returns a JSON representation of the student_view of this XBlock,
+        retrievable from the Course Block API.
+        """
+        return {
+            'id': self.url_name,
```
bradenmacdonald: This should just be `self.name`, which comes from `QuestionMixin`. Please do not use `url_name`, and don't change `url_name` to return `name` in the workbench.

bradenmacdonald:
@replaceafill The reason that the user state is not showing up is that `student_view_data` [is computed without considering any specific user's state](https://github.com/edx/edx-platform/blob/95e24a4b859e3f4fa1465e8bce4bb1bec15e3997/lms/djangoapps/course_api/blocks/transformers/student_view.py#L26-L69). We will have to find/invent a new approach to including the user state in the data sent to offline XBlocks. For now, please remove the student state from this API, and finish this PR up.

problem_builder/mentoring.py
```diff
@@ -905,6 +905,31 @@ def workbench_scenarios():
...
+            block = self.runtime.get_block(child_id)
+            if hasattr(block, 'student_view_data'):
+                components.append(block.student_view_data())
+        return {
+            'max_attempts': self.max_attempts,
+            'extended_feedback': self.extended_feedback,
+            'feedback_label': self.feedback_label,
+            'components': components,
+            'messages': {
+                message_type: get_message_label(message_type)
```
bradenmacdonald: We don't want the message Studio label here (`get_message_label`) - we don't care about that. We want the result of [`self. get_message_content(message_type)`](https://github.com/open-craft/problem-builder/blob/bfa131b63928ef2e60a1c9d50fd22ced957cb25c/problem_builder/mixins.py#L102) which will either be `None` or an HTML string.

bradenmacdonald:
👍  My above comments have been addressed, and this is working well from my manual testing of it, but I don't want to merge yet as described on the ticket.

replaceafill:
I squashed this from a7ca073.

itsjeyd:
Thanks @replaceafill! Merging now.

PR #153, opened 2017-08-03
[MCKIN-5253] - Student_view_user_data

e-kolpakov:
**Description:** This PR adds is prototype to return `student_view_data` from Mentoring Blocks.

**Testing**
1. Import [PB demo course](https://github.com/open-craft/demo-courses/tree/problem-builder)
2. Log to LMS as the `staff` user.
3. Go to http://localhost:8000/courses/:course_id/:block_id/handler/student_view_user_state
4. Check the data returned in the `student_view_user_state` fields. Make sure they are sufficient and make sense.

*Getting block ID*:

Two options:

1. http://localhost:8000/api/courses/v1/blocks/?course_id=course-<course_id>&depth=all&all_blocks=true = manual search
2. Open target block in studio, note the URL

### DIFF
--- a/problem_builder/answer.py
+++ b/problem_builder/answer.py
@@ -115,12 +115,12 @@ class AnswerMixin(XBlockWithPreviewMixin, XBlockWithTranslationServiceMixin, Stu
         if not data.name:
             add_error(u"A Question ID is required.")
 
-    def student_view_user_state(self, context=None):
+    def build_user_state_data(self, context=None):
         """
         Returns a JSON representation of the student data of this XBlock,
         retrievable from the Course Block API.
         """
-        result = super(AnswerMixin, self).student_view_user_state(context)
+        result = super(AnswerMixin, self).build_user_state_data(context)
         answer_data = self.get_model_object()
         result["answer_data"] = {
             "student_input": answer_data.student_input,
--- a/problem_builder/answer.py
+++ b/problem_builder/answer.py
@@ -32,7 +32,7 @@ from xblock.validation import ValidationMessage
 from xblockutils.resources import ResourceLoader
 from xblockutils.studio_editable import StudioEditableXBlockMixin, XBlockWithPreviewMixin
 from problem_builder.sub_api import SubmittingXBlockMixin, sub_api
-from .mixins import QuestionMixin, XBlockWithTranslationServiceMixin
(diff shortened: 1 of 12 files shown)


problem_builder/mixins.py
```diff
@@ -176,3 +177,25 @@ class NoSettingsMixin(object):
     def studio_view(self, _context=None):
         """ Studio View """
         return Fragment(u'<p>{}</p>'.format(self._("This XBlock does not have any settings.")))
+
+
+class StudentViewUserStateMixin(object):
+    NESTED_BLOCKS_KEY = "components"
+    INCLUDE_SCOPES = (Scope.user_state, Scope.user_info, Scope.preferences)
+
+    def student_user_view_user_state(self, context=None):
```
bradenmacdonald: In general, `student_user_view_user_state` may not simply be a set of fields... we may want to simplify/transform the data first. I think it's fine in this case though.

e-kolpakov: @bradenmacdonald In that case any block that need that override can override this method. OOP, right? :)

problem_builder/answer.py
```diff
@@ -115,6 +115,20 @@ def add_error(msg):
         if not data.name:
             add_error(u"A Question ID is required.")
 
+    def student_view_user_state(self, context=None):
```
bradenmacdonald: This needs to be an `@XBlock.handler` (but not `json_handler` since that only works with POST requests)

e-kolpakov: @bradenmacdonald student_view_data does not have it. In that case either `student_view_data` is broken or my assumptions are not quite correct. Do we need `student_view_user_state` to appear in the response from course API, or just as a handler on a block?

bradenmacdonald: We need `student_view_data` to appear in the course blocks API (as it already does), but `student_view_user_state` does not need to appear in the course blocks API. Instead we want `student_view_user_state` to be a handler, which we call just before displaying the block. It might make more sense to call it `get_student_view_user_state`, which is a more handler-like name.

e-kolpakov:
@pomegranited Thanks! However, both of us have missed that version was not bumped - submitted a followup: #155 - going to merge as soon as Ci build succeeds and than make a release tag

pomegranited:
Gah, sorry about that, @e-kolpakov !  Thanks for noticing.

PR #154, opened 2017-08-04
[OC-2856] Fixed deprecation warnings for the Answer and Share models

bradmerlin:
This PR simply adds `app_label`s to the problem-builder Answer and Share models (and silences Django deprecation warnings).

**JIRA tickets**: Implements [OC-2856](https://tasks.opencraft.com/browse/OC-2856), [PLAT-1644](https://openedx.atlassian.net/browse/PLAT-1644).

**Dependencies**: None

**Sandbox URL**: TBD - sandbox is being provisioned.

**Deployment targets**: edx.org and edge.edx.org

**Merge deadline**: ASAP

**Testing instructions**:

1. Ensure that the [problem-builder](https://github.com/open-craft/problem-builder) XBlock is installed and running on Studio
2. Check console output and check for absence of the following deprecation warnings:

```
2017-08-04 04:28:57,685 WARNING 23527 [py.warnings] base.py:116 - /edx/src/problem-builder/problem_builder/models.py:59: RemovedInDjango19Warning: Model class problem_builder.models.Share doesn't declare an explicit app_label and either isn't in an application in INSTALLED_APPS or else was imported before its application was loaded. This will no longer be supported in Django 1.9.
```

1. This is my first PR for Opencraft. Please be blunt about anything that I've missed.

**Reviewers**
- [x] @pomegranited
- [x] @bmedx

### DIFF
--- a/problem_builder/models.py
+++ b/problem_builder/models.py
@@ -35,6 +35,9 @@ class Answer(models.Model):
     """
 
     class Meta:
+        # Since problem_builder isn't added to INSTALLED_APPS until it's imported,
+        # specify the app_label here.
+        app_label = 'problem_builder'
         unique_together = (
             ('student_id', 'course_id', 'name'),
             ('student_id', 'course_key', 'name'),
@@ -69,4 +72,7 @@ class Share(models.Model):
     notified = models.BooleanField(default=False, db_index=True)
 
     class Meta(object):
+        # Since problem_builder isn't added to INSTALLED_APPS until it's imported,
+        # specify the app_label here.
+        app_label = 'problem_builder'
         unique_together = (('shared_by', 'shared_with', 'block_id'),)
--- a/circle.yml
+++ b/circle.yml
@@ -11,7 +11,7 @@ dependencies:
     - "pip install -r requirements.txt"
     - "pip install -r $VIRTUAL_ENV/src/xblock-sdk/requirements/base.txt"
     - "pip install -r $VIRTUAL_ENV/src/xblock-sdk/requirements/test.txt"
-    - "pip uninstall -y xblock-problem-builder && python setup.py sdist && pip install dist/xblock-problem-builder-2.6.9.tar.gz"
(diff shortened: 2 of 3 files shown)


macdiesel:
@doctoryes FYI.

problem_builder/models.py
```diff
@@ -35,6 +35,7 @@ class Answer(models.Model):
     """
 
     class Meta:
+        app_label = 'problem_builder'
```
pomegranited: Could you please add a comment here about why this attribute is required?
  Since it's not obvious from the [Django docs](https://docs.djangoproject.com/en/1.8/ref/models/options/#app-label), we want to make sure it isn't deleted down the track.

E.g., something about how it's imported as part of the [`OPTIONAL_APPS`](https://github.com/edx/edx-platform/blob/986afbfa383e2e6bcebfe801c86713f2bae1b9de/lms/envs/common.py#L2795) for lms/Studio, and therefore not added to the `INSTALLED_APPS` list until after it's imported.

problem_builder/models.py
```diff
@@ -69,4 +70,5 @@ class Share(models.Model):
     notified = models.BooleanField(default=False, db_index=True)
 
     class Meta(object):
+        app_label = 'problem_builder'
```
pomegranited: Same comment as above.

pomegranited:
Looks good @bradmerlin !  :+1:  once my comment re docstrings is addressed.

This fix removes the app warnings in the most straightforward way possible, given that `xblock-problem-builder` is not installed on every platform.

- [x] I tested this using `./manage.py lms migrate` on my devstack.  Verified that the `RemovedInDjango19Warning` warnings appeared when the `master` branch was installed, but they're gone with this branch installed.
- [x] I read through the code.
- [x] <strike>I checked for accessibility issues</strike>
- [ ] Includes documentation - see comments.

bmedx:
These changes seem fine. We've been changing our repositories to use tox for testing multiple versions of Python and Django which helps us prevent regressions from creeping in while we are in the process of the Django upgrade, I'd suggest adding that here if you have the time. Either way this PR is 👍

PR #156, opened 2017-08-10
Add docs about API methods which can be used form native apps

mtyaka:
This documents `student_view_data`, `student_view_user_state`, and custom handlers for each supported component.

**Notes and Concerns**:

The APIs are not very consistent between some components. Some APIs consume data in weird shape, most notably [freeform answer submit handler](https://github.com/open-craft/problem-builder/blob/502d6eeea33a364261ee36c3e5dcbd40134ec0f1/doc/Native%20APIs.md#post-submit-data), probably because it was never designed to be public.

Currently `components` entry in `student_view_user_state` is a dict where keys edX block keys (i.e. `'i4x://TstCrs/TS111/pb-answer/a0b04a13d3074229b6be33fbc31de233'`), but in `student_view_data` we typically use our internal customizable unique IDs (`self.name`).
We might want to make that more consistent.

I don't like how we simply expose all `Scope.user_state` fields in `student_view_user_state` by default. We should hand pick the fields that we want to expose directly instead, and we might also prefer to include data derived from user state in a form that is more suitable for API consumers. We perform normalizations in several places internally, which is skipped if we expose the fields directly.

**Reviewers**:

- [ ] @pomegranited

### DIFF
--- a/doc/Native APIs.md
+++ b/doc/Native APIs.md
@@ -335,6 +335,10 @@ Multiple Response Question (`pb-mrq`)
 - `question`: (string) The content of the question.
 - `message`: (string) General feedback provided when submitting.
 - `hide_results`: (boolean) `true` if results should be hidden.
+- `choices`: (array) A list of objects providing info about available
+  choices. See below for more info.
+- `tips`: (array) A list of objects providing info about tips defined for the
+  problem. See below for more info.
 
 ### `student_view_user_state`
 
--- a/doc/Native APIs.md
+++ b/doc/Native APIs.md
@@ -5,10 +5,19 @@ This documents the APIs that can be used to implement native wrappers. There are
 three types of APIs:
 
 - `student_view_data`: Exposes block settings and content as JSON data. Can be
-  retrieved via the edX Course Blocks API. Does not include user-specific data
-  (i.e. items derived from `Scope.user_state` fields).
-- `student_view_user_state`: XBlock handler which returns user state data in
-  JSON format.
+  retrieved via the edX
+  [Course Blocks API](http://edx.readthedocs.io/projects/edx-platform-api/en/latest/courses/blocks.html),
+  e.g.,
+  ```
+  GET https://<lms_server_url>/api/courses/v1/blocks/?course_id=<course_id>&username=<username>&depth=all&requested_fields=student_view_data&student_view_data=<xblock_types>
+  ```
+  Does not include user-specific data, which is available from `student_view_user_state`.
+- `student_view_user_state`: XBlock handler which returns the currently logged
+  in user's state data in JSON format (e.g. items derived from `Scope.user_state`
+  fields).
+  ```
+  GET https://<lms_server_url>/courses/<course_id>/xblock/<xblock_id>/handler/student_view_user_state
+  ```
 - Custom XBlock handlers used for submitting user input and recording user actions.
 
 Problem Builder (`problem-builder`)
@@ -16,45 +25,46 @@ Problem Builder (`problem-builder`)
 
 ### `student_view_data`
 
-- `max_attempts`: Max number of allowed attempts (integer).
-- `extended_feedback`: `true` if extended feedback is enabled for this block
-  (boolean).
-- `feedback_label`: Feedback message header (string).
-- `messages`: A dict with three string entries: `completed`, `incomplete`, and
-  `max_attempts_reached`. See below for info on each (object).
-- `components`: A list of `student_view_data` output of all immediate child
-  components which implement `student_view_data`. Child components which don't
-  implement `student_view_data` are omitted from the list (array).
+- `max_attempts`: (integer) Max number of allowed attempts.
+- `extended_feedback`: (boolean) `true` if extended feedback is enabled for this
+  block.
+- `feedback_label`: (string) Feedback message header.
+- `messages`: (object) A dict with three string entries: `completed`,
+  `incomplete`, and `max_attempts_reached`. See below for info on each.
+- `components`: (array) A list of `student_view_data` output of all immediate
+  child components which implement `student_view_data`. Child components which
+  don't implement `student_view_data` are omitted from the list.
 
 #### `messages`
 
-These messages are displayed as feedback, depending on the state of the block.
+These messages are displayed as feedback, depending on the state of the
+block. Any of these may be `null`.
 
-- `completed`: Message to be shown if the user successfully completes the
-  problem. May contain HTML (string).
-- `incomplete`: Message to be shown if the user did not successfully complete
-  the problem, but still has available attempts left. May contain HTML (string).
-- `max_attempts_reached`: Message to be shown if the user did not complete the
-  problem and has no available attempts left.
+- `completed`: (string) Message to be shown if the user successfully completes
+  the problem. May contain HTML.
+- `incomplete`: (string) Message to be shown if the user did not successfully
+  complete the problem, but still has available attempts left. May contain HTML.
+- `max_attempts_reached`: (string) Message to be shown if the user did not
+  complete the problem and has no available attempts left.
 
 ### `student_view_user_state`
 
-- `num_attempts`: Number of attempts used so far (integer).
-- `completed`: `true` if the user successfully completed the problem at least
-  once (boolean).
-- `student_results`: A list of user submitted answers with metadata for each
-  child component. See below for more info (array).
-- `components`: A dict of `student_view_user_state` output of all immediate
-  child components which implement `student_view_user_state`. Dict keys are
-  component IDs and the value corresponding to each key is the output of
-  `student_view_user_state`(object). Child components which don't implement
-  `student_view_user_state` are not included in the dict.
+- `num_attempts`: (integer) Number of attempts used so far.
+- `completed`: (boolean) `true` if the user successfully completed the problem at least once.
+- `student_results`: (array) A list of user submitted answers with metadata for
+  each child component. See below for more info.
+- `components`: (object) Contains the `student_view_user_state` output of all
+  immediate child components which implement `student_view_user_state`. Each key
+  is a component ID (string), and the value is the component's
+  `student_view_user_state` (object). Child components which don't implement
+  student_view_user_state are not included in the dict.
 
 #### `student_results`
 
 The `student_results` field is an array of two-element arrays. Each nested array
-has the form of `[component_name, data]`. Only components which the student
-already submitted are present in the array. The structure of the `data` object
+has the form of `[component_name, data]`. If the user has not made any
+submissions to particular component, then the element corresponding to that
+component may be omitted from the array. The structure of the `data` object
 depends on the type of the child component and is described separately for each
 type of child component below.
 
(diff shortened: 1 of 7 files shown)


doc/Native APIs.md
```diff
@@ -0,0 +1,417 @@
+Native API Documentation
+========================
+
+This documents the APIs that can be used to implement native wrappers. There are
+three types of APIs:
+
+- `student_view_data`: Exposes block settings and content as JSON data. Can be
+  retrieved via the edX Course Blocks API. Does not include user-specific data
+  (i.e. items derived from `Scope.user_state` fields).
```
bradenmacdonald: Can you please add the example URL for this?

```
GET https://<lms_server_url>/api/courses/v1/blocks/?course_id=<course_id>&username=<username>&depth=all&requested_fields=student_view_data
```

doc/Native APIs.md
```diff
@@ -0,0 +1,417 @@
...
+- `student_view_user_state`: XBlock handler which returns user state data in
+  JSON format.
```
bradenmacdonald: Please add the example URL here too:

```
GET https://<lms_server_url>/courses/<course_id>/xblock/<xblock_id>/handler/student_view_user_state
```

bradenmacdonald:
@mtyaka 

> I don't like how we simply expose all Scope.user_state fields in student_view_user_state by default. We should hand pick the fields that we want to expose directly instead, and we might also prefer to include data derived from user state in a form that is more suitable for API consumers. We perform normalizations in several places internally, which is skipped if we expose the fields directly.

I was afraid of that... Can you please create a ticket describing those changes we should make? We need to do them ASAP before the mobile team depends on the current API. See also https://github.com/open-craft/problem-builder/pull/153#discussion_r131273180 where I made the same point.

> Some APIs consume data in weird shape, most notably freeform answer submit handler

Can you please also make a ticket to fix that?

Also, if you want to just quickly make any of those changes as part of this ticket, feel free.

doc/Native APIs.md
```diff
@@ -0,0 +1,417 @@
...
+- `num_attempts`: Number of attempts used so far (integer).
+- `completed`: `true` if the user successfully completed the problem at least
+  once (boolean).
+- `student_results`: A list of user submitted answers with metadata for each
+  child component. See below for more info (array).
+- `components`: A dict of `student_view_user_state` output of all immediate
+  child components which implement `student_view_user_state`. Dict keys are
+  component IDs and the value corresponding to each key is the output of
+  `student_view_user_state`(object). Child components which don't implement
+  `student_view_user_state` are not included in the dict.
```
pomegranited: Slight rewording:

- `components`: (object) Contains the `student_view_user_state` output of all immediate child components which implement `student_view_user_state`. Each key is a component ID (string), and the value is the component's `student_view_user_state`(object).
  Child components which don't implement `student_view_user_state` are not included in the dict.

doc/Native APIs.md
```diff
@@ -0,0 +1,417 @@
...
+  feedback messages described under `messages` above. Can be `null` if no
+  messages has been defined. May contain HTML (string).
+- `max_attempts`: Same as `max_attempts` under `student_view_data`.
+- `num_attempts`: Same as `num_attempts` under `student_view_user_state`.
+
+
+Step Builder (`step-builder`)
+-----------------------------
+
+### `student_view_data`
```
pomegranited: I'm also seeing these fields in the JSON result:
```json
{  
   "active_step":0,
   "hide_prev_answer":true,
   "num_attempts":0,
   ...
}
```

doc/Native APIs.md
```diff
@@ -0,0 +1,417 @@
...
+
+#### `submit`
+
+The `submit` XBlock JSON handler is used for submitting student input. The
+`POST` data should be a dict where keys are child component names. The values
+depend on the component type and are described separately for each type of child
+component below.
+
+In addition to child component names, the `POST` data should also include the
+`active_step` field. The value should be the index of the current step.
```
pomegranited: Can you not submit answers to a step that is not currently active?  (Asking because I'm not sure)

doc/Native APIs.md
```diff
@@ -0,0 +1,417 @@
...
+--------------------------
+
+### `student_view_data`
+
+- `type`: Always equals `"sb-step"` for Mentoring Step components (string).
+- `title`: Step component's display name (string).
+- `show_title`: `true` if the title should be displayed (boolean).
+- `next_button_label`: Text label of the "Next Step" button (string).
+- `message`: Feedback or instructional message which should be displayed after
+  student submits the step (string).
```
pomegranited: Can be `null`.

doc/Native APIs.md
```diff
@@ -0,0 +1,417 @@
...
+
+When submitting the problem the data should equal to the string value of the
+selected choice.
+
+Rating Question (`pb-rating`)
+-----------------------------
+
+### `student_view_data`
+
+Identical to MCQ questions except that the `type` field always equals
```
pomegranited: Add link: [MCQ questions](#pb-mcq)

doc/Native APIs.md
```diff
@@ -0,0 +1,417 @@
...
+`"pb-rating"`.
+
+### `student_view_user_state`
+
+- `student_choice`: The value of the last submitted choice (string).
+
+### `student_results`
+
+Identical to MCQ questions.
```
pomegranited: Add link: [MCQ questions](#pb-mcq)

doc/Native APIs.md
```diff
@@ -0,0 +1,417 @@
...
+- `max_attempts`: Max number of allowed attempts (integer).
+- `components`: A list of `student_view_data` output of all immediate child
+  components which implement `student_view_data`. For `step-builder`, immediate
+  children are typically `sb-step` and `sb-review-step` type components. Child
+  components which don't implement `student_view_data` are omitted from the list
+  (array).
+
+### `student_view_user_state`
+
+Same as Problem Builder above, but also contains these additional fields:
```
pomegranited: Add link: [Problem Builder](#problem-builder-problem-builder)

doc/Native APIs.md
```diff
@@ -0,0 +1,417 @@
...
+- `score`: Student's score. One of `1` or `0` (integer).
+- `weight`: Child component's weight attribute (float).
+- `submission`: The value of the choice that the user selected (string).
+- `message`: General feedback. May contain HTML (string).
+- `tips`: HTML representation of tips. May be `null` (string).
+
+### POST Submit Data
+
+When submitting the problem the data should equal to the string value of the
+selected choice.
```
pomegranited: Please provide example.

doc/Native APIs.md
```diff
@@ -0,0 +1,417 @@
...
+  `student_view_user_state` above.
+- `completed`: Same as `completed` described under `student_view_user_state`
+  above.
+- `message`: One of the `completed`, `incomplete`, or `max_attempts_reached`
+  feedback messages described under `messages` above. Can be `null` if no
+  messages has been defined. May contain HTML (string).
+- `max_attempts`: Same as `max_attempts` under `student_view_data`.
+- `num_attempts`: Same as `num_attempts` under `student_view_user_state`.
+
+
```
pomegranited: As with the other handler, please provide example URL, e.g. 
```
POST https://<lms_server_url>/courses/<course_id>/xblock/<xblock_id>/handler/submit
```

doc/Native APIs.md
```diff
@@ -0,0 +1,417 @@
...
+`active_step` field. The value should be the index of the current step.
+
+#### `try_again`
+
+The `try_again` XBlock JSON handler can be used when the student reaches the
+final step. Sending a `POST` request to `try_again` will reset the problem and
+all of its children.
+
+Returns a JSON object containing the new value of `active_step`.
+
```
pomegranited: As with the other handlers, please provide example URL, e.g. 
```
POST https://<lms_server_url>/courses/<course_id>/xblock/<xblock_id>/handler/try_again
```

doc/Native APIs.md
```diff
@@ -0,0 +1,417 @@
...
+### `student_view_data`
+
+- `type`: Always equals `"pb-slider"` for Ranged Value Slider components (string).
+- `id`: Unique ID (name) of the component (string).
+- `title`: Display name of the component (string).
+- `hide_header`: `true` if header should be hidden (boolean).
+- `question`: The content of the question (string).
+- `min_label`: Label for the low end of the range (string).
+- `max_label`: Label for the high end of the range (string).
```
pomegranited: Also shown: `initial_value`: (integer)

doc/Native APIs.md
```diff
@@ -0,0 +1,417 @@
...
+These messages are displayed as feedback, depending on the state of the block.
+
+- `completed`: Message to be shown if the user successfully completes the
+  problem. May contain HTML (string).
+- `incomplete`: Message to be shown if the user did not successfully complete
+  the problem, but still has available attempts left. May contain HTML (string).
+- `max_attempts_reached`: Message to be shown if the user did not complete the
+  problem and has no available attempts left.
+
+### `student_view_user_state`
```
pomegranited: In fact, I see `"edx_notes_visibility"` for all the child block types.

doc/Native APIs.md
```diff
@@ -0,0 +1,417 @@
...
+  child components which implement `student_view_user_state`. Dict keys are
+  component IDs and the value corresponding to each key is the output of
+  `student_view_user_state`(object). Child components which don't implement
+  `student_view_user_state` are not included in the dict.
+
+#### `student_results`
+
+The `student_results` field is an array of two-element arrays. Each nested array
+has the form of `[component_name, data]`. Only components which the student
+already submitted are present in the array. The structure of the `data` object
```
pomegranited: If the user has not made any submissions, then this element may be omitted from the response.

doc/Native APIs.md
mtyaka: Thanks! I will update according to your suggestions. The example in the Course API docs actually uses both,`requested_fields=student_view_data` and `student_view_data=<block-type>`. I *think* I had to use both when playing with the API on edx/master some time ago, but maybe I just never tried without `requested_fields`.

doc/Native APIs.md
```diff
@@ -0,0 +1,417 @@
...
+- `student_view_user_state`: XBlock handler which returns user state data in
+  JSON format.
+- Custom XBlock handlers used for submitting user input and recording user actions.
+
+Problem Builder (`problem-builder`)
+-----------------------------------
+
+### `student_view_data`
+
+- `max_attempts`: Max number of allowed attempts (integer).
```
mtyaka: Good idea, will change.

doc/Native APIs.md
```diff
@@ -0,0 +1,417 @@
...
+- `messages`: A dict with three string entries: `completed`, `incomplete`, and
+  `max_attempts_reached`. See below for info on each (object).
+- `components`: A list of `student_view_data` output of all immediate child
+  components which implement `student_view_data`. Child components which don't
+  implement `student_view_data` are omitted from the list (array).
+
+#### `messages`
+
+These messages are displayed as feedback, depending on the state of the block.
+
```
mtyaka: Thanks! Will update.

doc/Native APIs.md
mtyaka: I intentionally omitted fields which I think are not needed. I think `step` and `next_step` are only used by the deprecated assessment mode, and `attempted` is not used anywhere AFAIK and we might want to remove it in  the future, so I don't want API consumers to start depending on it.

`"edx_notes_visibility"` is some edx thing not directly related to problem builder.

I will make a ticket to implement `student_view_user_state` for each block manually instead of exposing all `user_state` fields by default. It's more work, but that way we won't accidentally start exposing internal fields via the API, and have API consumers start depending on them.

doc/Native APIs.md
mtyaka: Thanks! WIll update.

doc/Native APIs.md
mtyaka: Thanks, will update.

doc/Native APIs.md
mtyaka: Good idea, will do.

doc/Native APIs.md
mtyaka: I removed those fields as part of this PR because `active_step` and `num_attempts` are derived from user state and therefore don't work correctly in `student_view_data`. You might have to publish a random unit in your course to force a refresh of the cache.

doc/Native APIs.md
mtyaka: Good idea, will do.

doc/Native APIs.md
mtyaka: You can't, the submit handler will refuse it:

https://github.com/open-craft/problem-builder/blob/3b500caf8c33035c9e3bc9ee49ae7b9ed5aa3d27/problem_builder/mentoring.py#L1161-L1162

doc/Native APIs.md
mtyaka: Will do.

doc/Native APIs.md
mtyaka: Thanks, will update.

doc/Native APIs.md
mtyaka: Will do.

doc/Native APIs.md
mtyaka: Will add, thanks.

doc/Native APIs.md
mtyaka: I removed `initial_value` in this PR because it depends on user state. You might have to refresh your cache.

mtyaka:
Thanks @pomegranited! I addressed your comments. If it looks good to you know, I will squash and merge.

doc/Native APIs.md
pomegranited: 👍

doc/Native APIs.md
pomegranited: @mtyaka - I just tried again, and now it's working for me with just `requested_fields=student_view_data` -- not quite sure why it wasn't before?

bradenmacdonald:
@pomegranited  I'm just going to merge this now, and it will get pulled into another version bump like #158.

PR #160, opened 2017-08-15
New course key migration script.

jcdyer:
I have a suspicion that the problem was with the way the queries depend on the previous queries getting committed.  I know edx.org uses a funky transaction mode for their MySQL instance which might cause problems with this behavior, and might also cause differences with a development box that uses a newer version of mysql (that might use a different transaction mode).  This implementation makes the individual queries independent of one another, and also only performs one update query per batch, rather than one per object.

I still need to do manual tests, but wanted to get feedback.

Compare the implementation to the source I borrowed from here: https://github.com/open-craft/problem-builder/pull/143/files#diff-cd075b20c253e110da4aef02f64019d0R38

@bradenmacdonald @feanil


problem_builder/management/commands/copy_deprecated_course_ids.py
```diff
@@ -0,0 +1,57 @@
...
+            help='Number of seconds to sleep before processing the next batch (default: 5).',
+            type=int,
+            default=5
+        )
+
+    def handle(self, *args, **options):
+        batch_size = options['batch_size']
+        sleep_time = options['sleep']
+        max_pk = Answer.objects.order_by('-pk')[0].pk
+        batch_start = 0
```
feanil: If we have to stop this command and re-start it, we'll want to start at a different ID. We should make this configurable.  Optionally it could look at the lowest PK where the new field is empty?  That may be more clever than necessary.

problem_builder/migrations/0004_copy_course_ids.py
```diff
@@ -0,0 +1,24 @@
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+def copy_course_id_to_course_key(apps, schema_editor):
```
bradenmacdonald: Should we just replace this forward migration with a noop in this `2.6.5.patch1` build, so devops doesn't have to bother manually faking it?

jcdyer: If we try to merge this branch into master, it will cause some pain to have a different version here.  If we don't try to merge this branch into master, this version of the management command will disappear and will never have lived on the master branch, which may make it hard to dig up later on, if it's needed for historical reasons.  

My suspicion is that faking the migration will be the lowest amount of pain, but I'll defer to @feanil's judgment on which approach they'd prefer.

jcdyer: Actually, if we're just going to cut a release directly off this branch it shouldn't matter.  We can still resolve conflicts and merge the branch afterward.  I've set the migration to noop.

bradenmacdonald: Yep, exactly. Let's release with this exact branch, and when we merge to master later (which we should, to keep the code, as you point out) we can resolve the migration conflict in favor of the existing migration that is not a noop.

jcdyer:
@bradenmacdonald @feanil The current version is working for a manual test, but there still seems to be an error in the circle-ci test suite (unrelated to my PR).  Sorry I squashed commits without thinking.  I fixed an off-by-one error and did some other clean-up to get the script working (I had the queryset filtered using __ge instead of __gte, the circle.yml file had some fiddly bits that needed fiddling, etc.).

With 1,000,000 Answer entries, batches of 5000 (the default) and no sleep between batches, All 200 batches run in about 90 seconds.  I reduced the default sleep time from 5 seconds between batches to 1 second between batches (Which means it will still spend more time sleeping than working, but not waste quite as much wall clock time).

Before this gets run, we will still need to 

1. tag a release. 
2. update edx-platform to use that release.
3. Do a test run on stage.

I have permission to do \#1 (just need thumbs on this PR), and create a PR for \#2 (after the tag exists), but will need devops to do \#3.

I will out of town the rest of this week.  I'll be able to do minor tasks (like #s 1 and 2 above), but won't be available for larger work.  We can either pick this up next week, or I can hand it off to @bradenmacdonald to shepherd through.

problem_builder/management/commands/copy_deprecated_course_ids.py
```diff
@@ -0,0 +1,69 @@
...
+            "Copying Answer.course_id field into Answer.course_key in batches of at most {}".format(
+                batch_size
+            )
+        )
+        while batch_start <= max_pk:
+
+            queryset = Answer.objects.filter(
+                pk__gte=batch_start,
+                pk__lt=batch_stop,
+                course_key='',
```
bradenmacdonald: Looking at https://github.com/open-craft/problem-builder/blob/master/problem_builder/migrations/0003_auto_20161124_0755.py it seems that when we created the `course_key` column, we made it nullable with a null default, so I suspect that the live server actually has null values in this column, not empty strings. (Note that in migration 0005 we later fixed that, but that migration isn't on edX.org yet).

Using this should work:
```python
            queryset = Answer.objects.filter(
                pk__gte=batch_start,
                pk__lt=batch_stop,
            ).filter(Q(course_key__isnull=True) | Q(course_key=''))
```

problem_builder/management/commands/copy_deprecated_course_ids.py
```diff
@@ -0,0 +1,69 @@
...
+            default=1,
+        )
+
+    def handle(self, *args, **options):
+        batch_size = options['batch_size']
+        sleep_time = options['sleep']
+        offset = options['offset']
+        max_pk = Answer.objects.order_by('-pk')[0].pk
+        batch_start = offset
+        batch_stop = batch_size
```
jcdyer: Fixed.  Thanks

problem_builder/management/commands/copy_deprecated_course_ids.py
jcdyer: Ok.  Updated.

jcdyer:
Renamed command to problem_builder_migrate_keys.  I'm going to tag a release.

bradenmacdonald:
@jcdyer Hmm, not sure how that happened. I have delete that tag as it was obviously incorrect. Please release independently. We'll have to use a slightly different name for the new tag, though.

jcdyer:
Release tagged as v2.6.5patch2

problem_builder/migrations/0004_copy_course_ids.py
feanil: Nice, that will make things a lot simpler, thanks @jcdyer @bradenmacdonald .

feanil:
@jcdyer when you have the edx-platform PR, tag me and @fredsmith on it.  He's on call this week but I'll be shepherding this through.

setup.py
```diff
@@ -73,7 +73,7 @@ def package_data(pkg, root_list):
     name='xblock-problem-builder',
     version='2.7.2',
     description='XBlock - Problem Builder',
-    packages=['problem_builder', 'problem_builder.v1'],
+    packages=['problem_builder', 'problem_builder.v1', 'problem_builder.management', 'problem_builder.management.commands'],
```
bradenmacdonald: Tip: It's even better to just [use `find_packages()`](https://github.com/open-craft/xblock-group-project-v2/commit/2d4406499cdda314e03d329991244cb0b7e974ae#diff-2eeaed663bd0d25b7e608891384b7298)

jcdyer:
@bradenmacdonald  This is ready for a final review before merging to master.  I had to remove the tests because the master branch expects course_key to be non-null, which makes them useless.  The reason I'm keeping the management command around is to give it a presence in the history of the master branch.

bradenmacdonald:
👍  Thanks @jcdyer !

PR #158, opened 2017-08-13
Bump version to 2.7.1.

bradmerlin:
Includes changes from [MCKIN-5248](https://edx-wiki.atlassian.net/browse/MCKIN-5248)

PR #161, opened 2017-08-16
OC-2959 freeform and MCQ submitted values are a json and not an array

cgopalan:
Originally the submission was of the following form:
- For MCQ (multiple-choice questions) it was string value.
- For long/free form answers, it was a json in an array [{<key>: <answer string>}]

This PR makes the submission for both MCQs and freeform answers to the same format, a json with a 'value' key ({"value": <answer string>}). This is so that we can add other info if needed in future.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -11,7 +11,7 @@ dependencies:
     - "pip install -r requirements.txt"
     - "pip install -r $VIRTUAL_ENV/src/xblock-sdk/requirements/base.txt"
     - "pip install -r $VIRTUAL_ENV/src/xblock-sdk/requirements/test.txt"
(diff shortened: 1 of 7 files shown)


bradenmacdonald:
@cgopalan Can you please also include an update to the documentation, https://github.com/open-craft/problem-builder/blob/master/doc/Native%20APIs.md , in this PR?

Code looks good - if you add the doc update, I will test and give +1 tomorrow.

bradenmacdonald:
@cgopalan Please add a version bump commit (like [this](https://github.com/open-craft/problem-builder/pull/158/files)), make sure the build is still green, then merge :)

PR #162, opened 2017-08-25
define ways to filter and map xblock features into build_user_state_data

bradmerlin:
This PR adds a way to allow developers to manually control what fields eventually appear in `student_view_user_state` via `build_user_state_data` by specifying attributes on XBlocks that inherit from `StudentViewUserStateMixin`

**JIRA tickets**: [OC-2958](https://tasks.opencraft.com/browse/OC-2958)

**Dependencies**: None

**Merge deadline**: None

**Testing instructions**:

**student_view_user_state**
1. When viewing an XBlock's `student_view_user_state` (e.g. http://localhost:8000/courses/<course_id>/xblock/<xblock_usage_id>/handler/student_view_user_state), modify fields by adding field names to the block's `USER_STATE_FIELDS` or field name -> function to `USER_STATE_TRANSFORMS`
2. See them appear in the JSON output

**student_view_data**
1. When viewing a course's `student_view_data`, notice that `block_id` is included.

**Reviewers**
- [ ] @bradenmacdonald

### DIFF
--- a/doc/Native APIs.md
+++ b/doc/Native APIs.md
@@ -25,6 +25,7 @@ Problem Builder (`problem-builder`)
 
 ### `student_view_data`
 
+- `block_id`: (string) The XBlock's usage ID.
 - `max_attempts`: (integer) Max number of allowed attempts.
 - `extended_feedback`: (boolean) `true` if extended feedback is enabled for this
   block.
@@ -101,6 +102,7 @@ Step Builder (`step-builder`)
 
 ### `student_view_data`
 
+- `block_id`: (string) The XBlock's usage ID.
 - `title`: (string) The display name of the component.
 - `show_title`: (boolean) `true` if the title should be displayed.
 - `weight`: (float) The weight of the problem.
@@ -158,6 +160,7 @@ Mentoring Step (`sb-step`)
 
 ### `student_view_data`
 
+- `block_id`: (string) The XBlock's usage ID.
 - `type`: (string) Always equals `"sb-step"` for Mentoring Step components.
 - `title`: (string) Step component's display name.
 - `show_title`: (boolean) `true` if the title should be displayed.
@@ -180,6 +183,7 @@ Review Step (`sb-review-step`)
 
 ### `student_view_data`
 
+- `block_id`: (string) The XBlock's usage ID.
 - `type`: (string) Always equals `"sb-review-step`" for Review Step components.
 - `title`: (string) Display name of the component.
 - `components`: (array) A list of `student_view_data` output of all immediate
@@ -193,6 +197,7 @@ Conditional Message component is always child of a Review Step component.
 
 ### `student_view_data`
 
+- `block_id`: (string) The XBlock's usage ID.
 - `type`: (string) Always equals `"sb-conditional-message"` for Conditional
   Message components.
 - `content`: (string) Content of the message. May contain HTML.
@@ -205,6 +210,7 @@ Score Summary (`sb-review-score`)
 
 ### `student_view_data`
 
+- `block_id`: (string) The XBlock's usage ID.
 - `type`: (string) Always equals `"sb-review-score"` for Score Summary
   components.
 
@@ -213,6 +219,7 @@ Per-Question Feedback (`sb-review-per-question-feedback`)
 
 ### `student_view_data`
 
+- `block_id`: (string) The XBlock's usage ID.
 - `type`: (string) Always equals `"sb-review-per-question-feedback"` for Score
   Summary components.
 
@@ -221,6 +228,7 @@ Long Answer (`pb-answer`)
 
 ### `student_view_data`
 
+- `block_id`: (string) The XBlock's usage ID.
 - `type`: (string) Always equals `"pb-answer"` for Long Answer components.
 - `id`: (string) Unique ID (name) of the component.
 - `weight`: (float) The weight of this component.
@@ -257,6 +265,7 @@ Multiple Choice Question (`pb-mcq`)
 
 ### `student_view_data`
 
+- `block_id`: (string) The XBlock's usage ID.
 - `type`: (string) Always equals `"pb-mcq"` for MCQ components.
 - `id`: (string) Unique ID (name) of the component.
 - `question`: (string) The content of the question.
@@ -294,7 +303,6 @@ Each entry in the `tips` array contains these values:
 - `weight`: (float) Child component's weight attribute.
 - `submission`: (string) The value of the choice that the user selected.
 - `message`: (string) General feedback. May contain HTML.
-- `tips`: (string) HTML representation of tips. May be `null`.
 
 ### POST Submit Data
 
@@ -308,6 +316,7 @@ Rating Question (`pb-rating`)
 
 ### `student_view_data`
 
+- `block_id`: (string) The XBlock's usage ID.
 Identical to [MCQ questions](#multiple-choice-question-pb-mcq) except that the
 `type` field always equals `"pb-rating"`.
 
@@ -329,6 +338,7 @@ Multiple Response Question (`pb-mrq`)
 
 ### `student_view_data`
 
+- `block_id`: (string) The XBlock's usage ID.
 - `type`: (string) Always equals `"pb-mrq"` for Multiple Response Question
   components.
 - `id`: (string) Unique ID (name) of the component.
@@ -341,6 +351,14 @@ Multiple Response Question (`pb-mrq`)
   choices. See below for more info.
 - `tips`: (array) A list of objects providing info about tips defined for the
   problem. See below for more info.
+
+#### `tips`
+
+Each entry in the `tips` array contains these values:
+
+- `content`: (string) The text content of the tip.
+- `for_choices`: (array) A list of string values corresponding to choices to
+  which this tip applies to.
 
 ### `student_view_user_state`
 
@@ -365,7 +383,6 @@ Each item in the `choices` array contains these fields:
 - `completed`: (boolean) Boolean indicating whether the state of the choice is
   correct.
 - `selected`: (boolean) `true` if the user selected this choice.
-- `tips`: (string) Tips formatted as a string of HTML.
 - `value`: (string) The value of the choice.
 
 ### POST Submit Data
@@ -378,6 +395,7 @@ Ranged Value Slider (`pb-slider`)
 
 ### `student_view_data`
 
+- `block_id`: (string) The XBlock's usage ID.
 - `type`: (string) Always equals `"pb-slider"` for Ranged Value Slider
   components.
 - `id`: (string) Unique ID (name) of the component.
@@ -409,6 +427,7 @@ Completion (`pb-completion`)
 
 ### `student_view_data`
 
+- `block_id`: (string) The XBlock's usage ID.
 - `type`: (string) Always equals `"pb-completion"` for Completion components.
 - `id`: (string) Unique ID (name) of the component.
 - `title`: (string) Display name of the problem.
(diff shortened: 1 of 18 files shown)


problem_builder/answer.py
```diff
@@ -53,6 +52,11 @@ class AnswerMixin(XBlockWithPreviewMixin, XBlockWithTranslationServiceMixin, Stu
     """
     Mixin to give an XBlock the ability to read/write data to the Answers DB table.
     """
+    USER_STATE_FIELDS = ['edxnotes_visibility', ]
```
bradenmacdonald: Please put a comment above: `# Define fields returned in student_view_user_state`

problem_builder/answer.py
```diff
@@ -53,6 +52,11 @@ class AnswerMixin(XBlockWithPreviewMixin, XBlockWithTranslationServiceMixin, Stu
...
+    USER_STATE_TRANSFORMS = {
+        'answer_data': lambda self: self.get_answer_data()
```
bradenmacdonald: Could we change this API to remove the need for a lambda function - so we can just write `'answer_data': self.get_answer_data` ?

problem_builder/answer.py
```diff
@@ -86,6 +90,14 @@ def get_model_object(self, name=None):
...
+    def get_answer_data(self):
+        answer_data = self.get_model_object()
+        return {
+            "student_input": answer_data.student_input,
+            "created_on": answer_data.created_on,
+            "modified_on": answer_data.modified_on,
+        }
+
     @property
     def student_input(self):
```
bradenmacdonald: This method should be removed, I think - not sure why it's here, as it conflicts with https://github.com/open-craft/problem-builder/blob/de315bb040008554be3e509ae0fc82b094c15ff0/problem_builder/answer.py#L171-L188 which is more correct.

problem_builder/answer.py
bradenmacdonald: What is this `edxnotes_visibility` field, and why do we care about it?

problem_builder/answer.py
```diff
@@ -86,6 +90,14 @@ def get_model_object(self, name=None):
 
         return answer_data
 
+    def get_answer_data(self):
+        answer_data = self.get_model_object()
+        return {
+            "student_input": answer_data.student_input,
```
bradenmacdonald: This isn't always right - we should return `self.student_input` instead which uses https://github.com/open-craft/problem-builder/blob/de315bb040008554be3e509ae0fc82b094c15ff0/problem_builder/answer.py#L171-L188 , the correct `student_input` algorithm.

problem_builder/answer.py
bradenmacdonald: Also, we don't return `created_on` nor `modified_on` for any other component types (MCQ, MRQ, etc.) so I think we should just get rid of `get_answer_data` and instead just put

```python
USER_STATE_TRANSFORMS = {
    'student_input': self.student_input
}
```

problem_builder/mrq.py
```diff
@@ -19,8 +19,13 @@
 #
 
 # Imports ###########################################################
+import json
 
 import logging
+from problem_builder.tests.unit.utils import DateTimeEncoder
+
+import webob
+from xblock.core import XBlock
```
bradenmacdonald: ^ You added a bunch of imports here that don't seem to be used?

problem_builder/mixins.py
```diff
@@ -185,21 +185,38 @@ def studio_view(self, _context=None):
...
         result = {}
         for _, field in self.fields.iteritems():
-            if field.scope in self.INCLUDE_SCOPES:
+            # Only insert fields if their scopes and field names match
+            if field.scope in self.INCLUDE_SCOPES and field.name in self.USER_STATE_FIELDS:
                 result[field.name] = field.read_from(self)
 
+        # Add fields according to transforms dict
+        for key, func in self.USER_STATE_TRANSFORMS.iteritems():
+            result[key] = func(self)
```
e-kolpakov: @bradmerlin nit: Overwriting values obtained from field is slightly suboptimal and might be a source of WTFs in the future. I'd suggest to merge it together with original loop:

```
for _, field in self.fields.iteritems():
    if scope in INCLUDE_SCOPES and name in INCLUDE_FIELDS:
        field_reader = USER_STATE_TRANSFORMS.get(field.name, field.read_from)
        result[field.name] = field_reader(self)
```

Also, (and I know naming is hard, so no pressure :)), but those `USER_STATE_TRANSFORMS` is misleading as those functions semantically are not transformers, but readers:

* Transformer takes one values and transforms it to some other - i.e. it takes field value and maps it to something else
* Reader takes an object that contains some value and returns that value - i.e. it takes XBlock and return a value (not necessarily read from a field)

So it might make sense to rename the collection as well.

problem_builder/answer.py
bradmerlin: @bradenmacdonald From the attr definition:
> Indicates whether Student Notes are visible in the course. Students can also show or hide their notes in the courseware.

It appeared everywhere before I started working on this PR. Kept it for compatibility reasons, but will remove.

bradenmacdonald:
@bradmerlin Can you please look into why so many of the integration tests have failed on CircleCI?

problem_builder/mentoring.py
```diff
@@ -239,6 +239,7 @@ class MentoringBlock(BaseMentoringBlock, StudioContainerWithNestedXBlocksMixin,
     ok to continue.
     """
     # Content
+    USER_STATE_FIELDS = ['active_step', 'num_attempts']
```
bradenmacdonald: `active_step` should not be exposed here - only on `MentoringWithExplicitStepsBlock`. Use of `MentoringBlock` itself with steps (mode = 'assessment') is deprecated and we don't need to support it.

problem_builder/mentoring.py
```diff
@@ -963,6 +967,11 @@ class MentoringWithExplicitStepsBlock(BaseMentoringBlock, StudioContainerWithNes
 
     editable_fields = ('display_name', 'max_attempts', 'extended_feedback', 'weight')
 
+    def build_user_state_data(self, context=None):
+        user_state_data = super(MentoringWithExplicitStepsBlock, self).build_user_state_data()
+        user_state_data['active_step'] = self.active_step_safe
```
bradenmacdonald: Do we need `active_step` in `USER_STATE_FIELDS` if we're setting it here explicitly anyways? Isn't that wasting cycles to set it then overwrite it?

problem_builder/mixins.py
```diff
@@ -185,19 +184,28 @@ def studio_view(self, _context=None):
...
+
+    To prevent unnecessary overloading of the build_user_state_data method,
+    you may specify `USER_STATE_FIELDS` to customise build_user_state_data
+    and student_view_user_state output.
     """
     NESTED_BLOCKS_KEY = "components"
     INCLUDE_SCOPES = (Scope.user_state, Scope.user_info, Scope.preferences)
 
+    # Default field names to include in build_user_state_data
+    USER_STATE_FIELDS = ['completed', 'num_attempts', 'student_results']
```
bradenmacdonald: I think it could potentially be confusing to have a set of default fields like this.

problem_builder/mrq.py
```diff
@@ -121,17 +120,15 @@ def calculate_results(self, submissions):
...
             choice_selected = choice.value in submissions
 
             if choice.value in self.required_choices:
                 if not choice_selected:
                     choice_completed = False
             elif choice_selected and choice.value not in self.ignored_choices:
                 choice_completed = False
-            for tip in self.get_tips():
-                if choice.value in tip.values:
-                    choice_tips_html.append(tip.render('mentoring_view').content)
```
bradenmacdonald: You have removed `tip.render('mentoring_view').content` and `loader.render_template('templates/html/tip_choice_group.html', ...)` from here, but you did not add that back in anywhere else. That's probably why so many integration tests etc. are failing. You must preserve the existing functionality in the student_view and the corresponding HTML (i.e. when you test this XBlock in the LMS/Studio, tips should still be working just as they always are).

When using this XBlock via it's student_view, at some point the tips must be converted to HTML using those templates. You can do that here (as was the case), elsewhere in python, or even in the overall HTML template, but it must be done somewhere.

It's only student_view_data where we definitely want none of this HTML wrapping the tips.

bradmerlin: @bradenmacdonald Ya I realised this over the weekend. I misunderstood your [reply](https://tasks.opencraft.com/browse/OC-2958?focusedCommentId=45063&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-45063) and assumed that the HTML wasn't important anywhere.

problem_builder/mixins.py
```diff
@@ -221,3 +238,57 @@ def student_view_user_state(self, context=None, suffix=''):
...
+        Applies `student_results_transforms()` output to current `student_results`
+        """
+        current_student_results = dict(student_results)
+        transformed_student_results = []
+        for question in self.steps:
+            if question.name in current_student_results:
+                question_results = question.get_results(current_student_results[question.name])
+                for key, f in self.student_results_transforms():
+                    if key in question_results:
+                        question_results[key] = f(question_results, question)
```
bradenmacdonald: Nit: 

I would personally find this code easier to follow if you replace the above `for` loop with
```python
if 'choices' in question_results:
    question_results['choices'] = self.transform_choices(question_results, question)
if 'tip' in question_results:
    question_results['tips'] = self.transform_tips(question)
```
Which would also remove 9 lines of code. You could also change `transform_choices` so it won't throw an error if `result['choices']` is not a valid key, and then you can get rid of the `if` statements and make this only two lines.

doc/Native APIs.md
```diff
@@ -365,9 +393,19 @@ Each item in the `choices` array contains these fields:
 - `completed`: (boolean) Boolean indicating whether the state of the choice is
   correct.
 - `selected`: (boolean) `true` if the user selected this choice.
-- `tips`: (string) Tips formatted as a string of HTML.
+- `tips`: (array) A list of objects providing info about tips defined for the
+  problem. See below for more info.
```
bradenmacdonald: I don't think we need the `tips` in the `user_state` since they're already in the `student_view_data`. We should have one consistent place to find them, and since they're not user-specific, the `student_view_data` makes more sense.

problem_builder/mixins.py
```diff
@@ -221,3 +238,57 @@ def student_view_user_state(self, context=None, suffix=''):
...
+                transformed_student_results.append([question.name, question_results])
+        return transformed_student_results
+
+    def transform_tips(self, result, question):
+        """
+        For MCQ, tips are at `student_results` root level.
+        """
+        return [tip.student_view_data() for tip in question.get_tips() if question.student_choice in tip.values]
```
bradenmacdonald: Actually, as mentioned elsewhere, these tips are not user-specific so I think we should only include them in `student_view_data`, not also here in the `user_state`.

bradmerlin: @bradenmacdonald  Is it not enough that we only output tips that are linked to the user's choice here?

bradenmacdonald: @bradmerlin  Yeah, it's not really a problem, but it doesn't seem useful to me, since the API consumer will already know what the relevant tips are, since it knows all the tips and it knows the learner's selected answer. So I think this is redundant.

bradenmacdonald:
@bradmerlin The CI build is failing - if you rebase on master it should be fixed, due to a change in `circle.yml` that we've since made.

bradmerlin:
@bradenmacdonald commits squashed into one commit over master, and removed tips from `student_view_user_state`. Can I merge?

PR #163, opened 2017-09-02
Support for MRQ, user ID and user e-mail in Instructor Tool

BogdanL:
This extends the Instructor Tool module to:
* Add support for MRQ in the HTML and CSV reports.
* Add user ID and user email columns to the CSV report.

**JIRA ticket**:  [OC-2997](https://tasks.opencraft.com/browse/OC-2997)

**Testing instructions**:
1. Add a couple of MRQ in a problem builder xblock
2. Submit answer(s) with a few users
3. Verify in the Instructor Tool that MRQs are filterable in the dropdown 
4. Make sure MRQs are displayed in the HTML table and included in the CSV export
5. Make sure user IDs and user emails are included in the CSV export but not in the HTML table

**Considerations**:
When choosing a MCQ filter in the Instructor Tool, also Rating questions (a subclass of MCQ) are listed. I'm assuming this is desired behaviour because technically ratings are also MCQ?

### DIFF
--- a/problem_builder/instructor_tool.py
+++ b/problem_builder/instructor_tool.py
@@ -135,6 +135,7 @@ class InstructorToolBlock(XBlock):
             return Fragment(u'<p>This interface can only be used by course staff.</p>')
         block_choices = {
             _('Multiple Choice Question'): 'MCQBlock',
+            _('Multiple Response Question'): 'MRQBlock',
             _('Rating Question'): 'RatingBlock',
             _('Long Answer'): 'AnswerBlock',
         }
--- a/problem_builder/mrq.py
+++ b/problem_builder/mrq.py
@@ -26,8 +26,9 @@ from xblock.fields import List, Scope, Boolean, String
 from xblock.validation import ValidationMessage
 
 from problem_builder.mixins import StudentViewUserStateMixin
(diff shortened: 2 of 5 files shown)


bradmerlin:
:+1:

- [x] I tested this: 
* I verified in the Instructor Tool that the MRQ filter in the dropdown exists and works
* I made sure that MRQs are displayed in the HTML table and included in the CSV export
* I made sure that user IDs and user emails are included in the CSV export but not in the HTML table
- [x] I read through the code
- [ ] Includes documentation: If you're keen, it would be nice to have some documentation about how to use the instructor tool (for course creators). For example, I had to ask about adding `pb-instructor-tool` to Advanced Modules.

bradmerlin:
@BogdanL please squash commits

bradenmacdonald:
@bradmerlin 

> it would be nice to have some documentation about how to use the instructor tool (for course creators). For example, I had to ask about adding pb-instructor-tool to Advanced Modules.

FYI I've now added the Instructor Tool to the Problem Builder demo course we have at https://github.com/open-craft/demo-courses which should make it a bit easier to get started with Problem Builder and the Instructor Tool.

PR #165, opened 2017-09-16
Student view data with display names

e-kolpakov:
**Description:** This PR adds `display_name` field to `student_view_data`.

**Author concerns:**

1. `student_view_data` is added as a method individually to each XBlock that needs it - there are no common base method - not fixing this due to being out of scope and lack of time.
1. `student_view_data` test coverage is quite superfluous - just a handful of blocks are checked - not fixing this due to being out of scope and lack of time.
1. Some XBlocks already provided `display_name` field via `title` key - not changing `title` as it is likely used by the clients and still adding `display_name` to provide a common denominator to all XBlocks.

**Merge deadline:** Monday, 2017-09-18.

**Testing:**

0. Install this version
1. Create a course with Problem builders (or just import from OpenCraft demo courses repo)
2. Read from LMS course content API: http://localhost:8000/api/courses/v1/blocks/block-v1%3AOrg1%2BPB1%2BT1%2Btype%40problem-builder%2Bblock%40979b4a84326941309413f3347e5b5d75/?depth=all&fields=student_view_data&username=staff&student_view_data=all
3. Make sure that `student_view_data` field contains `display_name` for all XBlocks that have it defined.
4. Make sure that documentation (NativeAPI.md) is updated for all XBlock that have display_name added to student_view_data

### DIFF
--- a/doc/Native APIs.md
+++ b/doc/Native APIs.md
@@ -26,6 +26,7 @@ Problem Builder (`problem-builder`)
 ### `student_view_data`
 
 - `block_id`: (string) The XBlock's usage ID.
+- `display_name`: (string) The XBlock's display name.
 - `max_attempts`: (integer) Max number of allowed attempts.
 - `extended_feedback`: (boolean) `true` if extended feedback is enabled for this
   block.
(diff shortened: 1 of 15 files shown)


e-kolpakov:
@bradenmacdonald Tests seem to fail due to unrelated reason - probably an intermittent DB failure. I've restarted the build.

PR #166, opened 2017-09-19
Remove course_id from answer model.

jcdyer:
No migration added, per request from @feanil 

OC-3033


bradenmacdonald:
@jcdyer BTW, do you know why this PR doesn't cause the "Your models have changes that are not yet reflected in a migration" error?

jcdyer:
Models are allowed to reflect a partial view of a table, so having extra fields and indexes (as long as the fields are either nullable or have a default value) is perfectly fine.

PR #167, opened 2017-09-19
Migration to remove deprecated course_id from db.

jcdyer:
Follow up PR to add migrations to remove course_id (and related index).

Depends on #166.

OC-3033


bradenmacdonald:
@jcdyer Do you think we could also remove the management commands in this second PR? Because they won't work once the `course_id` column is gone.

PR #170, opened 2017-09-29
Not quite functional implementation of tox testing

bmedx:
I was able to get all but ~12 tests working locally by upgrading the version of Selenium and geckodriver along with these changes. I'm not sure if the remaining tests are breaking due to the Firefox/Selenium upgrade or something else, but maybe this can be a starting point if someone more familiar with Problem Builder wants to pick up the work. 
cc: @bradenmacdonald

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@
 .idea/*
 dump.rdb
 problem_builder.tests.*
+.tox/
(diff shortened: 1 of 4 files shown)


dupainaulevain:
@bmedx I'm continuing your work and two years later the environment improved and it is apparently easier to move the test logic to tox. If you're interested, this is happening at https://github.com/open-craft/problem-builder/pull/228

dupainaulevain:
closing in favor of #228

PR #172, opened 2017-10-06
Add SwipeBlock for swipeable binary choice questions

kaizoku:
This PR creates a new component type for Binary Choice Questions, which will be utilized for a swiping mobile interface.

**JIRA tickets**: Implements OC-3175, MCKIN-5975

**Testing instructions**:

1. Add a new problem builder component.
2. Add a new "Swipable Binary Choice Question" sub-component
3. Add a two possible choices, and mark one as correct.

**Reviewers**
- [ ] (mtyaka)

**Settings**
Under `/admin/xblock_django/xblockconfiguration`, an option named "pb-swipe" must be created and enabled.

### DIFF
--- a/problem_builder/swipe.py
+++ b/problem_builder/swipe.py
@@ -1,0 +1,171 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2014-2015 Harvard, edX & OpenCraft
+#
+# This software's license gives you freedom; you can copy, convey,
+# propagate, redistribute and/or modify this program under the terms of
+# the GNU Affero General Public License (AGPL) as published by the Free
+# Software Foundation (FSF), either version 3 of the License, or (at your
+# option) any later version of the AGPL published by the FSF.
+#
+# This program is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU Affero
+# General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program in a file in the toplevel directory called
+# "AGPLv3".  If not, see <http://www.gnu.org/licenses/>.
+#
+
+# Imports ###########################################################
+import logging
+
+from xblock.fields import Scope, String
+from xblock.validation import ValidationMessage
+from xblockutils.resources import ResourceLoader
+
+from .mixins import StudentViewUserStateMixin
+from .questionnaire import QuestionnaireAbstractBlock
+from .sub_api import SubmittingXBlockMixin
+
+
+# Globals ###########################################################
+log = logging.getLogger(__name__)
+loader = ResourceLoader(__name__)
+
+
+# Make '_' a no-op so we can scrape strings
+def _(text):
+    return text
+
+
+# Classes ###########################################################
+class SwipeBlock(SubmittingXBlockMixin, StudentViewUserStateMixin, QuestionnaireAbstractBlock):
+    """
+    An XBlock used to ask binary-choice questions with a swiping interface
+    """
+    CATEGORY = 'pb-swipe'
+    STUDIO_LABEL = _(u"Swipeable Binary Choice Question")
+    USER_STATE_FIELDS = ['num_attempts', 'student_choice']
+
+    message = String(
+        display_name=_("Message"),
+        help=_(
+            "General feedback provided when submitting. "
+            "(This is not shown if there is a more specific feedback tip for the choice selected by the learner.)"
+        ),
+        scope=Scope.content,
+        default=""
+    )
+
+    student_choice = String(
+        # {Last input submitted by the student
+        default="",
+        scope=Scope.user_state,
+    )
+
+    correct_choice = String(
+        display_name=_("Correct Choice"),
+        help=_("Specify the value that students may select for this question to be considered correct."),
+        scope=Scope.content,
+        values_provider=QuestionnaireAbstractBlock.choice_values_provider,
+    )
+    img_url = String(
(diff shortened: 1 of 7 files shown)


problem_builder/swipe.py
```diff
@@ -0,0 +1,151 @@
...
+    def validate_field_data(self, validation, data):
+        """
+        Validate this block's field data.
+        """
+        super(SwipeBlock, self).validate_field_data(validation, data)
+
+        def add_error(msg):
+            validation.add(ValidationMessage(ValidationMessage.ERROR, msg))
+
+        if len(self.all_choice_values) != 2:
```
mtyaka: You should not add an error if there are no choices defined yet. Otherwise if you add a new swipe block in studio, and try to edit its attributes before adding the choices, it will fail to save.

problem_builder/swipe.py
```diff
@@ -0,0 +1,151 @@
...
+            add_error(
+                self._(u"You must have exactly two choices.")
+            )
+        if not data.correct_choice:
```
mtyaka: Same here, this should not add errors if no choices have been defined yet.

problem_builder/templates/html/swipeblock.html
```diff
@@ -0,0 +1,30 @@
+{% load i18n %}
+{% if not hide_header %}
+<h4 class="question-title" id="heading_{{ self.html_id }}">{{ self.display_name_with_default }}</h4>
+{% endif %}
+<fieldset class="choices questionnaire" id="{{ self.html_id }}">
+  <legend class="question field-group-hd">{{ self.question|safe }}</legend>
```
mtyaka: Nit: It would be good to include an `` element with the `src` attribute set to the expanded `img_url`.

problem_builder/templates/html/swipeblock.html
```diff
@@ -0,0 +1,31 @@
+{% load i18n %}
+{% if not hide_header %}
+<h4 class="question-title" id="heading_{{ self.html_id }}">{{ self.display_name_with_default }}</h4>
+{% endif %}
+<img src={{ self.expand_static_url(self.img_url) }}
```
mtyaka: This does not work; I get "Error: Variables and attributes may not begin with underscores: 'self._expand_static_url'"

mtyaka: Also missing end of `<img` tag.

mtyaka: You will also have to define an `alt` attribute: http://edx.readthedocs.io/projects/edx-developer-guide/en/latest/conventions/accessibility.html#alt-attributes

problem_builder/swipe.py
```diff
@@ -0,0 +1,164 @@
...
+            'weight': self.weight,
+            'score': 1 if correct else 0,
+        }
+
+    def get_last_result(self):
+        return self.get_results({'submission': self.student_choice}) if self.student_choice else {}
+
+    def submit(self, submission):
+        log.debug(u'Received Swipe submission: "%s"', submission)
+        result = self.calculate_results(submission['value'])
```
mtyaka: I get "'SwipeBlockWithMixins' object has no attribute 'calculate_results'" when I try to submit an answer.

problem_builder/templates/html/swipeblock.html
```diff
@@ -0,0 +1,33 @@
+{% load i18n %}
+{% if not hide_header %}
+<h4 class="question-title" id="heading_{{ self.html_id }}">{{ self.display_name_with_default }}</h4>
+{% endif %}
+{% if self.img_url.strip %}
+<img src={{ self.expand_img_url }} alt="" />
```
mtyaka: This still doesn't work. I think that's because it's missing `"` around the src value.

problem_builder/swipe.py
```diff
@@ -0,0 +1,171 @@
...
+        return {
+            'submission': submission,
+            'message': self.message_formatted,
+            'status': 'correct' if correct else 'incorrect',
+            'weight': self.weight,
+            'score': 1 if correct else 0,
+        }
+
+    def get_results(self, previous_result):
+        return self.calculate_results(self, previous_result['submission'])
```
mtyaka: Submitting the answer now works, but I'm getting "TypeError: calculate_results() takes exactly 2 arguments (3 given)" if I reload the page in the LMS after an answer has been submitted.

It should be `calculate_results(previous_result['submission'])` instead of `calculate_results(self, ...)`.

problem_builder/templates/html/swipeblock.html
```diff
@@ -0,0 +1,33 @@
+{% load i18n %}
+{% if not hide_header %}
+<h4 class="question-title" id="heading_{{ self.html_id }}">{{ self.display_name_with_default }}</h4>
+{% endif %}
+{% if self.img_url.strip %}
+<img src="{{ self.expand_img_url }}" alt="" />
```
mtyaka: @kaizoku This still doesn't work correctly. I think it's because the property is named `expanded_img_url`, not `expand_img_url`.

mtyaka:
Screenshot of large image in the Studio:

screen shot 2017-10-09 at 15 14 30

kaizoku:
Good idea, I added styling in `problem_builder/public/css/problem-builder.css` which is only loaded in studio, since you specified that's where the issue is.
If we want the styling for both studio and LMS I could move it into `questionnaire.css`, which is loaded for both.

mtyaka:
Thanks! :+1:

- [x] I tested this: I added the new Swipe block to a course, added two choices to it, verified that `student_view_data` and `student_user_state` return sensible results. Verified that the Swipe block looks good in Studio & LMS.
- [x] I read through the code
- [x] Includes documentation

PR #173, opened 2017-11-07
[MCKIN-5975] Make pb-swipe return proper student view data.

UmanShahzad:
This PR fixes some requirement issues with the Swipeable Binary Choice Questions component, which will be utilized for a swiping mobile interface.

**JIRA tickets**: Fixes MCKIN-5975

**Testing instructions**:

1. Add a new problem builder component.
2. Add a new "Swipable Binary Choice Question" sub-component
3. Add two possible choices, and mark one as correct.

**Reviewers**
- [ ] @bradenmacdonald @kaizoku 

**Settings**
Under `/admin/xblock_django/xblockconfiguration`, an option named "pb-swipe" must be created and enabled.


problem_builder/swipe.py
```diff
@@ -60,9 +60,9 @@ class SwipeBlock(SubmittingXBlockMixin, StudentViewUserStateMixin, Questionnaire
     )
 
     student_choice = String(
-        # {Last input submitted by the student
         default="",
         scope=Scope.user_state,
+        help=_("Last input submitted by the student.")
     )
 
     correct_choice = String(
```
bradenmacdonald: This should just be a `Boolean` called `correct`, and it should not have a `values_provider`.

problem_builder/swipe.py
```diff
@@ -60,9 +60,9 @@ class SwipeBlock(SubmittingXBlockMixin, StudentViewUserStateMixin, Questionnaire
     )
```
bradenmacdonald: We want to remove `QuestionnaireAbstractBlock`, because this swipe component shouldn't have children nor "choices", which is what that mixin adds.

We also want to replace the `question` mixin from `QuestionnaireAbstractBlock` with a simple

```python
    text = String(
        display_name=_("Text"),
        help=_("Text to display on this card. The student must determine if this statement is true or false."),
        scope=Scope.content,
        default="",
        multiline_editor=True,
    )
```

UmanShahzad: Ah, I see.

problem_builder/swipe.py
UmanShahzad: :+1:

UmanShahzad:
Sorry, I'm still fixing this. Misunderstood something.

UmanShahzad:
@bradenmacdonald Should be good now!

kaizoku:
This combination of mixins throws a traceback in the studio as soon as it's added, preventing the problem from being edited.

After adding SwipeBlock:
screenshot-2017-11-9 problem builder problem builder edx demonstration course studio

After navigating back to the unit:
screenshot-2017-11-9 unit unit edx demonstration course studio

```
Traceback (most recent call last):
  File "/edx/app/edxapp/edx-platform/cms/djangoapps/contentstore/views/preview.py", line 316, in get_preview_fragment
    fragment = module.render(preview_view, context)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/xblock/core.py", line 202, in render
    return self.runtime.render(self, view, context)
  File "/edx/app/edxapp/edx-platform/common/lib/xmodule/xmodule/x_module.py", line 1895, in render
    return self.__getattr__('render')(block, view_name, context)
  File "/edx/app/edxapp/edx-platform/common/lib/xmodule/xmodule/x_module.py", line 1302, in render
    return super(MetricsMixin, self).render(block, view_name, context=context)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/xblock/runtime.py", line 808, in render
    frag = view_fn(context)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/problem_builder/mixins.py", line 161, in author_view
    return self.mentoring_view(context)
AttributeError: 'SwipeBlockWithMixins' object has no attribute 'mentoring_view'
```

UmanShahzad:
@kaizoku Thanks for checking this out. I can't even get the Swipeable Block to show locally on my problem-builder, despite entering the setting, so it's hard to test this.

UmanShahzad:
@kaizoku Can you do a quick check on the most recent update?

problem_builder/mentoring.py
```diff
@@ -362,13 +362,10 @@ def allowed_nested_blocks(self):
...
-            from xblock_django.models import XBlockConfiguration
-            opt = XBlockConfiguration.objects.filter(name="pb-swipe")
-            if opt.count() and opt.first().enabled:
-                additional_blocks.append(SwipeBlock)
-        except ImportError:
-            pass
+        #from xblock_django.models import XBlockConfiguration
+        #opt = XBlockConfiguration.objects.filter(name="pb-swipe")
+        #if opt.count() and opt.first().enabled:
+        additional_blocks.append(SwipeBlock)
```
UmanShahzad: Sorry, woops, that was for testing.

doc/Native APIs.md
```diff
@@ -402,9 +382,18 @@ Each item in the `choices` array contains these fields:
...
 
 When submitting the problem the data should be a single object containing the
-`"value"` property which has the value of the selected choice.
-Example: `{"value": "blue"}`
+`"value"` property which has a boolean value indicating which swipe the student took:
+
+- Right -> True
+- Left -> False
+
+Example: `{"value": True}` if the student made a *rightwards* swipe.
```
bradenmacdonald: Nit: in JSON, booleans are not capitalized.

problem_builder/swipe.py
```diff
@@ -41,91 +40,59 @@ def _(text):
...
-        help=_(
-            "Specify the URL of an image associated with this question."
-        ),
+        help=_("Specifies whether the card is correct."),
         scope=Scope.content,
-        default=""
     )
-    editable_fields = QuestionnaireAbstractBlock.editable_fields + ('message', 'correct_choice', 'img_url',)
+
+    editable_fields = ('display_name', 'text', 'img_url',)
```
bradenmacdonald: `correct` should also be editable.

bradenmacdonald:
Oh, bad timing :-p
I'll take another look tomorrow.

UmanShahzad:
@bradenmacdonald Yep.. just saw that. Woops!

kaizoku:
Looking good @UmanShahzad. I can add and edit questions now.

By the way, nice simplification of this component. Organizing it this way makes a lot more sense. Sorry you have to fix my work.

I was getting one error when reloading the student_view:
```
  File "/edx/src/problem-builder/problem_builder/mentoring.py", line 656, in _get_standard_results                    
    child_result = child.get_last_result()                                                                            AttributeError: 'SwipeBlockWithMixins' object has no attribute 'get_last_result'                                      
```
Looks like that was another method added from one of the old mixins. I fixed it by adding:
```
    def get_last_result(self):
        return self.calculate_results(self.student_choice)
```

UmanShahzad:
@kaizoku No problem, and I'm grateful for your help on this. I've added back the 2 result functions.

kaizoku:
:+1:

- [✓] I tested this: (installed XBlock, added component, tested basic functionality)
- [✓] I read through the code
- [✓] I checked for accessibility issues
- [✓] Includes documentation

problem_builder/swipe.py
```diff
@@ -136,15 +116,9 @@ def student_view_data(self, context=None):
...
-            'message': self.message,
+            'text': self.text,
             'img_url': self.expand_static_url(self.img_url),
-            'choices': [
-                {'value': choice['value'], 'content': choice['display_name']}
-                for choice in self.human_readable_choices
-            ],
-            'weight': self.weight,
-            'tips': [tip.student_view_data() for tip in self.get_tips()],
+            'correct': self.student_choice == self.correct,
```
bradenmacdonald: This should just be `'correct': self.correct`

(We can't access student-specific fields like `student_choice` in `student_view_data` anyways - only in `student_view_user_state`.)

problem_builder/swipe.py
```diff
@@ -41,53 +43,52 @@ def _(text):
...
     )
-    editable_fields = QuestionnaireAbstractBlock.editable_fields + ('message', 'correct_choice', 'img_url',)
+
+    correct = Boolean(
+        display_name=_("Correct Choice"),
+        help=_("Specifies whether the card is correct."),
+        scope=Scope.content,
+    )
+
+    student_choice = String(
```
bradenmacdonald: This should be a `Boolean` field too. (Causes errors otherwise, when storing a boolean).

problem_builder/swipe.py
```diff
@@ -166,6 +140,20 @@ def expand_static_url(self, url):
...
-    @property
-    def expanded_img_url(self):
-        return self.expand_static_url(self.img_url)
+    def mentoring_view(self, context=None):
+        """ Render the swipe image, text & whether it's correct within a mentoring block question. """
+        return Fragment(
+            (
+                u'<img src="{img_url}" style="max-width: 100%;" />'
+                u'<p class="swipe-text">"{text}"</p>'
+                u'<p>{answer}</p>'
```
bradenmacdonald: We shouldn't show the answer in the `mentoring_view` since that's the view that students would see (in theory). You could display `student_choice` here if you wanted (either "no choice made", "Selected True", or "Selected False"), or nothing (just text and image is fine).

UmanShahzad: Ah right! I'll just remove the answer to keep it simple, then.

problem_builder/swipe.py
UmanShahzad: Ah, ok! :+1:

problem_builder/swipe.py
UmanShahzad: Woops, that's right.

PR #175, opened 2018-01-04
Emit interactions

e-kolpakov:
**Description:** This PR makes use of [Apros API](https://github.com/mckinseyacademy/mcka_apros/pull/1834) to submit interactions to SCORM shell

Mechanism employed s pretty generic (submitting JS XBlock event) and not tied to Apros.

**Dependecies:** Self-contained, but effect can be only seen by applying https://github.com/mckinseyacademy/mcka_apros/pull/1834

**Testing instructions:**
0. Create a course with Step Builder and have a review step in it. Make sure some questions are graded to test submitting correct grade, and number of attempts is limited.
1. In Apros complete the problem and go to review step
1.1. Expected result: [this line](https://github.com/mckinseyacademy/mcka_apros/pull/1834/files#diff-cc7ad4bcad263f3c78f2d38e62d5be9cR300) is hit and `interaction_data` carries correct values
1.2. Make sure it is only hit *once*
2. If local SCORM player is setup - make sure `cmi.interactions` collection is populated with data submitted - this is more of an integration test of entire feature, including this PR, https://github.com/mckinseyacademy/mcka_apros/pull/1834, https://github.com/mckinseyacademy/mcka_apros/pull/1813 and https://github.com/mckinseyacademy/scorm-shell/pull/1

### DIFF
--- a/problem_builder/public/js/mentoring_with_steps.js
+++ b/problem_builder/public/js/mentoring_with_steps.js
@@ -200,6 +200,7 @@ function MentoringWithStepsBlock(runtime, element) {
             // Reinstate default event handlers
             nextDOM.off('click');
             nextDOM.on('click', updateDisplay);
+            reviewButtonDOM.off('click');
             reviewButtonDOM.on('click', showGrade);
 
             var step = getActiveStep();
@@ -368,7 +369,7 @@ function MentoringWithStepsBlock(runtime, element) {
         var attemptsData = attemptsDOM.data();
 
         return {
-            "attempts_count": attemptsData.num_used,
+            "attempts_count": attemptsData.num_attempts,
             "attempts_max": attemptsData.max_attempts || "unlimited",
             "score": reviewStepDOM.find(".grade-result").data('score')
         }
@@ -489,6 +490,7 @@ function MentoringWithStepsBlock(runtime, element) {
         }
 
         reviewButtonDOM = $(element).find('.submit .input-review');
+        reviewButtonDOM.off('click');
         reviewButtonDOM.on('click', showGrade);
 
         tryAgainDOM = $(element).find('.submit .input-try-again');
@@ -498,6 +500,7 @@ function MentoringWithStepsBlock(runtime, element) {
         attemptsDOM = $('.attempts', element);
 
         reviewLinkDOM = $(element).find('.review-link');
+        reviewLinkDOM.off('click');
         reviewLinkDOM.on('click', showGrade);
 
         // Add click handler that takes care of links to steps on the extended review:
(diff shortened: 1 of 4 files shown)


tomaszgy:
:+1: given that the tests and documentation are added :)

- [x] I tested this:
1. Created a course with StepBuilder as described in testing instructions.
2. Ran it inside Moodle and tested what happens when user clicks `Review grade`:
a) saw [this line](https://github.com/mckinseyacademy/mcka_apros/pull/1834/files#diff-cc7ad4bcad263f3c78f2d38e62d5be9cR300) being hit once with correct data inside `interaction_data`
b) saw `cmi.interactions` being populated (an item is added to the collection through [setInteractionData](https://github.com/mckinseyacademy/scorm-shell/blob/master/template/scormdriver/indexAPI.html#L75-L84) function)
- [x] I read through the code
- [x] I checked for accessibility issues - it doesn't seem to be directly connected to accessibility
- [ ] Includes documentation - I'd add it in one place

EXTRA: would be nice to have tests here and docs (please see a comment in the code)

problem_builder/public/js/mentoring_with_steps.js
```diff
@@ -364,11 +365,22 @@ function MentoringWithStepsBlock(runtime, element) {
         });
     }
 
+    function buildInteractionData() {
```
tomaszgy: Shall we add a short description of what is this data for?

e-kolpakov: @tomaszgy I can't think of a better description rather than "this data is used in `xblock.interaction`" event". The thing is - the mechanism is pretty generic and actual use of data is left for XBlock runtime to determine - and the runtime isn't here. So, in my opinion, it's not a good idea to describe how this data is used by the runtime (since there might be different uses across runtimes).

tomaszgy:
@e-kolpakov, thanks sure, I'll approve it before Monday morning!

Just in case, please also note [the comment in OC task](https://tasks.opencraft.com/browse/OC-3366?focusedCommentId=56919&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-56919).

e-kolpakov:
@tomaszgy Answered there, but basically this is outside of scope of this PR - it actually belongs to https://github.com/mckinseyacademy/scorm-shell/pull/1

problem_builder/public/js/mentoring_with_steps.js
tomaszgy: Thanks - the current comments are very good :)

tomaszgy:
@e-kolpakov, just checked everything here - looks very good! :)

:+1: to merge it

PR #176, opened 2018-01-23
Fix student_value field default

jmbowman:
The `student_value` field of CompletionBlock was configured to default to `None`, which just gets coerced to `False` immediately upon initialization after throwing a `ModifyingEnforceTypeWarning`.  Setting the default directly to `False` bypasses this extra work and helps cut down on the number of warnings thrown in `edx-platform` test runs.

If the intent was for there to actually be no default value, the `default` keyword can be omitted, but that would be a change from the current behavior.

### DIFF
--- a/problem_builder/completion.py
+++ b/problem_builder/completion.py
@@ -82,7 +82,7 @@ class CompletionBlock(
     student_value = Boolean(
         help=_("Records student's answer."),
         scope=Scope.user_state,
-        default=None,
+        default=False,
     )
 
     editable_fields = ('display_name', 'show_title', 'question', 'answer')
--- a/setup.py
+++ b/setup.py
@@ -72,7 +72,7 @@ BLOCKS = [
 
 setup(
     name='xblock-problem-builder',
-    version='2.8.0',
+    version='2.8.1',
     description='XBlock - Problem Builder',
     packages=find_packages(),
     install_requires=[


bradenmacdonald:
Thanks @jmbowman!

@mtyaka or @itsjeyd: please review or schedule a review for next week :)

itsjeyd:
@bradenmacdonald Created [OC-3759](https://tasks.opencraft.com/browse/OC-3759) for this.

CC @mtyaka

jcdyer:
Looks good :+1:

Verified that the current behavior is coercing to False, and that this maintains that behavior, without raising the warning.

It looks like the *intended* behavior is to treat None differently from False. There is code that does this (See get_last_result at L127) but BooleanField doesn't seem to support passing defaults of None.  If we want to maintain that behavior (or rather, reintroduce it) we create a ticket to do so, but it may involve creating a NullableBooleanField (or a Nullable() wrapper type that can be applied to various field types.

jcdyer:
Changing the default seems to change the behavior, as evidenced by the failing tests, so I'm not entirely sure about your assumption that the value is getting coerced to False.  It looks like the code treats None and False differently.  See for example [problem_builder.completion.get_last_result](https://github.com/jmbowman/problem-builder/blob/master/problem_builder/completion.py#L127-L137).  If this is creating excessive warnings, the correct solution is probably to use a NullableBooleanField here, as we are apparently tracking three separate states: unsubmitted, submitted with False, and submitted with True.

jcdyer:
@jmbowman Sorry for the delay.  I forgot I needed to push your branch to open-craft/problem-builder to get the tests running.

jmbowman:
Ok, I finally got back to this and got the tests running locally so I can reproduce the failure.  And now for some reason I don't see the warning in the edx-platform test runs anymore.  I agree that a NullableBooleanField is probably a better solution here, but that seems like more effort than this is worth at the moment.  Should I just close this PR until and unless it starts causing problems again?

jcdyer:
@jmbowman Closing this sounds right to me.

PR #186, opened 2018-04-23
Fixed feedback close button in RTL

murad-hubib:
- Fixed feeback close button direction from right to left

**Steps to test the change:**
- Select an option from MCQ/MRQ and click submit
- Feedback close button should be on left in RTL

### DIFF
--- a/problem_builder/public/css/questionnaire.css
+++ b/problem_builder/public/css/questionnaire.css
@@ -111,3 +111,9 @@
     border-right: 2px solid #ddd;
     border-left: 0;
 }
+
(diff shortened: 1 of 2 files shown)


murad-hubib:
@bradenmacdonald @mtyaka  Please review this PR

xitij2000:
@bradenmacdonald  @e-kolpakov  Created OC-4427.

mtyaka:
Thanks @murad-hubib!

:+1:

- [x] I tested this: Verified that the close button is positioned incorrectly in RTL mode without this fix, and positioned correctly with the fix enabled.
- [x] I read through the code

PR #188, opened 2018-05-09
The standard mode should be the only mode.

symbolist:
https://tasks.opencraft.com/browse/OC-968

- Remove the "mode" parameter from existing Mentoring XBlock; make sure that all the checks involving it are either removed (if one of the branches makes sense) or rewritten using some other check (should avoid that as much as possible).
- Completely remove assessment logic from existing Mentoring block. The block should only provide functionality that was part of "standard" mode.
- Update the documentation accordingly.


symbolist:
@rocioar Just a heads up that this should be ready for review tomorrow. I have to give it one last look in the morning and update the docs.

problem_builder/mentoring.py
```diff
@@ -577,32 +546,9 @@ def get_message(self, completed):
...
-        """
-        Get the message to display to a student following a submission in assessment mode.
-        """
-        if not self.max_attempts_reached:
-            return self.get_message_content('on-assessment-review', or_default=True)
-        else:
-            return None
-
     @property
     def review_tips(self):
```
symbolist: `review_tips` are also used by `MentoringWithExplicitStepsBlock` so we still return the empty list here.

problem_builder/questionnaire.py
```diff
@@ -158,14 +158,12 @@ def author_edit_view(self, context):
...
+
         fragment = self.get_author_edit_view_fragment(context)
 
-        # Let the parent block determine whether to display buttons to add review-related child blocks.
-        # * Problem Builder units use MentoringBlock parent components, which define an 'is_assessment' property,
-        #   indicating whether the (deprecated) assessment mode is enabled.
-        # * Step Builder units can show review components in the Review Step.
         fragment.add_content(loader.render_template('templates/html/questionnaire_add_buttons.html', {
-            'show_review': getattr(self.get_parent(), 'is_assessment', True),
+            'show_review': isinstance(self.get_parent(), MentoringWithExplicitStepsBlock)
```
symbolist: Open to suggestions for this.

problem_builder/templates/html/mentoring.html
```diff
@@ -15,8 +15,7 @@ <h3>{{ title }}</h3>
   </div>
   {% endif %}
 
-  <div class="{{self.mode}}-question-block">
```
symbolist: Only used by css to add some styling for assessment mode. So no longer needed.

problem_builder/templates/xml/assessment_extended_feedback.xml
```diff
@@ -1,49 +0,0 @@
-<problem-builder url_name="mentoring-assessment-1" display_name="A Simple Assessment" weight="1" mode="assessment" max_attempts="2" extended_feedback="true">
```
symbolist: Workbench example for assessment mode.

problem_builder/templates/xml/mentoring_assessment.xml
```diff
@@ -1,42 +0,0 @@
-<problem-builder display_name="Mentoring Assessment Example" weight="1" mode="assessment" max_attempts="2" extended_feedback="true">
```
symbolist: Workbench example for assessment mode.

problem_builder/mentoring.py
```diff
@@ -577,32 +546,9 @@ def get_message(self, completed):
             # Student did not achieve a perfect score but can try again:
             return self.get_message_content('incomplete')
 
-    @property
-    def assessment_message(self):
```
symbolist: Only used in assessment mode.

symbolist:
Hi @rocioar! This is ready for review.

problem_builder/questionnaire.py
```diff
@@ -158,14 +158,12 @@ def author_edit_view(self, context):
...
         """
+        from .mentoring import MentoringWithExplicitStepsBlock
+
         fragment = self.get_author_edit_view_fragment(context)
 
-        # Let the parent block determine whether to display buttons to add review-related child blocks.
-        # * Problem Builder units use MentoringBlock parent components, which define an 'is_assessment' property,
-        #   indicating whether the (deprecated) assessment mode is enabled.
-        # * Step Builder units can show review components in the Review Step.
         fragment.add_content(loader.render_template('templates/html/questionnaire_add_buttons.html', {
```
rocioar: @symbolist could you add some comment here, I see you removed the previous one, cause it had some mention of assessment, a part of it is not valid anymore, but the steps section still is.

symbolist: @rocioar Sure. The comment was just an explanation for `getattr(self.get_parent(), 'is_assessment', True)` and I thought its replacement was evident enough to not require a comment.

symbolist:
@rocioar The translation directory hasn't been updated in 3 years. I had not noticed that earlier and tried to update the translations but there is also some configuration missing. So I do not think they are used.

rocioar:
@itsjeyd would you mind just confirm if it's ok not to update translations?

itsjeyd:
@rocioar @symbolist I would assume that McKinsey aren't using the translations, but edX will most likely want us to update them before updating the upstream version of problem-builder again (i.e., the version that's installed on edx.org). Not sure when that will happen, but if we can minimize the work involved in updating the upstream version by taking care of things like translations as we go along, that definitely tends to be helpful down the road.

I'll CC @bradenmacdonald for additional input about edX's plans for updating problem-builder.

bradenmacdonald:
@itsjeyd @symbolist @rocioar 

Some instructions for updating the translations are in another XBlock: https://github.com/open-craft/xblock-drag-and-drop-v2#i18n-compatibility

It's definitely not easy though. We've talked in OC-4004 about improving i18n-tools to better handle that flow, but so far don't know if the client wants us to proceed with that or not.

So for now: Please update the translations files if you can, as it would be nice, but don't consider it necessary (in other words, fixing the process for updating this block's translations that aren't used is out of not a required part of the scope).

rocioar:
@symbolist I will approve the PR once you address my nit, and then you can work on the translations. Please squash the commits before merging.

symbolist:
@rocioar  I have added back the existing comment for the Step Builder: https://github.com/open-craft/problem-builder/pull/188/files#diff-88effd2bde10c4ba755bff423957d7a6R165

I am going to skip updating the translations as part of this ticket since the existing configuration appears to not work correctly and it will likely require some time to get it set up right.

PR #190, opened 2018-05-21
Bump up version after removing modes.

symbolist:
https://tasks.opencraft.com/browse/OC-968

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -72,7 +72,7 @@ BLOCKS = [
 
 setup(
     name='xblock-problem-builder',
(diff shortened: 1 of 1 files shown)


symbolist:
@bradenmacdonald Is there any convention for version number updates? This is a major change, so should the version be updated to `3.0.0`? Or do you have any other suggestion.

bradenmacdonald:
@symbolist I guess we try to follow https://semver.org/ but we haven't had any particular policy. `3.0.0` sounds good to me :)

PR #195, opened 2018-06-04
Remove all reference to ThemableXBlockMixin

pomegranited:
Fixes deployment issues related to https://github.com/open-craft/problem-builder/pull/192 by removing all the Problem Builder themability code.

**Testing instructions**
1. Enable xblock custom themes in your devstack to simulate the production settings:
   ```"XBLOCK_SETTINGS": {
            "mentoring": {
                "theme": {
                    "package": "problem_builder",
                    "locations": ["public/themes/apros.css"]
                },
        }
     }
    ```
1. Ensure that you can add new Problem Builder units to courses in Studio.
1. Ensure that you can view Problem Builder units in LMS/Apros.
1. Ensure that the `lms.css` file is loaded in Studio Preview and LMS views.

**Author Notes & Issues**

1. There's still a `themed-xblock` class used in the [`mentoring.html`](https://github.com/open-craft/problem-builder/blob/master/problem_builder/templates/html/mentoring.html#L2) and [`mentoring_with_steps.html`](https://github.com/open-craft/problem-builder/blob/master/problem_builder/templates/html/mentoring_with_steps.html#L2) templates, but removing these would require changing both the `lms.css` file here, and the Apros CSS file which was moved into the theme.
  Please let me know if this needs doing too.

**Reviewer**
- [ ] @xitij2000


pomegranited:
Good point @xitij2000  -- I've fixed this.

PR #197, opened 2018-06-18
Fix student_value field warning

jmbowman:
Take 2 at fixing the `ModifyingEnforceTypeWarning` that appears in all management command execution attempts in edx-platform master with this XBlock installed.  The earlier attempt was in #176, and further context is in [TE-2585](https://openedx.atlassian.net/browse/TE-2585).  This is one of two remaining warnings that appears for routine management command execution on master, and I'd like to fix both of those in time for Hawthorn if possible.

Currently Hawthorn is slated to use problem-builder 2.7.7; are there any compelling reasons why it shouldn't be updated to use the latest release after this is fixed?

### DIFF
--- a/problem_builder/completion.py
+++ b/problem_builder/completion.py
@@ -23,7 +23,7 @@
 import logging
 
 from xblock.core import XBlock
-from xblock.fields import Scope, String, Boolean
+from xblock.fields import JSONField, Scope, String, UNSET
 from xblock.fragment import Fragment
 from xblockutils.studio_editable import StudioEditableXBlockMixin
 from xblockutils.resources import ResourceLoader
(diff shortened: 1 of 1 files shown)


jmbowman:
@jcdyer Could you trigger a test run and review this when you get a chance?  The tests passed for me locally, so hopefully this will meet a better fate than the last attempt.

bradenmacdonald:
@jmbowman Thanks! `problem-builder` is not included by default in Open edX though, AFAIK, so it doesn't matter for the release. It is used on edx.org, however. Upgrading to the latest version would be great. It used to be in `edx-private.txt` but I'm not sure where that file has moved to?

jmbowman:
The private requirements are now specified in the configuration repo: https://github.com/edx/configuration/blob/master/playbooks/roles/edxapp/defaults/main.yml#L489

jmbowman:
And while problem-builder may be excluded in Open edX production installations, it's currently included in the devstack edxapp Docker container.  So unless we change that to exclude the private requirements by default, the warning would still be a nuisance for development on Hawthorn.

jcdyer:
Build triggered.

jmbowman:
The tests passed here; any rough estimate on when someone will have time to review and/or merge?

jcdyer:
I'll take a look now.

jmbowman:
I considered putting something in XBlock directly, but:

* This case is the only need for it I've seen expressed so far
* The Hawthorn release is pending, and I'd rather not make any complicated multi-repo changes while trying to wrap that up

My inclination is to go with this for now, and refactor it later if a more general need for such nullable fields becomes apparent.

jmbowman:
@jcdyer Are you ok with going ahead and merging this?  I don't have permission to do so.

jcdyer:
Merged.

jmbowman:
Thanks!  Is this a good time to do a release, or should I update the edx.org configuration to install from the merge commit on master?  (At some point I'd like to see this package on PyPI also, but that can be tackled later.)

bradenmacdonald:
I'm in favor of a versioned release and putting this on PyPI. usually we try to include a minor version bump in each PR we merge to this repo but forgot here.

jmbowman:
Should I create a new PR for the version bump?  I could also attempt to set up automatic deployments to PyPI, but you'd need to add the appropriate credentials and diagnose any failures.  Most of the edX packages deploy to PyPI via Travis, might be more appropriate to use CircleCI here since it's already in use: https://circleci.com/blog/continuously-deploying-python-packages-to-pypi-with-circleci/

bradenmacdonald:
Sure, please open a PR for the version bump. And re PyPi, yeah, we'd likely want to use CircleCI and have it push to our own OpenCraft PyPI account.

PR #199, opened 2018-06-27
Version 3.1.3, automatic PyPI uploads

jmbowman:
This bumps the version and attempts to set up automated PyPI uploads via CircleCI following the advice in [their blog post on the topic](https://circleci.com/blog/continuously-deploying-python-packages-to-pypi-with-circleci/):

* Removed an obsolete pin on an old version of setuptools (the referenced bug was long since fixed)
* Added a deployment job.
* Only run the deployment job for git tag pushes, and only if the build job for the same push succeeds.

Before the automatic uploads will work, you need to add the `PYPI_PASSWORD` environment variable to the CircleCI settings, as explained [here](https://circleci.com/docs/2.0/env-vars/#overview) ("the Environment Variables page of the Build > Project > Settings in the CircleCI application").  This should be the PyPI password of the opencraft account.

It used to be necessary to manually register a package on PyPI before uploading the first release, but it looks like with the switch to Warehouse this is no longer needed.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -14,9 +14,6 @@ jobs:
             virtualenv venv
             . venv/bin/activate
             pip install -U pip wheel
-            # Temporarily pin setuptools to a specific version.
-            # See commit message of https://github.com/open-craft/problem-builder/commit/51277a34fb426724616618c1afdb893ab2de4c6b for more info:
-            pip install setuptools==24.3.1
             pip install -e git://github.com/edx/xblock-sdk.git@bddf9f4a2c6e4df28a411c8f632cc2250170ae9d#egg=xblock-sdk
             pip install -r requirements.txt
             pip install -r venv/src/xblock-sdk/requirements/base.txt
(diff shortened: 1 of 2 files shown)


jmbowman:
@bradenmacdonald Here's the version bump and auto-upload attempt; if it looks ok, please go ahead and trigger a build to make sure I didn't break anything in the CircleCI configuration.  I noticed that the last master build failed attempting to download Firefox, but the file still seems to be there so I'm hoping that was just a temporary Sourceforge outage.

bradenmacdonald:
It worked! https://pypi.org/project/xblock-problem-builder/

bradenmacdonald:
BTW, I'm assuming there's no issue with the PyPI package name (`xblock-problem-builder`) being different from the repo name (`problem-builder`) and the package name (`problem_builder`) ?

jmbowman:
Correct, the name of the package on PyPI is purely determined by the `name` field in `setup.py`, and often differs from those other characteristics of the code.

PR #200, opened 2018-06-30
Restructure i18n files

mahyard:
Via this PR:
1. Latest messages extracted
2. "locale" directory added to the main app and added to package_data
3. Underscore template files renamed with .underscore extension 

Some changes to README.md may be necessary.


pomegranited:
Thank you for this PR @mahyard !  Please ping me when you're ready for review.

mtyaka:
@mahyard It's fine to remove those files if they are not needed. I don't think translations ever worked correctly for this xblock.

The [XBlock documentation|http://edx.readthedocs.io/projects/xblock-tutorial/en/latest/edx_platform/edx_lms.html#internationalization-support] says that translations should be present in the `/<package>/translations/<lang>/LC_MESSAGES/text.po` though. The DnDv2 xblock seems to follow that approach: https://github.com/edx-solutions/xblock-drag-and-drop-v2/tree/master/drag_and_drop_v2/translations/en/LC_MESSAGES.

Please ping me when the PR is ready for review.

mahyard:
Hi @mtyaka @pomegranited 
It's ready to review

bradenmacdonald:
@mtyaka 

> I just checked the development branch and it seems that the completable-by-viewing code is still not present.

I thought that we had merged the JS i18n code onto the solutions fork, so we could be using that now, at least on that fork?

pomegranited:
@bradenmacdonald @mtyaka  We merged https://github.com/zyegfryed/django-statici18n/pull/41 to support JS i18n in XBlocks, but have only used it on DnDv2 so far (cf https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/156).

mtyaka:
@bradenmacdonald Did you copy-paste a wrong quote? I don't see how completable-by-viewing code is relevant :)

@pomegranited Do we need to open a PR to update the version of `django-statici18n` in edx/edx-platform? Are there any other changes changes to edx/edx-platform and/or edx/XBlock required for namespaced JS translations to work?

pomegranited:
@mtyaka 
> Do we need to open a PR to update the version of django-statici18n in edx/edx-platform? 

Not required -- it only needs to be [used by the XBlock](https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/156/files#diff-b4ef698db8ca845e5845c4618278f29aR4) when generating the JS files shipped with that xblock, and so doesn't require the platform to use the same version.

> Are there any other changes changes to edx/edx-platform and/or edx/XBlock required for namespaced JS translations to work?

Nope, the changes are isolated to the XBlock itself.  Namespacing protects the platform's JS from the XBlock's specific translations, and only just changes how the XBlock access its own translations.

mtyaka:
@pomegranited Oh excellent, thanks for the info!

bradenmacdonald:
@mtyaka lol whoops. I've edited my comment. I meant to reply to "Since the i18n does not yet support JS files, I am merging this PR as an intermediate solution, but we will want to switch to i18n service in the future."

PR #201, opened 2018-07-02
Update README to reflect PyPI usage

bradenmacdonald:
Follow up to #199 to update the docs accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,6 +191,12 @@ To install old verions of Problem Builder (< v3.1.3) on an Open edX installation
 TAG='v2.6.5' pip install "git+https://github.com/open-craft/problem-builder.git@$TAG#egg=xblock-problem-builder==$TAG"
 ```
 
+Note that Problem Builder requires [xblock-utils](https://github.com/edx/xblock-utils).
+If you are installing it into a virtualenv used by edx-platform, xblock-utils should
+already be installed. But if you are installing it into another virtualenv, you may
+need to first install xblock-utils manually (recent versions of it are not available
+on PyPI so will not be automatically installed).
+
 See [Usage Instructions](doc/Usage.md) for how to enable in Studio.
 
 Publishing to PyPI
--- a/README.md
+++ b/README.md
@@ -37,13 +37,14 @@ containing a free-form question, two MCQs and one MRQ:
 Installation
 ------------
 
-Install the requirements into the Python virtual environment of your
-`edx-platform` installation by running the following command from the
-root folder:
+You can install Problem Builder from [PyPI](https://pypi.org/project/xblock-problem-builder/)
+using this command:
 
-```bash
-$ pip install -r requirements.txt
 ```
+pip install xblock-problem-builder
+```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -37,13 +37,14 @@ containing a free-form question, two MCQs and one MRQ:
...
-Install the requirements into the Python virtual environment of your
-`edx-platform` installation by running the following command from the
-root folder:
+You can install Problem Builder from [PyPI](https://pypi.org/project/xblock-problem-builder/)
+using this command:
 
-```bash
-$ pip install -r requirements.txt
 ```
+pip install xblock-problem-builder
```
pomegranited: @bradenmacdonald  CC @jmbowman I get an error when I run this: 
```
(venv) $ pip install xblock-problem-builder==3.1.3
Collecting xblock-problem-builder==3.1.3
  Using cached https://files.pythonhosted.org/packages/d5/5d/12eb466347984534705efceed7da59883df004a1ce71f4d04b9fa51f1191/xblock_problem_builder-3.1.3-py2-none-any.whl
Collecting XBlock (from xblock-problem-builder==3.1.3)
  Using cached https://files.pythonhosted.org/packages/71/8e/cd778ea775fee0badffa33cfaa908dbd341c06aaf73cc70d9b5d930b4925/XBlock-1.2.1.tar.gz
Collecting xblock-utils (from xblock-problem-builder==3.1.3)
  Could not find a version that satisfies the requirement xblock-utils (from xblock-problem-builder==3.1.3) (from versions: )
No matching distribution found for xblock-utils (from xblock-problem-builder==3.1.3)
```

And sure enough, pypi.org doesn't have an [`xblock-utils` package as requested by setup.py](https://github.com/open-craft/problem-builde/blob/ec20cafd1469d1becca11253fc25085b7e6f54d7/setup.py#L101), it's called [`xblock-openedx-utils`](https://pypi.org/project/xblock-openedx-utils/) instead!

bradenmacdonald: @pomegranited edX actually doesn't publish xblock-utils to PyPI. That seems to be something posted by Stanford years ago and it's out of date (v0.9.0 vs 1.1.1). It seems that [edX installs xblock-utils from git](https://github.com/edx/edx-platform/blob/9b92b21bb6b95ffe0a9faffaf657f3346837d6b2/requirements/edx/github.in#L97)

So I think for now we should just merge this anyways and consider it a non-pip-installable prereq. Any edx-platform virtualenv will already have xblock-utils installed anyways.

CC @jmbowman in case you have any input.

pomegranited: @bradenmacdonald  Huh, ok!  So could we instead remove the `install_requires` reference to `xblock-utils` from [setup.py](https://github.com/open-craft/problem-builder/blob/master/setup.py#L101)?

bradenmacdonald: Well it is still a requirement, so I don't want to remove it. I have updated the README to explain that in 6d62e61c988f4d8ff252df230dbc1f281688a177

pomegranited: (y)

PR #202, opened 2018-08-06
[OC-4912] [MCKIN-7492] Removed the themed-xblock selector to allow overrides by themes (v3)

xitij2000:
Fix issues with theme CSS overrides due to overly-specific selectors.


problem_builder/mentoring.py
```diff
@@ -979,7 +977,6 @@ def student_view(self, context):
             'children_contents': children_contents,
         }))
         fragment.add_css_url(self.runtime.local_resource_url(self, 'public/css/problem-builder.css'))
-        fragment.add_css_url(self.runtime.local_resource_url(self, 'public/css/lms.css'))
```
pomegranited: If we're removing this `lms.css` CSS file, then why bother keeping it (and editing it) below?

problem_builder/public/css/problem-builder.css
```diff
@@ -118,16 +118,62 @@
 
 .mentoring .checkmark-incorrect {
     font-size: 22pt;
-    color: #ff0000;
+    color: #c1373f;
```
pomegranited: nit: where does this color come from?  Does it match an edX standard color somewhere?

problem_builder/mentoring.py
xitij2000: Oh sorry! I cherry-picked the v2 changes here, but I guess that doesn't delete files. I'll delete lms.css

problem_builder/public/css/problem-builder.css
xitij2000: I just moved it from lms.css here. I don't see it used in edX, so not sure why this was picked.

pomegranited: Ah ok, understood.  👍

xitij2000:
@pomegranited I updated the circle CI config to run ``apt-get update`` before installing packages, and it's fixed now. I'm keeping it a separate commit since it's not related to the rest of the work.

PR #203, opened 2018-08-06
Removed the themed-xblock selector to allow overrides by themes (v2)

xitij2000:
Fix issues with theme CSS overrides due to overly-specific selectors.


mtyaka:
Thanks @xitij2000! I tested this and found one problem - the correct/incorrect icons do not appear for me in the LMS. The fix is simple, I suggested it in inline comments and tested it, so I don't think this needs another round of reviews and I can approve it now.

:+1:

- [x] I tested this: I verified that the merged styles look correct after the removal of lms.css. I also verified they work correctly together with the apros theme.
- [x] I read through the code

PR #204, opened 2018-08-27
Make static urls absolute MCKIN-8062-8059

jawad-khan:
Made static urls absolute in Problem builder, mrq and mcq.

### DIFF
--- a/problem_builder/mixins.py
+++ b/problem_builder/mixins.py
@@ -1,7 +1,10 @@
 import json
+import re
 
 import webob
 from lazy import lazy
+from urlparse import urljoin
+from django.conf import settings
 from problem_builder.tests.unit.utils import DateTimeEncoder
 from xblock.core import XBlock
 from xblock.fields import String, Boolean, Float, Scope, UNIQUE_ID
@@ -261,3 +264,30 @@ class StudentViewUserStateResultsTransformerMixin(object):
         except KeyError:
             pass
         return dictionary
+
+
+class MakeURLAbsoluteMixin(object):
+    url_pattren = re.compile('<img\s*src="(.*)"')
+
+    def make_url_absolute(self, text):
+        if not text:
+            return text
+
+        try:
+            from static_replace import replace_static_urls
+        except ImportError:
+            return text
+
+        lms_relative_url = replace_static_urls(text, course_id=self.course_id)
+        return self._append_lms_base(lms_relative_url)
+
+    def _append_lms_base(self, text):
+        urls = self.url_pattren.findall(text)
+        if not urls:
+            return text
+
+        lms_base = settings.ENV_TOKENS.get('LMS_BASE')
+        scheme = 'https' if settings.HTTPS == 'on' else 'http'
+        lms_base = '{}://{}'.format(scheme, lms_base)
+        absolute_url = urljoin(lms_base, urls[0])
+        return text.replace(urls[0], absolute_url)
--- a/problem_builder/mixins.py
+++ b/problem_builder/mixins.py
@@ -1,10 +1,7 @@
 import json
-import re
 
 import webob
 from lazy import lazy
-from urlparse import urljoin
-from django.conf import settings
 from problem_builder.tests.unit.utils import DateTimeEncoder
 from xblock.core import XBlock
 from xblock.fields import String, Boolean, Float, Scope, UNIQUE_ID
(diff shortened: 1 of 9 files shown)


problem_builder/mixins.py
```diff
@@ -261,3 +264,30 @@ def delete_key(self, dictionary, key):
...
+    def make_url_absolute(self, text):
+        if not text:
+            return text
+
+        try:
+            from static_replace import replace_static_urls
+        except ImportError:
+            return text
+
+        lms_relative_url = replace_static_urls(text, course_id=self.course_id)
```
bradenmacdonald: I don't think this code is correct; I think it should be https://github.com/edx-solutions/xblock-drag-and-drop-v2/blob/987b7dc424d744633bad55d7d45b7ebdf3105a1d/drag_and_drop_v2/drag_and_drop_v2.py#L828-L845

problem_builder/mixins.py
```diff
@@ -261,3 +264,30 @@ def delete_key(self, dictionary, key):
...
+        return self._append_lms_base(lms_relative_url)
+
+    def _append_lms_base(self, text):
+        urls = self.url_pattren.findall(text)
+        if not urls:
+            return text
+
+        lms_base = settings.ENV_TOKENS.get('LMS_BASE')
```
bradenmacdonald: XBlocks should not be importing the LMS django settings, since that goes outside of the XBlock API. What problem are we solving by using a full absolute URL instead of just the result of `replace_static_urls` ?

janomanningham: @bradenmacdonald this has nothing to do with drag and drop. we're context switching to support graphic images and static links in general to appear on mobile for problem builder and step builder. if you recall from this ticket: https://edx-wiki.atlassian.net/browse/MCKIN-7725, we reached an implementation decision to fix these two xblocks so that it always returns absolute links only, without any changes from the course authoring team. So, I believe @jawad-khan had pb and sb call expand_static_url on file fields or self.runtime.replace_urls() on HTML fields before serving them via its API. so based on this, does @jawad-khan PR make sense to you?

bradenmacdonald: @janomanningham  I totally understand that, and I agree we want to add `expand_static_url` here. I just am not sure if the extra step of adding `LMS_BASE` is required here, and it goes around the XBlock API, which should generally be avoided.

The reason I mention drag and drop is that we applied a similar fix to drag and drop a long time ago, and adding the LMS_BASE part wasn't necessary then AFAIK.

jawad-khan: @bradenmacdonald I have used expand_static_url now(thanks for the suggestion).
I was using LMS_BASE because it was easy for mobile side to just show the data as it is. I have talked to them and they say they will make changes on their end to append base url.
I have removed LMS_BASE part now.

bradenmacdonald: @jawad-khan Thanks! Sounds great to me.

bradenmacdonald:
@jawad-khan Your reviewer will be @UmanShahzad - no sign off from me needed.

problem_builder/mixins.py
```diff
@@ -266,28 +263,26 @@ def delete_key(self, dictionary, key):
         return dictionary
 
 
-class MakeURLAbsoluteMixin(object):
-    url_pattren = re.compile('<img\s*src="(.*)"')
+class ExpandStaticURLMixin(object):
 
-    def make_url_absolute(self, text):
+    def expand_static_url(self, text):
```
bradenmacdonald: Nit: should be `expand_static_urls` since this scans text and replaces multiple URL occurrences in the supplied `text`, unlike the dndv2 code for `expand_static_url` which accepts a single URL string.

jawad-khan:
@UmanShahzad I don't have any test instructions for this change.
Also there aren't  any new tests.

problem_builder/mixins.py
```diff
@@ -261,3 +261,28 @@ def delete_key(self, dictionary, key):
...
+        for this.
+        """
+        if not text:
+            return text
+
+        if hasattr(self.runtime, 'replace_urls'):
+            text = self.runtime.replace_urls(text)
+        elif hasattr(self.runtime, 'course_id'):
+            # edX Studio uses a different runtime for 'studio_view' than 'student_view',
+            # and the 'studio_view' runtime doesn't provide the replace_urls API.
```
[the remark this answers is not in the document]
jawad-khan: @UmanShahzad For me on LMS runtime doesn't have "replace_urls", so I have to to use replace static urls. So mixin approach seems best in this case for me.

jawad-khan:
@UmanShahzad 
These are places where I have applied my changes:
MCQ choices.
MCQ question text.
MCQ choices on feedback screen.
MCQ feedback message (tips ) on feedback screen.
MCQ message complete.
MCQ message incomplete.
MRQ choices.
MRQ question text.
MRQ choices on feedback.
MRQ feedback message on feedback.
MRQ message complete.
MRQ message incomplete.
FTE question.
FTE message complete.

You can try with variations of above fields to ensure that everything is working fine.

UmanShahzad:
@jawad-khan Thanks a lot! I will get back to you today.

problem_builder/mixins.py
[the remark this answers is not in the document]
UmanShahzad: I see, OK.

jawad-khan:
@UmanShahzad Can you please squash commits and merge it. Thanks.

UmanShahzad:
@jawad-khan Will do in a moment.

xitij2000:
@jawad-khan This PR was targeted to master which is Problem Builder v3. Does this change also need to be applied to v2 which I believe is currently in use? If so please create a PR targeting the problem-builder-v2 branch.

jawad-khan:
Thanks @xitij2000. I am making a PR agains problem-builder-v2.

PR #209, opened 2018-10-26
add french, german and japanese and udpate remaining languages

umarmughal824:
Add french, german and japanese languages support and update remaining languages.


antoviaque:
@umarmughal824 Thank you for the contribution! We will schedule a review for this as soon as we get a chance.

umarmughal824:
@kaizoku Can you please check whether these unstranslated strings are being translated in other languages or not?

xitij2000:
@kaizoku The video and ooyala block embeds also come from separate blocks I think, so they would need to be translated in those blocks. 

@umarmughal824 I have checked and they are not. These are probably missed since they are defined in xblock-utils so would need to be translated there, so for this PR we are good. I am inclined to merge it. 

However, I notice that you are targeting the master branch which is for problem builder 3. Could you open the same PR targeting the problem builder v2 branch?

PR #210, opened 2018-10-31
MCKIN-8751: Add markers to set completion criteria.

jcdyer:
MCKIN-8751

In addition to the mentioned block, it looks like some others may not have been configured properly.

Open question: should answer recap be excluded or complete-on-view?

### DIFF
--- a/problem_builder/answer.py
+++ b/problem_builder/answer.py
@@ -281,6 +281,7 @@ class AnswerRecapBlock(AnswerMixin, StudioEditableXBlockMixin, XBlock):
     """
     A block that displays an answer previously entered by the student (read-only).
     """
+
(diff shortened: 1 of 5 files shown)


xitij2000:
@jcdyer There seems to be a failing test, could you look into that?

jcdyer:
@xitij2000 Will do.  It looks unrelated, but I'll investigate.

jcdyer:
@xitij2000 Looks like the same thing @lgp171188 just ran into and fixed.  I'm cherry picking his fix.

jcdyer:
Tests passed, RFR.

xitij2000:
@jcdyer Could you cherry pick this to master as well?

PR #211, opened 2018-10-31
Fix failing CI checks

lgp171188:
This PR fixes the issues causing the CI checks to fail.

* Run apt-get update before installing packages (cherry-picked from the master branch).
* Fix code style issues.
* Add a delay in the failing test for the result block content to show up.

PR #212, opened 2018-11-01
Add a delay in the test for the result block content to show up

lgp171188:
This fixes an integration test failure. The change is based on the fix in https://github.com/open-craft/problem-builder/pull/211


viadanna:
@lgp171188 Since this only touches the test, can we trust CircleCI and forego manually running the tests?

lgp171188:
@viadanna, yes, it is okay to trust the Circle CI checks and there is no need to run the tests manually :+1:

viadanna:
Sorry for the holdup @lgp171188 @xitij2000

PR #218, opened 2018-11-29
[BB-643] Fix the StopIteration exception error

lgp171188:
When the submission of a user is not found, calling `next()` on the generator expression throws a `StopIteration` exception which was not handled. This PR adds a `try-except` handler for the exception. The `except` block does nothing because this situation is handled in the following `if` condition.

**Testing instructions**
1. Clone and Install problem_builder with the PR branch.
1. From studio Add problem builder block containing MCQ/MRQ and Instructor tool block.
1. Go to Apros and attempt question in problem builder block.
1. Do the previous step for a few users.
1. Delete one of the users used in the previous step.
1. Go to Instructor tool block in apros and click search and it will start exporting for all blocks.
1. Verify that there was no exception thrown and `(<student ID>, 'N/A', 'N/A')` is returned for that user corresponding to that submission.
1. Alternatively, modify [this line](https://github.com/open-craft/problem-builder/pull/218/files#diff-62503b7f5c87c7f593fba9e2a54320fbR215) to return `{}` and verify that `(<student ID>, 'N/A', 'N/A')` is returned for that user corresponding to that submission.

### DIFF
--- a/problem_builder/tasks.py
+++ b/problem_builder/tasks.py
@@ -120,7 +120,10 @@ def _extract_data(course_key_str, block, user_id, match_string):
     # - For each submission, look up student's username, email and answer:
     answer_cache = {}
     for submission in submissions:
-        username, _user_id, user_email = _get_user_info(users, submission)
+        username, _user_id, user_email = users.get(
+            submission['student_id'],
+            (submission['student_id'], 'N/A', 'N/A')
+        )
         answer = _get_answer(block, submission, answer_cache)
 
         # Short-circuit if answer does not match search criteria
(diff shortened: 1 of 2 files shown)


problem_builder/tasks.py
```diff
@@ -194,32 +197,24 @@ def _get_submissions(course_key_str, block, user_id):
...
         anonymoususerid__anonymous_user_id__in=anonymous_ids
-    ).annotate(anonymous_user_id=F('anonymoususerid__anonymous_user_id')).iterator()
+    ).annotate(
+        anonymous_user_id=F('anonymoususerid__anonymous_user_id')
+    ).values(
+        'anonymous_user_id', 'username', 'id', 'email'
+    ).iterator()
+
+    return {
+        user['anonymous_user_id']: (user['id'], user['username'], user['email']) for user in users
```
xitij2000: The tuple order here is not the same as the one returned by _get_user_info, which means that the reports contain the user id in place of the username.

lgp171188: @xitij2000, thanks for catching this. I have fixed this now.

PR #225, opened 2019-02-01
BB-785 deleting Answer data via Django signals

john2x:
This PR adds a Django signal handler to delete `Answer` objects when a `django.contrib.auth.User` object is deleted from LMS.

The `Answer` model has a field for `student_id` which stores a string [taken](https://github.com/open-craft/problem-builder/blob/john2x/BB-785-delete-xblock-custom-model/problem_builder/answer.py#L67) from `student.AnonymousUserId.anonymous_user_id`.

The signal receiver function connects to `AnonymousUserId` instead of `User` because the former will get deleted via CASCADE when the latter is deleted.

# Test instructions

## Minimal Setup

Create a test course with at least a "Long Answer Question" component from this package

## Expanded Setup

The expanded setup is similar to the minimal setup, but would include modules/components from the XBlocks listed in [the Jira ticket](https://tasks.opencraft.com/browse/BB-785?focusedCommentId=101483&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-101483).

## Verify behavior

1. Check out this branch and install this package into your devstack
2. Start the course as a user
3. Submit answers to the modules/components
4. As you are going through the course, make sure to add some bookmarks
   - Author's note: To add bookmarks, I had to do it via the LMS UI and not Apros
5. After completing the course, verify the records stored for the user:
    ```
    >>> user = User.objects.get(username='john')
    >>> Bookmark.objects.filter(user=user).count()
    1
    >>> StudentModule.objects.filter(student=user).count()
    27
    >>> XModuleStudentInfoField.objects.filter(student=user).count()
    0
    >>> XModuleStudentPrefsField.objects.filter(student=user).count()
    0
    >>> AnonymousUserId.objects.filter(user=user).count()
    2
    >>> Answer.objects.filter(student_id__in=AnonymousUserId.objects.filter(user=user).values_list('anonymous_user_id', flat=True)).count()
    1
    ```
6. Delete the user
    ```
    >>> user.delete()
    ```
7. Verify the records for the user have been deleted
    ```
    >>> Bookmark.objects.filter(user=user).count()
    0
    >>> StudentModule.objects.filter(student=user).count()
    0
    >>> XModuleStudentPrefsField.objects.filter(student=user).count()
    0
    >>> AnonymousUserId.objects.filter(user=user).count()
    0
    >>> Answer.objects.filter(student_id__in=AnonymousUserId.objects.filter(user=user).values_list('anonymous_user_id', flat=True)).count()
    0
    >>> Answer.objects.all.count()
    0
    ```

# Author's notes and concerns

- Added `django-pyfs` as a dependency in `requirements-dev.txt` because `run_tests.py` was failing to run without it
- The [`test_signal_receiver_connection` test](https://github.com/open-craft/problem-builder/compare/john2x/BB-785-delete-xblock-custom-model?expand=1#diff-87000eee015fac540c3d5af65ad462c5R47) is currently skipped. Need to figure out how to run it.

### DIFF
--- a/problem_builder/models.py
+++ b/problem_builder/models.py
@@ -27,9 +27,8 @@ from django.contrib.auth.models import User
 try:
     # workaround so we don't explicitly import the AnonymousUserId model from LMS
     from student.models import AnonymousUserId
-    ANONYMOUS_USER_ID_IMPORTED = True
 except ImportError:
-    ANONYMOUS_USER_ID_IMPORTED = False
+    AnonymousUserId = None
 
 
 # Classes ###########################################################
@@ -92,5 +91,5 @@ def delete_anonymous_user_answers(sender, **kwargs):
     Answer.objects.filter(student_id=instance.anonymous_user_id).delete()
 
 
-if ANONYMOUS_USER_ID_IMPORTED:
+if AnonymousUserId:
     pre_delete.connect(delete_anonymous_user_answers, sender=AnonymousUserId)
--- a/problem_builder/models.py
+++ b/problem_builder/models.py
@@ -21,7 +21,15 @@
 # Imports ###########################################################
 
 from django.db import models
+from django.db.models.signals import pre_delete
 from django.contrib.auth.models import User
+
+try:
+    # workaround so we don't explicitly import the AnonymousUserId model from LMS
+    from student.models import AnonymousUserId
+    ANONYMOUS_USER_ID_IMPORTED = True
+except ImportError:
+    ANONYMOUS_USER_ID_IMPORTED = False
 
 
 # Classes ###########################################################
@@ -72,3 +80,17 @@ class Share(models.Model):
         # specify the app_label here.
         app_label = 'problem_builder'
         unique_together = (('shared_by', 'shared_with', 'block_id'),)
+
+
+# Signals ###########################################################
+
+def delete_anonymous_user_answers(sender, **kwargs):
+    """
+    Delete Answer records when an AnonymousUserId is deleted.
+    """
+    instance = kwargs['instance']
+    Answer.objects.filter(student_id=instance.anonymous_user_id).delete()
+
+
+if ANONYMOUS_USER_ID_IMPORTED:
+    pre_delete.connect(delete_anonymous_user_answers, sender=AnonymousUserId)
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -6,3 +6,4 @@ transifex-client==0.12.1
 edx-i18n-tools==0.4.5
 pycodestyle==2.4.0
 pylint==0.28.0
+django-pyfs==2.0
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -6,4 +6,3 @@ transifex-client==0.12.1
 edx-i18n-tools==0.4.5
 pycodestyle==2.4.0
 pylint==0.28.0
-django-pyfs==2.0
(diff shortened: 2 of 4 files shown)


problem_builder/models.py
```diff
@@ -72,3 +80,17 @@ class Meta(object):
...
+def delete_anonymous_user_answers(sender, **kwargs):
+    """
+    Delete Answer records when an AnonymousUserId is deleted.
+    """
+    instance = kwargs['instance']
+    Answer.objects.filter(student_id=instance.anonymous_user_id).delete()
+
+
+if ANONYMOUS_USER_ID_IMPORTED:
+    pre_delete.connect(delete_anonymous_user_answers, sender=AnonymousUserId)
```
xitij2000: Here is an alternate approach which will make this easier to test:

Create a function called ``get_anonymous_user_id_model`` that imports and returns the AnonymousUserId model. This can be then stubbed by the tests to use a different model or a mock to test it. 

As it stands these tests will only run in edx-platform I think.

requirements-dev.txt
```diff
@@ -6,3 +6,4 @@ transifex-client==0.12.1
 edx-i18n-tools==0.4.5
 pycodestyle==2.4.0
 pylint==0.28.0
+django-pyfs==2.0
```
xitij2000: Not sure why this was added.

john2x: @xitij2000 I  had to install `django-pyfs` manually to be able to run the test script. I will remove this since it's not part of the task.

problem_builder/models.py
john2x: @xitij2000 Hmm I've tried mocking the model, but it didn't work as expected since the condition to `connect` the signal happens when the `models.py` module is loaded, so the mock didn't have the chance to replace it yet.

Here's what I've tried:

**`models.py`**
```python
def get_anonymous_user_id_model():
    """
    Get AnonymousUserId via a function so it can be mocked in tests.
    """
    try:
        return AnonymousUserId
    except ImportError:
        pass
...
if get_anonymous_user_id_model():
    pre_delete.connect(delete_anonymous_user_answers, sender=get_anonymous_user_id_model())
```

Maybe I'm missing something?

Also, the `delete_anonymous_user_answers` function itself is already tested by [this test](https://github.com/open-craft/problem-builder/pull/225/files#diff-87000eee015fac540c3d5af65ad462c5R36). I'm wondering if we still need to test the condition that `connect`s the signal receiver?

xitij2000: @john2x Yeah, you're right, the main thing is tested anyway, and signal handling is a core mechanic so doesn't really need testing. 

I'll test is manually in the devstack, I think that should be enough.

problem_builder/models.py
```diff
@@ -21,8 +21,16 @@
...
 from django.db import models
+from django.db.models.signals import pre_delete
 from django.contrib.auth.models import User
 
+try:
+    # workaround so we don't explicitly import the AnonymousUserId model from LMS
+    from student.models import AnonymousUserId
+    ANONYMOUS_USER_ID_IMPORTED = True
+except ImportError:
+    ANONYMOUS_USER_ID_IMPORTED = False
```
xitij2000: Rather than have a separate variable for, you can do:

```python
try:
    # workaround so we don't explicitly import the AnonymousUserId model from LMS
except ImportError:
    AnonymousUserId = None
...
if AnonymousUserId is not None:
    pre_delete.connect(delete_anonymous_user_answers, sender=AnonymousUserId)
```

john2x: @xitij2000 Thanks! Yes, that does look better. I've updated the condition.

PR #234, opened 2019-05-02
BB-1128 Update circle.yml

xirdneh:
Parallelize test to run quality and unit/integration tests in different containers.

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -1,11 +1,781 @@
 [MASTER]
+
+# A comma-separated list of package or module names from where C extensions may
+# be loaded. Extensions are loading into the active Python interpreter and may
+# run arbitrary code
+extension-pkg-whitelist=
+
+# Add files or directories to the blacklist. They should be base names, not
+# paths.
 ignore=problem_builder/migrations
 
-[REPORTS]
-reports=no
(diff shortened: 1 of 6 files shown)


pylintrc
```diff
@@ -1,11 +1,781 @@
...
+# Disable the message, report, category or checker with the given id(s). You
+# can either give multiple identifiers separated by comma (,) or put this
+# option multiple times (only on the command line, not in the configuration
+# file where it should appear only once).You can also use "--disable=all" to
+# disable everything first and then reenable specific checks. For example, if
+# you want to run only the similarities checker, you can use "--disable=all
+# --enable=similarities". If you want to run only the classes checker, but have
+# no Warning level messages displayed, use"--disable=all --enable=classes
+# --disable=W"
+disable=blacklisted-name,
```
giovannicimolin: @xirdneh Why are we disabling all of this?

xirdneh: @giovannicimolin Yeap, not sure why, tho.
I created the configuration using the flags that were [in `tox.ini`](https://github.com/open-craft/problem-builder/blob/master/tox.ini#L22)

PR #236, opened 2019-06-12
BB-1357 Add transifex configuration and translation make tragets

xirdneh:
This PR updates `Makefile` to add a few targets to manage translations and transifex push/pull operations.

**JIRA tickets**:

**Discussions**: 

**Dependencies**: None

**Screenshots**: 
Screen Shot 2019-06-13 at 10 17 03 PM

**Sandbox URL**: TBD - sandbox is being provisioned.

**Merge deadline**: "None" if there's no rush, "ASAP" if it's critical, or provide a specific date if there is one.

**Testing instructions**:

1. Run `make build_dummy_translations`
2. Run 'make validate_translations`

1. There seems to be a problem when running `make extract_translations`. The issue seems to be in the code but I haven't found the line where the issue is that might be outside of this ticket scope.


viadanna:
@xirdneh It seems the `extract_translations` isn't just broken, it also breaks every other option that makes use of `compile_translations`.

I also found that the `locale` symlink should be in the `problem_builder` directory but there's a locale directory there out of sync with `translations`.

Seems this will take somewhat longer than first estimated.

xirdneh:
@viadanna I guess there was some issue with how the compiled translations were created. I removed the out of sync folder and recompiled the translations. I think it's working now. The maketargets work now. I'm double checking stuff by hand.

xirdneh:
@viadanna I've added a ss of the problem builder translated to polish to make sure it works

xirdneh:
@viadanna Ah, that's why it worked when I removed the compiled files.
I did noticed that duplicate entry and it works when removing it from the `.po` file.
So, we could remove the tag line that's redundant.

viadanna:
Actually, a change to the Makefile to use msgcat instead can fix the compilation step:
```
diff --git a/Makefile b/Makefile
index d9e4aa1..dc734a9 100644
--- a/Makefile
+++ b/Makefile
@@ -16,9 +16,8 @@ upgrade:
 
 extract_translations: ## extract strings to be translated, outputting .po files
        cd $(WORKING_DIR) && i18n_tool extract
-       mv $(EXTRACTED_DJANGO) $(EXTRACTED_TEXT)
-       tail -n +20 $(EXTRACTED_DJANGOJS) >> $(EXTRACTED_TEXT)
-       rm $(EXTRACTED_DJANGOJS)
+       msgcat $(EXTRACTED_DJANGO) $(EXTRACTED_DJANGOJS) > EXTRACTED_TEXT
+       rm $(EXTRACTED_DJANGO) $(EXTRACTED_DJANGOJS)
        sed -i'' -e 's/nplurals=INTEGER/nplurals=2/' $(EXTRACTED_TEXT)
        sed -i'' -e 's/plural=EXPRESSION/plural=\(n != 1\)/' $(EXTRACTED_TEXT)

```

xirdneh:
@viadanna Or better yet, we could use [msguniq](https://www.gnu.org/software/gettext/manual/html_node/msguniq-Invocation.html)

xirdneh:
@viadanna Ah, I see. That's probably a better solution. We could remove `tail` then since that's basically what is doing.

xirdneh:
@viadanna I've removed tail and use `msgcat` to concatenate python and JS tanslations.
Probably a good idea to do it across the board in other xblocks.

xirdneh:
@viadanna Done, thanks!

PR #241, opened 2019-08-28
Update translations (autogenerated message)

OmarIthawi:
We've just got `fr_CA` translated to :100: :tada: . I'm updating the upstream repo.

 - [x] Get the `fr_CA` :100: translated
 - [x] Pull from Transifex
 - [x] Test on devstack
 - [ ] Get it reviewed
 - [ ] Bump up the version?
 - [ ] Merge the PR

cc: @pomegranited


pomegranited:
Thank you for these PRs @OmarIthawi !  Will get them reviewed in the next couple of days.

OmarIthawi:
Thank you @pomegranited!

BTW, there's an idea to use the @edx-transifex-bot to self manage such PRs on a regular basis as seen in https://github.com/edx/edx-platform/pull/21394

I think that'll save time on both of the creator and the reviewer.

OmarIthawi:
@pomegranited I've tested it on devstack and it seems to be working.

It also has the [default XBlock fields in English problem](https://discuss.openedx.org/t/translate-the-default-xblock-data), so I had to edit those via Studio.

<kbd>image</kbd>

I've also found that the copyright text is not translatable, it's a bug that I might address in another time:

https://github.com/open-craft/problem-builder/blob/35700373be867a0d6ad83885c843d278990845ae/problem_builder/templates/html/mentoring_with_steps.html#L33-L36

OmarIthawi:
Thank you @pomegranited for your help! I really appreciate it.

PR #247, opened 2019-11-07
Murad/12189 feedback close btn

murad-hubib:
- Added close button icon in HTML i-e "&times;" and removed css definition

**How to test:**
- Select an option from options and a feedback window will be open
- If not, you can click on red/green icon to see it
- The window should display a close button
- There is a change in location of code so apparently it will not effect the visuals

**Ticket:** 
https://edx-wiki.atlassian.net/browse/MCKIN-12189

@xitij2000 Please review, squash and merge

### DIFF
--- a/problem_builder/public/css/questionnaire.css
+++ b/problem_builder/public/css/questionnaire.css
@@ -100,6 +100,11 @@
     font-size: 0;
 }
 
+.mentoring .questionnaire .choice-tips .close:before,
+.mentoring .questionnaire .feedback .close:before {
+    font-size: 20px;
+}
+
 .mentoring .choices-list .choice-selector {
     display: table-cell;
 }
--- a/problem_builder/public/css/questionnaire.css
+++ b/problem_builder/public/css/questionnaire.css
@@ -97,6 +97,7 @@
     font-family: FontAwesome;
     font-style: normal;
     text-align: center;
+    font-size: 0;
 }
 
 .mentoring .choices-list .choice-selector {
--- a/problem_builder/templates/html/tip_choice_group.html
+++ b/problem_builder/templates/html/tip_choice_group.html
@@ -3,4 +3,4 @@
     {{ tip_html|safe }}
   {% endfor %}
 </div>
-<button class="close icon-remove-sign fa fa-times-circle"></button>
+<button class="close icon-remove-sign fa fa-times-circle">&times;</button>
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,7 @@ from setuptools.command.install import install
 
 # Constants #########################################################
 
-VERSION = '3.4.4'
+VERSION = '3.4.7'
 
 # Functions #########################################################
 


Agrendalath:
Thank you for the contribution, @murad-hubib. I'll review this today.

murad-hubib:
> Thank you for the contribution, @murad-hubib. I'll review this today.

Thanks for heads up @Agrendalath

Agrendalath:
@murad-hubib, I tested this on Ironwood devstack and it didn't show up until I manually disabled `display: none` rule for the div in browser. Should I configure something to see it without such workaround?

This is how it looks on the `master` version.
master

This is how it looks after the changes.
changed
Is it supposed to look like this? It's pretty odd to have two "close" buttons there.

murad-hubib:
@Agrendalath Thanks for pointing out the redundancy of icons. I have removed the icon from html and now i will add it from mcka-theme. Please go ahead.

P.S: Yes the close icon is by default is hidden in xblock.

Agrendalath:
@murad-hubib, unfortunately I won't have time to review this today. I can do it on Sunday. Please resolve the conflicts in the meantime.

problem_builder/public/css/questionnaire.css
```diff
@@ -67,13 +67,13 @@
...
-    top: 5px;
-    right: 3px;
-    width: 18px;
-    height: 21px;
-    font-family: FontAwesome;
-    font-style: normal;
-    text-align: center;
+    top: 12px;
+    right: 12px;
+    font-size: 0;
```
Agrendalath: Unfortunately setting this to 0 makes the "close" button invisible in standard theme when `display: none` is disabled:
noicon

I also tested this with a custom theme and it seems to be working the same way.

murad-hubib: @Agrendalath I beg your pardon. I wrote SCSS instead CSS and that thing breaked into css file. I have fixed it. Please check now

Agrendalath: Ah, right. No worries!

Agrendalath:
@murad-hubib, please resolve the conflict - it has been changed from `div` to `button` in #248.

Agrendalath:
Also please bump the version in `setup.py`.

murad-hubib:
@Agrendalath Resolved conflict and bumped the version

Agrendalath:
@murad-hubib, thanks! I'll check this tomorrow.

problem_builder/public/css/questionnaire.css
```diff
@@ -90,13 +90,15 @@
     display: none;
     cursor: pointer;
     position: absolute;
-    top: 5px;
-    right: 3px;
-    width: 18px;
-    height: 21px;
-    font-family: FontAwesome;
-    font-style: normal;
-    text-align: center;
```
Agrendalath: Do we actually need to remove this? Wouldn't it be sufficient to just add `font-size: 0;` or am I missing something??

murad-hubib: @msaqib52 Please get this done by shahid to get this PR merged asap

murad-hubib: @Agrendalath Ah! I have removed all the css changed for this fix. Now we have left only 2 files with 2 changes. Please go ahead and merge after review

Agrendalath: Thanks for the changes, @murad-hubib. I'll review this tomorrow.

problem_builder/public/css/questionnaire.css
```diff
@@ -97,6 +97,7 @@
     font-family: FontAwesome;
     font-style: normal;
     text-align: center;
+    font-size: 0;
```
Agrendalath: I double-checked your other PR and isn't it actually redundant here? We already have `display:none` and you've overridden the font size [here](https://github.com/mckinseyacademy/mcka-theme/pull/227/files#diff-7f8e5c657cdce1929b229a9122c572feR56), so it seems safe to remove it as well.

problem_builder/templates/html/tip_choice_group.html
```diff
@@ -3,4 +3,4 @@
     {{ tip_html|safe }}
   {% endfor %}
 </div>
-<button class="close icon-remove-sign fa fa-times-circle"></button>
+<button class="close icon-remove-sign fa fa-times-circle">&times;</button>
```
Agrendalath: Shouldn't we use `icon-remove-sign fa fa-times-circle` **or** `&times;` to avoid workarounds to these duplicated buttons?

problem_builder/public/css/questionnaire.css
murad-hubib: @Agrendalath It's actually a tricky fix. Font awesome icon is rendering in :before element so the font size for the element doesn't matter. That's why I have given 20px font to the :before element.

problem_builder/templates/html/tip_choice_group.html
murad-hubib: Removing a class may break some css rules if any other party is also using this xblock. So I wanted to keep the things intact to avoid any issue

Agrendalath: Right, good point :+1:

Agrendalath:
@murad-hubib, thanks for explaining the approach. Please bump the version in `setup.py`. I cannot do this myself, because GitHub is marking `setup.py` as up-to-date for some reason.

murad-hubib:
> @murad-hubib, thanks for explaining the approach. Please bump the version in `setup.py`. I cannot do this myself, because GitHub is marking `setup.py` as up-to-date for some reason.

@Agrendalath Thanks for the understanding. I have bumped the version please go ahead.

Agrendalath:
@murad-hubib, done. I created the release too.

murad-hubib:
Thanks @Agrendalath for prompt response

PR #251, opened 2019-11-18
MCKIN-12152 AF#40-MRQ-Inform assistive technologies of changes in content

nasirhjafri:
MCKIN-12152 AF#40-MRQ-Inform assistive technologies of changes in content

[Issue]
When the submit button is activated, there is new content added to the page without notifying the user through an announcement.  Feedback text appears after the submit button.

[Resolution]
Upon activating the "Submit" button, additional content gets added beneath this button that provides feedback on the response submitted. Upon activating this button, please ensure to set focus to this newly added textual content using the JavaScript focus method, so that assistive technologies such as screen readers would be aware of this content update and would render this content to the screen reader users. Also, place a tabindex of -1 on this dynamically added feedback content container, so that it doesn't appear in the tab order.

### DIFF
--- a/problem_builder/public/js/mentoring_standard_view.js
+++ b/problem_builder/public/js/mentoring_standard_view.js
@@ -32,6 +32,7 @@ function MentoringStandardView(runtime, element, mentoring) {
             if (messagesDOM.html().trim()) {
                 messagesDOM.prepend('<div class="title1">' + mentoring.data.feedback_label + '</div>');
                 messagesDOM.show();
+                messagesDOM.focus();
             }
         }
 
--- a/problem_builder/templates/html/mentoring.html
+++ b/problem_builder/templates/html/mentoring.html
@@ -41,7 +41,7 @@
       <div class="attempts" data-max_attempts="{{ self.max_attempts }}" data-num_attempts="{{ self.num_attempts }}"></div>
     </div>
     {% endif %}
-    <div class="messages"></div>
+    <div class="messages" tabindex="-1"></div>
     <div class="assessment-review-tips"></div>
   </div>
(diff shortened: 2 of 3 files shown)


nasirhjafri:
@xitij2000 Please review.

xitij2000:
@nasirhjafri  I've added this to the review backlog, but unfortunately we will probably only be able to get to it next sprint since it wasn't included in the original PR list sent last sprint.

Agrendalath:
@nasirhjafri, please rebase it on the `master` and bump the version in `setup.py`.

nasirhjafri:
@Agrendalath Rebased with the master.

lgp171188:
@nasirhjafri, @Agrendalath, I will review the changes today and post my feedback after that. FYI.

lgp171188:
@nasirhjafri, I tried testing the changes in this PR on the Ironwood devstack. However, I cannot trigger the expected behaviour at all and the element containing the `messages` class is always hidden. So does this require testing on the custom LMS frontend?

lgp171188:
@nasirhjafri, I tested the changes on the custom LMS frontend opened in the Microsoft Edge browser, using the NVDA screen reader on Windows 10 and verified that it works as described. However, when using the built-in screen reader on Windows 10, it didn't work the same way. Since all the screen reader software use the same HTML elements and attributes, I am not sure how and why this happens.

If you have any idea about this, please let me know. Otherwise, this PR looks good to be merged. I will approve after checking your response.

nasirhjafri:
@lgp171188 I don't have an idea about Windows 10 screen reader. We can merge this one as the recommended screen reader is NVDA.

lgp171188:
@nasirhjafri, no problem. I will squash and merge the changes after the CI checks pass and then tag a release and ping you.

PR #253, opened 2019-12-04
MCKIN-12253 text provided for bookmark

musmanmalik:
Due to the lack of textual equivalents, assistive technology users such as screen readers will not be able to the meaning conveyed via CSS images as visual users can see or read. They are not able to know if the answer is correct or wrong.  Secondly, when the page is viewed removing the specified color for the page or in high contrast mode for readability purpose, this CSS icon will disappear from the page.  So low vision users who view the page in high contrast mode will not be able to see it on the screen to know if the answer is correct or wrong. 

So we are providing text in span tag with sr-only class. Which means these element will have no effect on UI but these span will be available to screen reader. This is for making these control more accessible to meet the accessibility criteria.

We are working on accessibility remediation that is why we are doing these changes.


musmanmalik:
@Agrendalath  Please review this pr.

Agrendalath:
@musmanmalik, thanks for the contribution! I'll review this tomorrow.

Agrendalath:
@musmanmalik, sorry, I need to postpone this by one day.

musmanmalik:
@Agrendalath  Okay , No problem

problem_builder/public/js/mentoring_with_steps.js
```diff
@@ -83,12 +83,18 @@ function MentoringWithStepsBlock(runtime, element) {
...
+            checkmark.append('<span class="sr-only">'+gettext("Correct")+'</span>')
         } else if (response.step_status === 'partial') {
             checkmark.addClass('checkmark-partially-correct icon-ok fa-check');
             checkmark.attr('aria-label', checkmark.data('label_partial'));
+            checkmark.find('span.sr-only').remove();
+            checkmark.append('<span class="sr-only">'+gettext("Partially correct")+'</span>')
         } else {
             checkmark.addClass('checkmark-incorrect icon-exclamation fa-exclamation');
             checkmark.attr('aria-label', checkmark.data('label_incorrect'));
+            checkmark.find('span.sr-only').remove();
```
Agrendalath: Could we extract it before the `if` instead of repeating it three times?

problem_builder/public/js/questionnaire.js
```diff
@@ -136,10 +136,12 @@ function MCQBlock(runtime, element) {
                     choiceDOM.addClass('correct');
                     choiceResultDOM.addClass('checkmark-correct icon-ok fa-check');
                     choiceResultDOM.attr('aria-label', choiceResultDOM.data('label_correct'));
+                    choiceResultDOM.append('<span class="sr-only">'+gettext("Correct")+'</span>');
                 } else {
                     choiceDOM.addClass('incorrect');
                     choiceResultDOM.addClass('checkmark-incorrect icon-exclamation fa-exclamation');
                     choiceResultDOM.attr('aria-label', choiceResultDOM.data('label_incorrect'));
+                    choiceResultDOM.append('<span class="sr-only">'+gettext("Incorrect")+'</span>');
```
Agrendalath: Don't we want to remove these spans for MCQ, as we are doing in `problem_builder/public/js/mentoring_with_steps.js`?

Agrendalath:
@musmanmalik, I tested this on my local devstack with browser's console and it works correctly. I just left two small questions.

problem_builder/public/js/mentoring_with_steps.js
musmanmalik: Yes I am Accommodating this. Thank you for pointing.

problem_builder/public/js/questionnaire.js
Agrendalath: Sure, thanks for explaining.

musmanmalik:
Thank you so much @Agrendalath  for your quick response. Please create a release tag as well.

Agrendalath:
@musmanmalik, sure, done :)

PR #254, opened 2019-12-12
MCKIN-12576 Stopped videos while moving to next step

nasirhjafri:
MCKIN-12576 Stopped videos while moving to the next step

If there is more than one step in assessment and we play a video, the video keeps playing in the background when we move to the next step.

### DIFF
--- a/problem_builder/public/js/mentoring_with_steps.js
+++ b/problem_builder/public/js/mentoring_with_steps.js
@@ -170,7 +170,7 @@ function MentoringWithStepsBlock(runtime, element) {
     }
 
     function stopVideos() {
-        $(element).find('video').trigger('pause')
+        $(element).find('video').trigger('pause');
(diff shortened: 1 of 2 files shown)


nasirhjafri:
@xitij2000 Please review it.

xitij2000:
@nasirhjafri Sure. I've put it in for review next sprint. 

That said I think this approach is too specific and only works for one specific type of video player. I think you should make it more generic so it works with other kinds of video embeds as well. You can do so  you can do so by using the underlying HTML video player API. i.e. find all HTML5 video elements and pause them.

nasirhjafri:
@xitij2000 Thanks, I've pushed the recommended changes. Please review it in the next sprint.

nasirhjafri:
Thanks, @xitij2000 for reviewing.

PR #256, opened 2020-01-06
MCKIN-12287 Feedback icons changed to buttons for MCQs and Multi rating questions

musmanmalik:
We are working on accessibility. This is also one the the reported. This fix has been done for MRQs case previously. Now This fix is for MCQs and multi rating questions.

At a basic level, all custom controls within a web application should provide identifying information about the component. This includes a name, type (role) and may contain state, value, and description information. When custom elements do not expose identifying information, users of assistive technology may not be able to identify and interact with the element. For example, without knowing the role or state of an element, the user of a screen reader may not know what actions can be performed on an element even if the element exposes a textual name. Without a proper role, assistive technology interactions that rely on roles being identified may not become available to the user.

The best approach is to use the <button> element and appropriate aria attributes to indicate all the required information.  This will make the icons keyboard focusable by default.

@xitij2000  @Agrendalath  Please review this pull request. Thanks.


problem_builder/templates/html/mcqblock.html
```diff
@@ -10,10 +10,10 @@ <h4 class="question-title" id="heading_{{ self.html_id }}">{{ self.display_name_
       <label class="choice-label"
              for="{{ self.html_id }}_{{ forloop.counter }}"
              aria-describedby="feedback_{{ self.html_id }} choice_tips_{{ self.html_id }}-{{ forloop.counter }}">
-        <span class="choice-result fa icon-2x"
+        <button class="choice-result fa icon-2x"
```
giovannicimolin: @Agrendalath Changing `span` to `button` has implications in `master`, where not styling is applied on top of the block.
Check [this comment](https://github.com/open-craft/problem-builder/pull/257#discussion_r366371230) for more context.

musmanmalik: @giovannicimolin  necessary style has already been applied while fixing this for MRQs. So that is its appearing without any UI disturbance here.

Agrendalath: Thanks for noticing, @giovannicimolin.
@musmanmalik, could you please ping me on this PR after resolving it?

musmanmalik: @Agrendalath @giovannicimolin  There are no implications of this since necessary style overridden in [for mrqs](https://github.com/open-craft/problem-builder/pull/249/files) and in this PR.
 
Only one scenario left [for](https://github.com/open-craft/problem-builder/pull/257) which i will fix and push in [this pr](https://github.com/open-craft/problem-builder/pull/257) soon.

xitij2000: @musmanmalik Note that @giovannicimolin is talking about breaking on the vanilla upstream edx/edx-platform LMS not Apros. Have you tested on the vanilla edx-platform? If not please do and fix the styling issues there as well.

musmanmalik: @xitij2000 yes I have tested on the vanilla edx-platform as well. It looks like for mrqs and mcqs

mcqs case
mrq case


Further I have also pushed style changes for [PR](https://github.com/open-craft/problem-builder/pull/257) for assessment case which now looks as following
assesment case

cc: @giovannicimolin

xitij2000: The checkmark padding looks off. Can you add back the vertical padding so it aligns properly with the button text?
Also, you can just use `background: none`.

musmanmalik: @xitij2000 Is that fine now?
after padding
after padding1

musmanmalik: @xitij2000  could you please look into this?

Agrendalath: @musmanmalik, I'll review this PR on Monday.

musmanmalik: @Agrendalath Please review this.

Agrendalath: This one seems to be outdated. Marking as resolved for now.

musmanmalik:
@Agrendalath Please review it again we have fixed identified issue. Thank you

musmanmalik:
@Agrendalath  I have pushed the resolution for 1,2. As for 3 is concerned it is not in the scope of this ticket. We have some follow up tickets which will deal the navigation order and your identified concern since its also not a ripple of this pr.

Agrendalath:
@musmanmalik, I confirm that the first and the second issue has been resolved.
Unfortunately the styling of `choice-tips-container` seems to be broken now. Please see the comparison below.
<details><summary><b>Master</b></summary><p>
image
</p></details>
<details><summary><b>This PR</b></summary><p>
image
</p></details>

musmanmalik:
@Agrendalath Please have a look now

PR #257, opened 2020-01-09
MCKIN-12768 Feedback icons changed to button for averall assessment case

musmanmalik:
We are working on accessibility. This is also one the the reported. This fix has been done for MRQs already and fro MCQs and multi rating questions in progress. In this pull request this is fixed for assessment checkmark.

At a basic level, all custom controls within a web application should provide identifying information about the component. This includes a name, type (role) and may contain state, value, and description information. When custom elements do not expose identifying information, users of assistive technology may not be able to identify and interact with the element. For example, without knowing the role or state of an element, the user of a screen reader may not know what actions can be performed on an element even if the element exposes a textual name. Without a proper role, assistive technology interactions that rely on roles being identified may not become available to the user.

The best approach is to use the element and appropriate aria attributes to indicate all the required information. This will make the icons keyboard focusable by default.

@xitij2000  @Agrendalath @lgp171188  Please review this pull request.


xitij2000:
@musmanmalik Please ping only me on such pull request since I am responsible for prioritising and assigning these PRs.

musmanmalik:
@xitij2000 Okay sure

problem_builder/templates/html/mentoring_with_steps.html
```diff
@@ -14,9 +14,9 @@ <h3>{{ title }}</h3>
     {% endfor %}
 
     <div class="submit">
-      <span class="step-overall-checkmark fa icon-2x fa-fw" aria-label=""
+      <button class="step-overall-checkmark fa icon-2x fa-fw" aria-label=""
```
giovannicimolin: @musmanmalik Switching from `span` to `button` changes the look of the element when running on `master`.


Changing the element to a button is not the best way of making it focusable.
The same functionality can be achieved by adding `tabindex="0"` to the element's properties.
Check more info [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex).

musmanmalik: @giovannicimolin Thanks for giving your review.

Can you please let me now where are you testing? Since I have have used Xblock-Sdk for testing where i not able to found this use case. 

"tabindex" could not be used here because it will disturb the element navigation because we are appending this module in various locations at various pages. "tabindex" is recommended to use when content not in actual page like popup modals or dialogue box case. 

So here what we do to accommodate the UI like MRQs,MCQs and multi rating question.

I am trying to reproduce this and will fix this UI issue.

Thank you for highlighting.

xitij2000: > Can you please let me now where are you testing? Since I have have used Xblock-Sdk for testing where i not able to found this use case.

We always test on edx-platform at the very least. It needs to work there perfectly since that is the primary audience.

>"tabindex" could not be used here because it will disturb the element navigation because we are appending this module in various locations at various pages. "tabindex" is recommended to use when content not in actual page like popup modals or dialogue box case.

A `tabindex` of 0 should not disturb navigation, it is the default value, and only means maintain the current tab order. 

That said a button is probably preferable with the addition of styling overrides.

giovannicimolin: @musmanmalik As I've mentioned before `tabindex="0"` doesn't interfere on the navigation order, just makes the object focusable.

>Can you please let me now where are you testing? Since I have have used Xblock-Sdk for testing where i not able to found this use case.

I'm testing the changes on a Open edX `ironwood` devstack.

musmanmalik: @giovannicimolin can you please check I have added some style in order to tackle identified issue in this case.

musmanmalik: @xitij2000  could you please look into this?

giovannicimolin: @musmanmalik I'll review this soon.

giovannicimolin: @musmanmalik I've tested this again, and the button styles are still being applied to the element.
Without any feedback, when selecting the button: :x:
hover_no_feedback
The screen reader tool reads "button" without any context when hovering over this.

With feedback, focusing the button: :x:
hover_feedback

With feedback: :heavy_check_mark: 
no_hover

Either add more styles to fix the issue or switch the element to use `tabindex=0`.

musmanmalik: @giovannicimolin  Point 1: Resolved

point 2: If we made span focus-able by tabindex  it will behave as button on button focus.
span having tabindex
span with tabindex


point 3 : Already good.

giovannicimolin: @musmanmalik Point 2 is to show that the feedback looks like a button on focus:
image
It's still looking like a button (see the screenshot above).

musmanmalik: @giovannicimolin  Thank you

I have fixed above issue. Could you please check this again.

problem_builder/public/css/questionnaire.css
```diff
@@ -141,3 +141,11 @@
     right: auto;
     left: 3px;
 }
+
+.assessment-question-block .submit button.step-overall-checkmark {
+    border: 0;
+    padding: 0;
+    background: none;
+    box-shadow: none !important;
+    outline: none;
```
giovannicimolin: @musmanmalik It's looking good now, but can you remove the `outline: none;` tag? It hurts accessibility.
After that I can approve, merge the PR and create a new tag.

musmanmalik: Thank you @giovannicimolin  Kindly merge and create a new release tag.

musmanmalik: @giovannicimolin  Please have a look. Thanks.

giovannicimolin: @musmanmalik I'm waiting to see the outcome of https://github.com/open-craft/problem-builder/pull/256 because it can affect the accessibility here.
Can you address the latest comments there?

musmanmalik: @giovannicimolin we have addressed the reported issues.

This PR is only for one checkmark/ as you already looking step-overall-checkmark. And this is not dependent on #256 so we can proceed this one.

Agrendalath:
@musmanmalik, as #256 PR is now merged, could you please rebase this PR on `master` branch? Also please bump the version in `setup.py` to ` v3.4.9`.
cc: @giovannicimolin

musmanmalik:
@Agrendalath @giovannicimolin  Please merge this and create release tag

PR #258, opened 2020-01-10
BOM-1146

awais786:
For last one month in edx-platform `test_migrations_are_in_sync` test was disabled due to some schema changes. Yesterday this test enabled and failing due to some new migrations changes under python3. It was due to the prefix `b` in few fields.

### DIFF
--- a/problem_builder/migrations/0001_initial.py
+++ b/problem_builder/migrations/0001_initial.py
@@ -17,9 +17,9 @@ class Migration(migrations.Migration):
                 ('name', models.CharField(max_length=50, db_index=True)),
                 ('student_id', models.CharField(max_length=32, db_index=True)),
                 ('course_id', models.CharField(max_length=50, db_index=True)),
-                ('student_input', models.TextField(default=b'', blank=True)),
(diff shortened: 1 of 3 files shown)


awais786:
@jmbowman Please review.

jmbowman:
Looks good to me; @bradenmacdonald , could we get a review and release from someone on your end?  We made a lot of similar fixes in edx-platform for the Python 3 rollout, this just makes sure that Django under Python 2 and 3 both agree that no new migration is needed.

bradenmacdonald:
👍 LGTM, thanks!

bradenmacdonald:
I've triggered a build; this is good to merge if the build is green.

jmbowman:
@bradenmacdonald It looks like the coverage error broke the deployment to PyPI; I'm guessing it has to do with an unpinned dependency on coverage, which recently released version 5 that changed the data file format.

bradenmacdonald:
@jmbowman Thanks for the heads up. Hopefully this will fix it: https://github.com/open-craft/problem-builder/pull/259

PR #259, opened 2020-01-10
Pin coverage to v4

bradenmacdonald:
Quick PR to see if pinning `coverage` will fix the CI issues that arose in https://github.com/open-craft/problem-builder/pull/258#issuecomment-573208358

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -68,7 +68,7 @@ jobs:
             sudo apt-get update
             virtualenv venv
             source venv/bin/activate
-            pip install coverage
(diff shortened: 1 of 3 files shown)


lgp171188:
@bradenmacdonald in the `v4.0.0` release that we have created for Django 2.2 support, we had to upgrade `coverage` to `5.0.3`, the same as the version pinned in the `xblock-sdk` repository, to get the `problem-builder` CI checks, the coverage job in particular, to pass on that branch. FYI.

PR #272, opened 2020-04-14
Fixing django2 compatible items.

awais786:
edX devstack giving errors with some missing django2 migrations/models requirements
[on_delete required in Django 2.0](https://docs.djangoproject.com/en/1.11/ref/models/fields/#django.db.models.ForeignKey)


bradenmacdonald:
This has been fixed in https://github.com/open-craft/problem-builder/tree/problem-builder-v4 / tag `v4.0.0`. Please install that version.

Note that as far as I know, problem-builder isn't included on default edX devstacks, so it's up to you to update your own installation.

PR #274, opened 2020-04-17
MCKIN-15673 Japanese translations corrections

musmanmalik:
Japanese translations corrections


problem_builder/step.py
```diff
@@ -251,8 +251,13 @@ def resource_string(path):
 
     def get_translation_content(self):
         try:
+            language = utils.translation.get_language().split('-')
+            if len(language) == 2:
+                new_lang = language[0] + "_" + language[1].upper()
+            else:
+                new_lang = language
             return self.resource_string('public/js/translations/{lang}/textjs.js'.format(
-                lang=utils.translation.get_language(),
```
viadanna, line 255: Didn't this work for Japanese? Why was it removed in favor of a manual approach?

problem_builder/public/js/translations/ja_JP/textjs.js
```diff
@@ -29,7 +29,11 @@
     "Results retrieved on <%= creation_time %> (<%= seconds %> second).": [
       "\u7d50\u679c\u3092<%= creation_time %> (<%= seconds %> \u79d2) \u3067\u53d6\u5f97\u3057\u307e\u3057\u305f\u3002"
     ], 
-    "The report is currently being generated\u2026": "\u73fe\u5728\u30ec\u30dd\u30fc\u30c8\u3092\u4f5c\u6210\u4e2d\u3067\u3059\u2026"
+    "The report is currently being generated\u2026": "\u73fe\u5728\u30ec\u30dd\u30fc\u30c8\u3092\u4f5c\u6210\u4e2d\u3067\u3059\u2026",
```
viadanna, line 32: While I can test this and review it works, I cannot attest the translation.

problem_builder/step.py
musmanmalik: Here we are receiving language code like ja-jp, de-DE while we have repos ja_JP ,de_DE , etc. If we don't do like this the default en is returning. That is we need to do this.

musmanmalik: So for four character language code we need to change last 3 characters. `-` to `_` last two characters to upper case e.g. `jp` to `JP`.

problem_builder/public/js/translations/ja_JP/textjs.js
musmanmalik: We got these translations from trusted translator.

problem_builder/step.py
bradenmacdonald: Might be good to have a comment here to make that more clear.

musmanmalik: @bradenmacdonald  Thank you for highlighting. I have added a comment. Please proceed.
Thanks

problem_builder/public/js/translations/ja_JP/textjs.js
```diff
@@ -29,7 +29,11 @@
...
+    "You have used {num_used} of 1 submission.": [
+      "上限{max_attempts}回のうち {num_used} 回提出しました。",
+      "上限1回のうち {num_used} 回提出しました。"
+    ],
```
viadanna, line 36: Not sure I understand correctly, but I think you have plural first, when it's expected to be second, e.g. https://github.com/open-craft/problem-builder/blob/d024a6ce2024adfe07facb72822aed16137d441f/problem_builder/public/js/translations/fr_CA/textjs.js#L34-L36

musmanmalik:
@bradenmacdonald @viadanna  Please review this. A test is failing which seems not related to the changes in this pr. This may be a flaky test case. Rest if all is good kindly merge this and create a release tag. 

Thanks

problem_builder/public/js/translations/ja_JP/textjs.js
musmanmalik: @viadanna  have you tested this one? I have tested this and working fine as written in this pr.

viadanna: image
This was the result for a setup with 3 attempts.

viadanna:
Alright, merged and tagged, @musmanmalik

problem_builder/public/js/translations/ja_JP/textjs.js
musmanmalik: This is result is after first attempt out of three

musmanmalik: `"You have used 1 of 3 submissions" needs to translate to "上限3回のうち１回提出しました。"`

Translator has given us the same

PR #276, opened 2020-04-23
MCKIN-15914 Tips should not show if empty

moeez96:
Jira ticket: https://edx-wiki.atlassian.net/browse/MCKIN-15914

Issue:
Even if tips are empty for a choice in an MRQ, html is added to `tip_choice_group.html` anyway. This shows an empty tip box to user which is undesired UX.

Solution:
Do not add html in tips if not tips exist for [condition here](https://github.com/open-craft/problem-builder/blob/3badc1c33783e0ae44017dad64352fbf72eb6930/problem_builder/public/js/questionnaire.js#L253) to hold true.


moeez96:
@xitij2000 Please review, merge and tag.

xitij2000:
@moeez96 Problem Builder master is now at v4, however v4 is not compatible with solutions currently. So please create a PR for the problem-builder-v3 branch as well. 

Till solutions is compatible with v4 every PR to v3 should come with a master PR as well so that they are in sync.

moeez96:
@xitij2000 Base for this PR has been set to problem-builder-v3
A new [PR](https://github.com/open-craft/problem-builder/pull/280) has been opened with master as base.

rusrushal13:
@xitij2000  @moeez96 Thanks for the PR, I have checked it on both the devstacks(v3 and master) LGTM :+1:

moeez96:
@rusrushal13 @xitij2000 Please merge, and tag since I dont have access.

PR #280, opened 2020-04-30
MCKIN-15914 Tips should not show if empty

moeez96:
Jira ticket: https://edx-wiki.atlassian.net/browse/MCKIN-15914
Related PR: https://github.com/open-craft/problem-builder/pull/276

Issue:
Even if tips are empty for a choice in an MRQ, html is added to tip_choice_group.html anyway. This shows an empty tip box to user which is undesired UX.

Solution:
Do not add html in tips if not tips exist for condition here to hold true.


rusrushal13:
@xitij2000 @moeez96 I have checked it, it works fine, LGTM :+1:

moeez96:
@rusrushal13 @xitij2000 Please merge, and tag since I dont have access.

PR #282, opened 2020-05-08
[BB-2320] Added the MySQL Readme for Problem builder

rusrushal13:
Issue: [BB-2320](https://tasks.opencraft.com/browse/BB-2320)

Testing Instructions:

* Try running problem builder as given in the README, check everything is running fine.

Reviewers: 
- [ ] @xitij2000

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,13 @@ Firstly, create a virtualenv:
 ~/xblock_development $ . venv/bin/activate
 ```
 
+The MySQL client library headers are required for the `mysqlclient` library used by `problem-builder`.
+On Ubuntu 16.04, it can be installed by running
+
+```bash
+sudo apt-get install libmysqlclient-dev
+```
+
 Now run the following commands from the problem builder repo
 root to install the problem builder dependencies:
 
@@ -91,8 +98,24 @@ DATABASES['default']['NAME'] = 'var/workbench.db'
 ```
 
 Because this XBlock uses a Django model, you need to sync the database
-before starting the workbench. Run this from the XBlock repository
-root:
+before starting the workbench. `problem-builder` uses MySQL 5.6 and hence you can quickly spin up
+an instance of MySQL 5.6 with docker. Run the following command to start a MySQL 5.6 server.
+
+```bash
+docker run --rm -it -p 3306:3306 -e MYSQL_ROOT_PASSWORD=rootpw -e MYSQL_DATABASE=db circleci/mysql:5.6 mysqld --character-set-server=latin1 --collation-server=latin1_swedish_ci
+```
+
+By default, the `xblock-sdk` uses the SQLite database by default but MySQL
+can be used by specifying an environment variable `WORKBENCH_DATABASES` in
+the following format.
+
+```bash
+export WORKBENCH_DATABASES='{"default": {"ENGINE": "django.db.backends.mysql", "NAME": "db", "USER": "root", "PASSWORD": "rootpw", "HOST": "127.0.0.1", "OPTIONS": {"charset": "utf8mb4"}}}'
+```
+
+Ensure that the database name and credentials match the ones configured in the docker container.
+
+Run this from the XBlock repository root:
 
 ```bash
 $ ./manage.py migrate --settings=workbench.settings_pb
--- a/README.md
+++ b/README.md
@@ -1,5 +1,4 @@
-Problem Builder and Step Builder
---------------------------------
+## Problem Builder and Step Builder
 
 [![Circle CI](https://circleci.com/gh/open-craft/problem-builder.svg?style=svg)](https://circleci.com/gh/open-craft/problem-builder)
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -91,8 +98,24 @@ DATABASES['default']['NAME'] = 'var/workbench.db'
 ```
 
 Because this XBlock uses a Django model, you need to sync the database
-before starting the workbench. Run this from the XBlock repository
-root:
+before starting the workbench. `problem-builder` uses MySQL 5.6 and hence you can quickly spin up
+an instance of MySQL 5.6 with docker. Run the following command to start a MySQL 5.6 server.
+
+```bash
+docker run --rm -it -p 3306:3306 -e MYSQL_ROOT_PASSWORD=rootpw -e MYSQL_DATABASE=db circleci/mysql:5.6 mysqld --character-set-server=latin1 --collation-server=latin1_swedish_ci
```
xitij2000, line 105: Is there a reason to not just use the standard mysql:5.6 image?
I think you can just as easily use:
```docker run --rm -it -p 3306:3306 -e MYSQL_ROOT_PASSWORD=rootpw -e MYSQL_DATABASE=db mysql:5.6```

File: README.md [hunk 2]
```diff
@@ -91,8 +98,24 @@ DATABASES['default']['NAME'] = 'var/workbench.db'
 ```
 
 Because this XBlock uses a Django model, you need to sync the database
-before starting the workbench. Run this from the XBlock repository
-root:
+before starting the workbench. `problem-builder` uses MySQL 5.6 and hence you can quickly spin up
```
xitij2000, line 101: I think it would be clearer to say:

> Testing `problem-builder` in the workbench requires MySQL instead of the standard SQLite configuration it uses. You can quickly spin up an instance of MySQL with Docker using the following command:

README.md
xitij2000: Additionally I would recommend not using the default port in case it's already used by an existing instance of mysql. So:

```suggestion
docker run --rm -it -p 3307:3306 -e MYSQL_ROOT_PASSWORD=rootpw -e MYSQL_DATABASE=db mysql:5.6
```

README.md
```diff
@@ -91,8 +98,24 @@ DATABASES['default']['NAME'] = 'var/workbench.db'
...
+an instance of MySQL 5.6 with docker. Run the following command to start a MySQL 5.6 server.
+
+```bash
+docker run --rm -it -p 3306:3306 -e MYSQL_ROOT_PASSWORD=rootpw -e MYSQL_DATABASE=db circleci/mysql:5.6 mysqld --character-set-server=latin1 --collation-server=latin1_swedish_ci
+```
+
+By default, the `xblock-sdk` uses the SQLite database by default but MySQL
```
xitij2000, line 108: By default is coming twice here. If we mention the reliance on SQLite before, we can just say that the database config can be overridden using environment variables. 

That said, the readme explicitly has you define the database in line 97, so this setting will just not be used. You need to remove that line as well.

README.md
```diff
@@ -91,8 +98,24 @@ DATABASES['default']['NAME'] = 'var/workbench.db'
...
+can be used by specifying an environment variable `WORKBENCH_DATABASES` in
+the following format.
+
+```bash
+export WORKBENCH_DATABASES='{"default": {"ENGINE": "django.db.backends.mysql", "NAME": "db", "USER": "root", "PASSWORD": "rootpw", "HOST": "127.0.0.1", "OPTIONS": {"charset": "utf8mb4"}}}'
```
xitij2000, line 113: Again, let's keep it simple and not add unnecessary options or settings. Even the default edxapp doesn't set a custom charset. 
```suggestion
export WORKBENCH_DATABASES='{"default": {"ENGINE": "django.db.backends.mysql", "NAME": "db", "USER": "root", "PASSWORD": "rootpw", "HOST": "127.0.0.1", "PORT": 3307}}'
```

README.md
rusrushal13: @xitij2000 I have tested it with standard SQL image, and it works, thanks.
Updated the port too :+1:

README.md
rusrushal13: Thanks, I have updated it :+1:

README.md
rusrushal13: I tried without options too, it works, I was taking the circle ci configurations for my database configs directly :)

rusrushal13:
> I think the readme doesn't work as of yet. I faced a number of failures:
> 
>     1. The xblock-sdk's `make install` command files. The incorrect version is used here, it should be 0.1.8 instead of 0.1.7
> 
>     2. The migration command failed since the current readme asks you to override `DATABASES['default']['NAME'] = 'var/workbench.db'`

@xitij2000 Thanks for the review, I had some issues, which @lgp171188 helped me to resolve it; Now I have tested it on an Ubuntu 16.04(Xenial64) VM with python 3.5.2 installed and it works there fine. Let me know if you face any issues :)

README.md
```diff
@@ -87,22 +97,36 @@ Create the following configuration file in `workbench/settings_pb.py`:
...
-Because this XBlock uses a Django model, you need to sync the database
-before starting the workbench. Run this from the XBlock repository
-root:
+By default, the `xblock-sdk` uses the SQLite database but MySQL
+can be used by specifying an environment variable `WORKBENCH_DATABASES` in
+the following format.
+
+```bash
+export WORKBENCH_DATABASES='{"default": {"ENGINE": "django.db.backends.mysql", "NAME": "db", "USER": "root", "PASSWORD": "rootpw", "HOST": "127.0.0.1", "PORT": "3306"}}'
+```
```
lgp171188, line 108: @rusrushal13, I think this paragraph, can be moved to after the `docker run` command. What do you say?

README.md
rusrushal13: Thanks, I have updated it :)

README.md
rusrushal13: I have updated the paragraph.

README.md
rusrushal13: Thanks, yeah this makes it clearer, I have updated it :D

README.md
rusrushal13: First I thought the same, but running the export command before running the database looks better to me, that's why I have kept it before the docker command. What do you think?

README.md
```diff
@@ -94,38 +97,36 @@ Create the following configuration file in `workbench/settings_pb.py`:
...
-Because this XBlock uses a Django model, you need to sync the database
-before starting the workbench. `problem-builder` uses MySQL 5.6 and hence you can quickly spin up
-an instance of MySQL 5.6 with docker. Run the following command to start a MySQL 5.6 server.
+By default, the `xblock-sdk` uses the SQLite database but MySQL
+can be used by specifying an environment variable `WORKBENCH_DATABASES` in
+the following format.
 
 ```bash
-docker run --rm -it -p 3306:3306 -e MYSQL_ROOT_PASSWORD=rootpw -e MYSQL_DATABASE=db circleci/mysql:5.6 mysqld --character-set-server=latin1 --collation-server=latin1_swedish_ci
+export WORKBENCH_DATABASES='{"default": {"ENGINE": "django.db.backends.mysql", "NAME": "db", "USER": "root", "PASSWORD": "rootpw", "HOST": "127.0.0.1", "PORT": "3306"}}'
```
xitij2000, line 107: The port is set to 3306 here but 3307 later.

xitij2000:
:+1: The changes look good. Could you also update xblock-sdk submodule here to the same version as that in the requirements file? Other than that it is good to go. 
- [x] I tested this: tested setting up based on readme
- [x] I read through the code
- [x] Includes documentation

README.md
rusrushal13: Thanks for the catch, I have updated the PORT :)

PR #287, opened 2020-06-16
fix js translations for answer block

msaqib52:


taniwha:
@msaqib52 @xitij2000 Despite this being the same patch (minus version number) as for v3, I am unable to verify that it works. I have checked eo as well as ko-kr, both show untranslated English for that field. The v3 branch works well.

PR #291, opened 2020-07-23
MCKIN-16959 Korean Translation Update-v3

musmanmalik:
Korean Translation Update-v3


musmanmalik:
@xitij2000 Please review.

PR #294, opened 2020-08-19
MCKIN-21527 Problem Builder (FTE, Assessment, MCQ, MRQ) - On opening these modules, none of the text is translated on notifications dropdown.

wasifarbisoft:
**Issue:**
Problem Builder's `gettext` and `ngettext` override the global translations of the edx app. Hence strings of apps added outside the scope of problem builder doesn't get translated specially edx-notification, this is due to the step_util.js and util.js in which `gettext` and `ngetext` are declared globally.

**Solution:**  
New variables are introduced for the problem builder i.e `pb_gettext` and `pb_ngettext` so that global `gettext` and `ngettext` don't get overridden and their usage is updated elsewhere.

### DIFF
--- a/problem_builder/instructor_tool.py
+++ b/problem_builder/instructor_tool.py
@@ -176,9 +176,7 @@ class InstructorToolBlock(XBlock, I18NService):
                 new_lang = language[0] + "_" + language[1].upper()
             else:
                 new_lang = utils.translation.get_language()
-            return self.resource_string('public/js/translations/{lang}/textjs.js'.format(
-                lang=new_lang
-            ))
+            return self.resource_string('public/js/translations/{lang}/textjs.js'.format(lang=new_lang))
         except IOError:
             return self.resource_string('public/js/translations/en/textjs.js')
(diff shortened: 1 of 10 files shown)


wasifarbisoft:
@xitij2000  Can you please review this pull request?

wasifarbisoft:
@xitij2000 
I have updated the Pull Request. Please review it.

xitij2000:
@wasifarbisoft I just tested this from scratch, and I don't think this approach will work. While it does solve the problem from browser side, it breaks extracting translations. I think the best approach here is to make the gettext and ngettext declaraton from being global to local to the function, instead of renaming them everywhere, which would mean that translations can no longer be extracted because the tooling searches for gettext not pb_gettext.

wasifarbisoft:
Thanks @xitij2000 ,

I am closing this Pull Request and based on your last comments I have created a new Pull Request
https://github.com/open-craft/problem-builder/pull/298

Please review that thanks!

PR #267, opened 2020-03-09
MCKIN-12690 Fix Score rounding

moeez96:
Jira ticket: https://edx-wiki.atlassian.net/browse/MCKIN-12690
Python 3:
Decimal(0.485) = 0.4849999
Decimal(48.5) = 48.5

Due to above mentioned, rounding is in-accurate.
We need to multiply score by 100 before giving it as input to Decimal(xxx).

PR #296, opened 2020-08-24
MCKIN-22514 rounding off correctly

mudassir-hafeez:
Jira ticket https://edx-wiki.atlassian.net/browse/MCKIN-22514

this PR keeps up to date master for #297 
> There is an endpoint that was missed in #267 which needs to be addressed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,7 @@ from setuptools.command.install import install
 
 # Constants #########################################################
 
-VERSION = '4.0.6'
+VERSION = '4.0.7'
 
 # Functions #########################################################
 
--- a/problem_builder/mentoring.py
+++ b/problem_builder/mentoring.py
@@ -932,7 +932,7 @@ class MentoringWithExplicitStepsBlock(BaseMentoringBlock, StudioContainerWithNes
 
         return Score(
             float(score),
-            int((Decimal(score) * 100).quantize(Decimal('1.'), rounding=ROUND_HALF_UP)),
+            int(Decimal(score * 100).quantize(Decimal('1.'), rounding=ROUND_HALF_UP)),
             correct,
             incorrect,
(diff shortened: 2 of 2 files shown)


xitij2000:
@mudassir-hafeez I think solutions still uses version 3 so you will also need a PR against the v3 branch.

setup.py
```diff
@@ -28,7 +28,7 @@
 
 # Constants #########################################################
 
-VERSION = '4.0.6'
+VERSION = '3.4.23'
```
aayushagra, line 31: @mudassir-hafeez Why is this downgraded?

aayushagra (commented):
Hi @mudassir-hafeez can you please explain the following change:

setup.py
mudassir-hafeez: @aayushagra thank you for taking my attention towards it. Actually, if you see releases, we were using two different versions of releases. one is for Ironwood Open edX releases and other is for normal releases. After Ironwood upgrade, releases should be as they were before. @xitij2000 can you please comment on that what should be the right release version from now.

mudassir-hafeez:
Oh got it, we are using 4 series of releases for master and 3 series against the v3 branch

mudassir-hafeez:
@aayushagra I have updated the version accordingly. can you please review it once more? thanks

aayushagra:
@mudassir-hafeez Looks good. Can you please squash your changes down to one commit before merging?

aayushagra:
:+1:

- [x] I tested this: Tested that the code works as intended and tests pass
- [x] I read through the code
- [] I checked for accessibility issues
- [] Includes documentation
- [] I made sure any change in configuration variables is reflected in the corresponding client's `configuration-secure` repository.

PR #297, opened 2020-08-25
MCKIN-22514 round off correctly for v3

mudassir-hafeez:
Jira ticket https://edx-wiki.atlassian.net/browse/MCKIN-22514

There is an endpoint that was missed in #267 which needs to be addressed.

### DIFF
--- a/problem_builder/mentoring.py
+++ b/problem_builder/mentoring.py
@@ -932,7 +932,7 @@ class MentoringWithExplicitStepsBlock(BaseMentoringBlock, StudioContainerWithNes
 
         return Score(
             float(score),
-            int((Decimal(score) * 100).quantize(Decimal('1.'), rounding=ROUND_HALF_UP)),
+            int(Decimal(score * 100).quantize(Decimal('1.'), rounding=ROUND_HALF_UP)),
             correct,
             incorrect,
             partially_correct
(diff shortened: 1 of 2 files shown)


mudassir-hafeez:
@xitij2000 here is the PR for v3. Can you please review it?

aayushagra (approved):
Good to go

mudassir-hafeez:
Thank you @aayushagra for reviewing this PR. we have the same #296  for the master branch. Can you please review that too?

aayushagra:
:+1:

- [x] I tested this: Tested that changes work and all tests pass
- [x] I read through the code
- [ ] I checked for accessibility issues
- [ ] Includes documentation
- [ ] I made sure any change in configuration variables is reflected in the corresponding client's `configuration-secure` repository.

mudassir-hafeez:
@aayushagra It seems like you didn't create the release tag version as it should be. I think it would be a typo and should be `v3.4.23` prefix semantic version with “v” rather than this `3.4.23`, and same for 4.0.7

xitij2000:
@mudassir-hafeez I've updated the tags. We've been a bit inconsistent about this at times.

mudassir-hafeez:
@xitij2000 thank you for updating the tags.

PR #310, opened 2020-09-28
MCKIN-25170 Japanese string translation update

musmanmalik:
Japanese string translation update


aayushagra (changes requested):
@musmanmalik 

This fails when i run compile_translations. See errors below:

Execution of msgfmt failed: /home/i7/opencraft/extra/problem-builder/problem_builder/locale/ar/LC_MESSAGES/text.po:1554: 'msgstr[1]' is not a valid Python format string, unlike 'msgid_plural'. Reason: The character that terminates the directive number 1 is not a valid conversion specifier.

CC: @xitij2000

musmanmalik:
@aayushagra Thanks for highlighting, I have rectified. Please proceed.

aayushagra (approved):
:+1:

- [x] I tested this: I read the code, checked out and ran compile_translations on my local system, checked that all tests pass, installed it on my juniper devstack and made sure problem-builder still works, changed my language to arabic and japanese and ensured problem-builder still works correctly
- [x] I read through the code
- [ ] Includes documentation
- [ ] I made sure any change in configuration variables is reflected in the corresponding client's `configuration-secure` repository.

PR #312, opened 2020-10-08
Add translations to the package data

andrey-canon:
## Description
Since the translation is not included in the package data the following error is raised` FileNotFoundError: [Errno 2] No such file or directory: '/edx/app/edxapp/venvs/edxapp/lib/python3.5/site-packages/problem_builder/translations/config.yaml'`, so this forces to  install  the package from github in editable mode.

This pr includes the translation in the package data an remove locale

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -103,7 +103,7 @@ setup(
     entry_points={
         'xblock.v1': BLOCKS,
     },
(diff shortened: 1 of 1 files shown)


aayushagra:
@andrey-canon I had to revert the PR since it needs a version bump. Since I don't have access to your branch, would you be able to add a commit and bump the version in setup.py? Alternatively i can do it but better if it's part of the same PR

PR #319, opened 2020-11-24
Fix deprecated import path

bradenmacdonald:
Removes this warning on newer edX versions (Koa and forward):

```
2020-11-24 20:33:43,166 WARNING 4695 [py.warnings] [user None] [ip None] warnings.py:109 - /edx/src/problem-builder/problem_builder/models.py:29: DeprecatedEdxPlatformImportWarning: Importing student instead of common.djangoapps.student is deprecated
```

Backwards compatible with Juniper and Ironwood.

### DIFF
--- a/problem_builder/models.py
+++ b/problem_builder/models.py
@@ -20,14 +20,21 @@
 
 # Imports ###########################################################
 
+from django.conf import settings
 from django.contrib.auth.models import User
 from django.db import models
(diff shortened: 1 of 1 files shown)


xitij2000 (approved):
:+1:
- [x] I tested this: tested on master devstack to make sure the warning disappears and the block works otherwise. 
- [x] I read through the code

PR #324, opened 2021-01-29
fix: deprecated edx-platform import references

kdmccormick:
Commit:
```
For Lilac and beyond, make sure the following:
* courseware
* static_replace
* xblock_django

are instead imported as:
* lms.djangoapps.courseware
* common.djangoapps.static_replace
* common.djangoapps.xblock_django

refactor: centralize logic of edx-platform imports
in new module platform_dependencies.py.

Bump version to 4.1.11
```

Check out https://github.com/open-craft/problem-builder/pull/323 for context on the underlying import issue and our approach for handling it across different Open edX releases.

This PR just takes the strategy from #323 and applies it to three other instances of deprecated edx-platform imports. It centralizes the logic in a new file `platform_dependencies.py` as to avoid duplicating the nested `try-except` block across the project.

Related to the [import shims removal](https://github.com/edx/edx-platform/pull/25932)

### DIFF
--- a/problem_builder/v1/upgrade.py
+++ b/problem_builder/v1/upgrade.py
@@ -37,12 +37,16 @@ import six
 from lxml import etree
 from six import StringIO
 
-from courseware.models import StudentModule
 from mentoring import MentoringBlock
 from problem_builder.mentoring import MentoringBlock as NewMentoringBlock
 
+from .platform_dependencies import StudentModule
 from .studio_xml_utils import studio_update_from_node
 from .xml_changes import convert_xml_to_v2
+
+
+if not StudentModule:
+    raise ImportError("Could not import StudentModule from edx-platform courseware app.")
 
 
 def upgrade_block(store, block, from_version="v1"):
--- a/problem_builder/mentoring.py
+++ b/problem_builder/mentoring.py
@@ -349,13 +349,11 @@ class MentoringBlock(
         except ImportError:
             pass
 
-        try:
-            from xblock_django.models import XBlockConfiguration
+        from .platform_dependencies import XBlockConfiguration
+        if XBlockConfiguration:
             opt = XBlockConfiguration.objects.filter(name="pb-swipe")
             if opt.count() and opt.first().enabled:
                 additional_blocks.append(SwipeBlock)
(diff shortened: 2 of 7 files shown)


problem_builder/v1/upgrade.py
```diff
@@ -37,14 +37,18 @@
...
 from mentoring import MentoringBlock
 from problem_builder.mentoring import MentoringBlock as NewMentoringBlock
 
+from .platform_dependencies import StudentModule
 from .studio_xml_utils import studio_update_from_node
 from .xml_changes import convert_xml_to_v2
 
 
+if not StudentModule:
+    raise ImportError("Could not import StudentModule from edx-platform courseware app.")
```
kdmccormick, line 49: Previous to this PR, `upgrade.py` would raise an ImportError if edx-platform is not importable.

`from .platform_dependencies import StudentModule`, though, will fail silently.

So, I added this if-statement to replicate the old error behavior.

kdmccormick:
@bradenmacdonald , when you have a chance, could you also take a look at this PR? I _think_ these should be the last deprecated imports in problem-builder.

bradenmacdonald:
Sure, I can take a look soon. Thanks for these, @kdmccormick !

@xitij2000 FYI on this PR, would you [also] like to review it?

bradenmacdonald (approved):
:+1:

- [x] I tested this: quick test:

```
$ make lms-shell
# ./manage.py lms shell 
>>> from problem_builder.platform_dependencies import StudentModule
>>> StudentModule
<class 'lms.djangoapps.courseware.models.StudentModule'>
>>> from problem_builder.models import AnonymousUserId
>>> AnonymousUserId
<class 'common.djangoapps.student.models.AnonymousUserId'>
```

- [x] I read through the code
- [x] I checked for accessibility issues: n/a
- [x] Includes documentation: nice comments yes

xitij2000:
The only comment I'd have here is that in most similar situations we put such code in a file called `compat.py`. That said I think `platform_dependencies` is clearer, so all good.

kdmccormick:
Thanks all!

PR #330, opened 2021-08-15
limit the html template name to class names that are declared in the …

lpm0073:
…problem-builder package. 
The current logic for assigning the html template name only works for classes that are declared in the problem-builder repository. If, on the other hand, you subclass a descendent of QuestionnaireAbstractBlock in your own XBlock project then the logic breaks.

My proposed logic goes as follows:
the html template_name must match the lower-case name of *this* class, AND, 
the html needs to exist in this repository, meaning that *this* class needs
to be a class in this repo that inherits QuestionnaireAbstractBlock, noting however
that *we* could be anything that subclassed QuestionnaireAbstractBlock, within
or outside of this repo. The current logic does not consider a class that is external
to problem-builder itself.

We therefore need to filter the class hierarchy of *this* in order to disregard classes 
that are not part of the problem-builder package. Python's mro() meta
function presents the class hierarchy in the correct order, thus, the correct class 
name to use for the html template will be the first item in the filtered list.


lpm0073:
Looks like your circleci tests are also encouraging me to update the requirements, thus, off i go to take care of that. back soon with another attempt.

lpm0073:
CircleCI reports the following update error:

E: Repository 'http://deb.debian.org/debian buster-updates InRelease' changed its 'Suite' value from 'stable-updates' to 'oldstable-updates'
N: This must be accepted explicitly before updates for this repository can be applied. See apt-secure(8) manpage for details.

This appears to be something which OpenCraft needs to address before CircleCI actually evaluates the details of my pull request.

arbrandes:
@lpm0073, looks like you closed this before I got to it.  Does this mean you're no longer interested in pursuing this idea?  Let me know if you are, and I'll see to finding an internal reviewer.

lpm0073:
Hi Adolfo, no, on the contrary, I'm very focused on Problem Builder. I only
deleted the pull request bc I'd only intended for it to be a 1-line
modification.

I'll isolate this one modification and resubmit

lpm0073:
Adolfo, btw, each of the three circleci build steps fail due to something happening within the linux instance that hosts the circleci processes. As best I can tell, this is not something that I can control nor remedy. Is that correct?



Lawrence McDaniel
__________________________________
📡  Full-Stack Developer
CV: lawrencemcdaniel.com | linkedin.com/in/lawrencemcdaniel
Blog: blog.lawrencemcdaniel.com
Code Samples: horrors.lawrencemcdaniel.com | github.com/lpm0073
Contact: ***@***.*** | +1 (617) 834-6172
Address: 210 Broadway, Suite 201, Cambridge, MA 02139

PR #331, opened 2021-08-18
fix generation of the template path filename so that it correctly eva…

lpm0073:
…luates subclasses defined in a different XBlock, outside of this repository.

the naming convention of html template_name is, it must match the lower-case name of *this* class, AND
obviously, the html template itself should actually exist in this repository. Note however, that the current logic 
incorrectly evaluates a comprehensive list of classes in the hierarchy, including for example, classes 
that are defined outside of this repository, yielding an invalid template_name and a resulting internal "file not found" exception.

This pull request refactors the logic to introspect the class inheritance chain such that it
will exclude classes in the hierarchy that are not actually defined inside this problem-builder repository.

Methodology: Python's built-in mro() class hierarchy introspection method generates a clean, complete 
list of class inheritance. The list is ordered with "self" in position [0] and the Object class in position [-1]. 
After filtering this list for the case module_name == "problem_builder" we want the first item in the list. That is, we want the last class defined in the problem-builder repository to have been inherited from QuestionnaireAbstractBlock.

Expected results of this logic will/should be (as of this commit): mcqblock.html or mrqblock.html.

### DIFF
--- a/problem_builder/questionnaire.py
+++ b/problem_builder/questionnaire.py
@@ -102,15 +102,25 @@ class QuestionnaireAbstractBlock(
             return self.resource_string('public/js/translations/en/textjs.js')
 
     def student_view(self, context=None):
-        name = getattr(self, "unmixed_class", self.__class__).__name__
-
-        template_path = 'templates/html/{}.html'.format(name.lower())
-
         context = context.copy() if context else {}
         context['self'] = self
         context['custom_choices'] = self.custom_choices
         context['hide_header'] = context.get('hide_header', False) or not self.show_title
 
+        # the html template_name must match the lower-case name of *this* class, AND, 
+        # the html needs to exist in this repository, meaning that *this* class needs
+        # to be a class in this repo that inherits QuestionnaireAbstractBlock, noting however
+        # that *we* could be anything that subclassed QuestionnaireAbstractBlock, within
+        # or outside of this repo.
+        # 
+        # We therefore need to filter the class hierarchy in order to disregard classes 
+        # that are not part of this repository. Fortunately, mro() presents the class hierarchy
+        # in the correct order, thus, the correct class name to use for the html
+        # template will be the first item in the filtered list.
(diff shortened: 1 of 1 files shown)


lpm0073:
each of the three circleci build steps fail due to something happening within the linux instance that hosts the circleci processes. As best I can tell, this is not something that I can control nor remedy. Is that correct?

arbrandes:
@lpm0073, apologies for the delay in ACKing this, but it's now on our radar again. :)

arbrandes:
@lpm0073 

> each of the three circleci build steps fail due to something happening within the linux instance that hosts the circleci processes

[This WIP PR](https://github.com/open-craft/problem-builder/pull/332) is likely intended to fix that.

And on that note, @giovannicimolin and @xitij2000, would either of you have time to review Lawrence's PR, by any chance?

giovannicimolin:
@arbrandes I can review this PR once the CI checks are fixed. CC @lpm0073

lpm0073:
Let me know if its something simple (black, flake8, etcetera) that I might be able to do pre-commit. 

Lawrence McDaniel
__________________________________
📡  Full-Stack Developer
CV: lawrencemcdaniel.com | linkedin.com/in/lawrencemcdaniel
Blog: blog.lawrencemcdaniel.com
Code Samples: horrors.lawrencemcdaniel.com | github.com/lpm0073
Contact: ***@***.*** | +1 (617) 834-6172
Address: 210 Broadway, Suite 201, Cambridge, MA 02139

giovannicimolin:
@lpm0073 We're updating it to match the latest supported release on https://github.com/open-craft/problem-builder/pull/332. No need for any follow-ups here for now.

xitij2000:
@giovannicimolin  Perhaps we can test this locally since the integration tests are passing there so as not to block this ticket?

giovannicimolin:
@lpm0073 Sorry for the delay here, the master branch CI is now fixed. Can you rebase this PR? I'll review it early this week. Thanks!

CC @xitij2000

PR #334, opened 2022-01-07
feat: add support for celery 5.x

iamsobanjaved:
While testing celery upgrade PR in the edx-platform (https://github.com/edx/edx-platform/pull/29046/), found out that this xblock isn't compatible with celery 5.x. So this PR is making this xblock celery 5.x compatible. This xblock is private requirement of edx-platform

### DIFF
--- a/problem_builder/tasks.py
+++ b/problem_builder/tasks.py
@@ -3,7 +3,7 @@ Celery task for CSV student answer export.
 """
 import time
 
-from celery.task import task
+from celery import shared_task
 from celery.utils.log import get_task_logger
 from django.contrib.auth.models import User
 from django.db.models import F
@@ -23,7 +23,7 @@ from .sub_api import sub_api
 logger = get_task_logger(__name__)
 
 
-@task()
+@shared_task()
 def export_data(course_id, source_block_id_str, block_types, user_ids, match_string):
     """
     Exports student answers to all supported questions to a CSV file.
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,7 @@ from setuptools.command.install import install
 
 # Constants #########################################################
 
-VERSION = '5.0.0'
+VERSION = '5.1.0'
 
 # Functions #########################################################
 


robrap:
Thanks @iamsobanjaved.
1. I assume this would be a breaking change that would require a major version change?
2. I will have @gabor-boros, who worked on https://github.com/edx/edx-platform/pull/29046, support landing fixes of any issues you find while testing his PR for deployment.

gabor-boros:
Oh, nice! Thank you for letting us know.

I'll review this by the end of Monday and cut a new release with a version bump.

robrap:
@gabor-boros: On our end, @iamsobanjaved is leading the testing/deployment efforts at this point. I'll let him answer your questions.

However, 2 thoughts:
1. It would probably help to share here links to the two PRs for context. Maybe they need to be created first? What repo(s)?
2. [inform] We use celery flower, and are running into some testing issues around the transition/upgrade that may cause additional delays.

gabor-boros:
@robrap @iamsobanjaved 

I just checked and `shared_task` was available in older Celery versions too, therefore no need to merge those PRs I mentioned before. Correct me if I'm wrong, but the `shared_task` behaviour didn't change over the 4.4.7 (used in edx-platform) and 5+, so this is not a breaking change to me, but more like a fix, so the next version number of this package could be `5.1.0`  as it is more like a "backward compatible addition". Or did I get something wrong?

@iamsobanjaved before merging this, could you please bump the version number?

iamsobanjaved:
@gabor-boros Yes, this change is backward compatible so bumped the version to `5.1.0`

gabor-boros (approved):
👍  🎉 @iamsobanjaved Thank you for the PR again.

- [x] I read through the code
- [x] I checked celery documentation

gabor-boros:
@iamsobanjaved FYI: the new package is released https://pypi.org/project/xblock-problem-builder/

PR #349, opened 2023-05-26
fix: do not use the `runtime.replace_urls` when the runtime is not ready

Agrendalath:
This happens while generating the Block Structures.

Please see the description of https://github.com/openedx/xblock-drag-and-drop-v2/pull/318 for a detailed explanation of this scenario. Unlike that change, this one is backward-compatible with Maple. I didn't switch this XBlock to the `replace_urls` runtime service because it will require more changes - therefore, it makes more sense to schedule a separate task to remove all deprecated runtime attributes from here in one pass and perform detailed regression tests. This one is only a simple fix.


bradenmacdonald (approved):
Thanks!

bradenmacdonald:
@Agrendalath Can you please tag a release so this gets onto PyPI?

PR #459, opened 2025-06-04
feat: update ubuntu version to latest

Faraz32123:


Faraz32123:
@Agrendalath As you can see CI failures for first commit, what are your suggestions?

PR #482, opened 2026-03-03
chore: Upgrade Python requirements

Agrendalath:
`pip-tools` needed a manual version bump.

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -6,22 +6,22 @@
 #
 appdirs==1.4.4
     # via fs
-asgiref==3.8.1
+asgiref==3.11.1
     # via django
-boto3==1.38.37
+boto3==1.42.59
     # via fs-s3fs
-botocore==1.38.37
+botocore==1.42.59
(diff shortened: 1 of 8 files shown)


samuelallan72 (changes requested):
@Agrendalath please check this - CI is failing because pkg_resources is missing - it may need a similar fix to https://github.com/open-craft/xblock-html/pull/181/changes/dd23380735a9a12cf53aa88053b69f6781b11e3b . Please also note my comment about local_resource_url on that commit too. :)

Agrendalath:
@samuelallan72, the last supported version for this XBlock was Ulmo. We will no longer be maintaining it. Ref: https://github.com/open-craft/problem-builder/pull/420.

samuelallan72:
Thanks for confirming @Agrendalath :+1:

samuelallan72:
@Agrendalath since this is the case, could you check https://github.com/open-craft/problem-builder/pull/483 please? :)

PR #483, opened 2026-07-28
ci: disable auto dependency upgrades

samuelallan72:
## Description

We're no longer maintaining this, so we don't want to deal with the automated dependency upgrade PRs.

## Supporting information

Ref: https://github.com/open-craft/problem-builder/pull/482#issuecomment-5092907213

Private-ref: https://tasks.opencraft.com/browse/SE-6622

## Testing instructions

- verify this is what we want
- verify this correctly removes all references to the auto dep updates

## Author concerns

Do we want to just archive this repo completely, rather than removing this? If we're not maintaining this any more, archiving is probably the way to go. cc @bradenmacdonald


Agrendalath:
@xitij2000, can we archive this repo? If yes, we can probably close this PR.

cc: @samuelallan72

xitij2000:
@Agrendalath  Not yet, we are still trying to work on a replacement.

samuelallan72:
@xitij2000 does the readme warning need updating? :

> Problem Builder is no longer being developed by OpenCraft. It will be supported with compatibility and security fixes till the Ulmo release, after which this repository will be stop receiving further updates and eventually be archived.

Are we planning to maintain this until we find a replacement? Did I hear that this will stop working altogether in a future openedx-platform release?

xitij2000:
@samuelallan72 Yes, till we properly deprecate this and have a replacement, we need to maintain it.

